Issue #6, opened 2015-09-17
Minimize dependencies for package


satra:
- [ ] provenance optional (depends on nipype/rdflib)
- [ ] nipype optional (change command line interfaces to native functions)

hanke:
I agree, both would be nice. Making rdflib optional is a must (IMHO). Do you think it is feasible to simply inject copies of the nipype Interfaces to make nipype a weaker dependency?

satra:
@hanke - i started looking at this and the nipype interfaces can definitely be incorporated in some way. i just realized how much we have abstracted out of the interface (node execution ensures local symlinks to ensure that dcmnii only converts the relevant files, formatting the command line, etc.,.).

also i think it would be nice to encode provenance (since rdflib is a python only library, i'm not too concerned about that). but nipype isn't pure python (in terms of dependencies - traits requires compiling).

so it's possible, the question really is, should we?

mgxd:
after discussing with satra, some thoughts:

dcmstack makes installation harder than it should, we could:
* move grouping and csaheader functionality from `dcmstack` to `nibabel`
* drop `dcmstack` dependency

will require some deeper understanding of dcmstacks implemention

mgxd:
heudiconv now fails with the latest pydicom - more incentive to drop `dcmstack` dependency (#167)

yarikoptic:
I would agree that I think it makes sense to switch to a native call to dcm2niix to avoid dependency on a lovely but relatively heavy nipype from which we just use a single interface.

PR #9, opened 2015-09-30
Make prov optional, turn it off by default.

chrisfilo:
This partially addresses #6.


hanke:
LGTM, +1, (like), ...

PR #16, opened 2016-06-15
enh - dicom compression for bids format

mgxd:
LMK other suggestions!


satra:
@mgxd - can we also add the additional info to the json always.

mgxd:
@satra done

satra:
i can't see where the json info is added to the json output by dcm2niix.

essentially:

1. when dcm2niix is used to generate bids, then the remaining dicom key-value pairs should be added to the bids json
2. when bids is not used, then the standard json info is added.

so from an algorithm perspective, the standard json info should always be created, but the bids info from dcm2niix gets added in the case of the bids option.

bin/heudiconv
```diff
@@ -238,14 +231,25 @@ def embed_nifti(dcmfiles, niftifile, infofile, force=False):
     #orig_nii.update_header()
     #orig_nii.to_filename(niftifile)
     meta = ds.NiftiWrapper(new_nii).meta_ext.to_json()
-    with open(infofile, 'wt') as fp:
+    with open(infofile, 'at') as fp:
```
mgxd: @satra essentially I add the info to the scaninfo file here, if bids option is called this is added after dcm2niix's, if not this is this only info in the file

satra: have you tested this with and without the bids flag? and in both cases all the other dicom info is there?

mgxd: yeah both cases still produce the dicom info - [this](http://pastebin.com/6dT8ssGE) is how the bids info would look

satra: that's not a valid json file i believe.

instead of appending text, read the json using `json.load` and then add the bids keys or read the bids and add the dcmmeta-keys. and then write it back out.

PR #17, opened 2016-06-25
Added dependencies to readme

jpellman:


satra:
thanks - we should be setting up a proper setup.py for this as well.

PR #19, opened 2016-07-06
BF+DOC+ENH: converted tabs, fixed use of some variables, some PEP8ing, typos

yarikoptic:
Just had a look at the beast.  I must say "it needs tests" ;-)  Please review since I did some code changes, and hopefully have done de-tabification correctly.

Would you happen to have some DICOMs, probably on phantoms which could be used to exercise those heuristics?  then we could establish all the travis etc 

@chrisfilo may be you have a better alternative already for DICOM -> BIDS we could use with our soon-to-come Siemens 3T? ;-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,16 @@
 # HeuDiConv
 
-This is a flexible dicom converter for organizing brain imaging data into
+This is a flexible DICOM converter for organizing brain imaging data into
 structured directory layouts.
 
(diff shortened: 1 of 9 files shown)


chrisfilo:
For now heudiconv is the best one-stop-shop automatic solution for DICOM -> BIDS conversion. You can also write your own set of custom scripts based on dcm2niix (which creates valid BIDS json sidecars).

PR #24, opened 2016-07-11
added --dbg to cmdline to fall into interactive debugger in case of error

yarikoptic:
"borrowed" from datalad ;)  thus initiated using some kind of lgr etc -- good things to come eventually I bet ;)

ah -- with this compare:

```shell
$> bin/heudiconv  -d /tmp -s S -c dcm2niix -f asd   
  File "bin/heudiconv", line 623, in <module>
    is_bids=args.bids)
  File "bin/heudiconv", line 429, in convert_dicoms
    sdir = dicom_dir_template % sid
TypeError: not all arguments converted during string formatting
```

to

```shell
$> bin/heudiconv --dbg -d /tmp -s S -c dcm2niix -f asd
  File "bin/heudiconv", line 623, in <module>
    is_bids=args.bids)
  File "bin/heudiconv", line 429, in convert_dicoms
    sdir = dicom_dir_template % sid
TypeError: not all arguments converted during string formatting
()
> /home/yoh/deb/gits/pkg-exppsy/heudiconv/bin/heudiconv(429)convert_dicoms()
-> sdir = dicom_dir_template % sid
(Pdb) print dicom_dir_template
/tmp
```

### DIFF
--- a/bin/heudiconv
+++ b/bin/heudiconv
@@ -26,6 +26,8 @@ import tarfile
 import dicom as dcm
 import dcmstack as ds
 
+import logging
+lgr = logging.getLogger('heudiconv')
 
 def save_json(filename, data):
     """Save data to a json file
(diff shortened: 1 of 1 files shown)


satra:
@yarikoptic - what do you think about these options?

https://github.com/nipy/nipype/blob/master/bin/nipype_display_crash#L66

allows both normal python and ipython debugger support.

satra:
@yarikoptic - will you be updating this pull request :)

yarikoptic:
May be... ;-) could be done as a next PR ;-)

PR #26, opened 2016-08-03
BF: defined outname_bids as early as possible to could be used + import again Node, Function before use

yarikoptic:
otherwise call 
```
(venv-heudiconv)kodiweera@head1:/data/BIDS/heudiconv$ /data/BIDS/heudiconv/bin/heudiconv -d '/data/BIDS/prisma/MPRAGE_GRAPPA2_%s/*' -f /data/BIDS/heudiconv/heuristics/convertall.py -s 0005 -c dcm2niix -o /data/BIDS/prisma/output2 -b
```
was bugging out on us... this thing needs tests!

### DIFF
--- a/bin/heudiconv
+++ b/bin/heudiconv
@@ -262,6 +262,7 @@ def convert(items, anonymizer=None, symlink=True, converter=None,
         prefix = item[0]
         print('Converting %s' % prefix)
         dirname = os.path.dirname(prefix + '.ext')
+        outname_bids = prefix + '.json'
         print(dirname)
         if not os.path.exists(dirname):
             os.makedirs(dirname)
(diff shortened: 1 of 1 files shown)


satra:
can we use some phantom data for tests?

yarikoptic:
yes, we should, and I will definitely provide a set from our brand new Siemens whenever it becomes operational (hopefully in few weeks)... meanwhile I have some DICOMs from Phillips lying around, but they aren't my "target" ;)

yarikoptic:
but I think for this particular and probably many other issues it needs real unit-tests by excercising those functions with hypothetical data structures, and possibly mocking out some functions expecting stuff being provided from dicoms.

yarikoptic:
so should we wait for tests or just get it merged for now?  anyone to bless? ;)

satra:
blessed

PR #29, opened 2016-09-07
enh: create dockerized version of heudiconv

satra:


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM continuumio/miniconda
+
+MAINTAINER <satra@mit.edu>
+
+RUN apt-get update && apt-get upgrade -y && apt-get install -y g++ && apt-get clean -y && apt-get autoclean -y && apt-get autoremove -y
+RUN cd /tmp && git clone https://github.com/neurolabusc/dcm2niix.git && cd dcm2niix/console/ && g++ -O3 -I. main_console.cpp nii_dicom.cpp jpg_0XC3.cpp ujpeg.cpp nifti1_io_core.cpp nii_ortho.cpp nii_dicom_batch.cpp  -o dcm2niix -DmyDisableOpenJPEG -DmyDisableJasper && cp dcm2niix /usr/local/bin/
```
yarikoptic: Fwiw we do provide dcm2niix from NeuroDebian, and there are NeuroDebian docker images
Just saying

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM continuumio/miniconda
+
+MAINTAINER <satra@mit.edu>
+
+RUN apt-get update && apt-get upgrade -y && apt-get install -y g++ && apt-get clean -y && apt-get autoclean -y && apt-get autoremove -y
+RUN cd /tmp && git clone https://github.com/neurolabusc/dcm2niix.git && cd dcm2niix/console/ && g++ -O3 -I. main_console.cpp nii_dicom.cpp jpg_0XC3.cpp ujpeg.cpp nifti1_io_core.cpp nii_ortho.cpp nii_dicom_batch.cpp  -o dcm2niix -DmyDisableOpenJPEG -DmyDisableJasper && cp dcm2niix /usr/local/bin/ 
+RUN conda install -y -c conda-forge nipype && pip install https://github.com/moloney/dcmstack/archive/master.zip
+RUN curl -O https://raw.githubusercontent.com/nipy/heudiconv/master/bin/heudiconv && chmod +x heudiconv && cp heudiconv /usr/local/bin/
```
yarikoptic: My point is that this docker image will be completely irreproducible itself since two tools installed straight from git repos master without tracking their versions (I don't think any of them uses git describe to assign their versions, do they?)

PR #31, opened 2016-09-17
NF+ENH: make heuidiconv being able to layout the files deducing subject_id and using other DICOM fields

yarikoptic:
main goal is to provide sorting of dicoms deducing full hierarchy (investigator/study/sub-) from dicoms folder (possibly carrying multiple sessions eventually).  Submitting as WiP (work in progress) to seek input/blame/etc.  Theoretically (since no tests ;) ) it should remain compatible with existing uses. Most representative change/summary is in commit 60f7aa11d44a05f40d837f4c352a0fdf8c5e3221 :

```
Enhancements:
 - use named tuple for SeqInfo for sensible access to the collected fields
 - added patient_id, study_description and referring_physician_name
   to the SeqInfo
 - Q&D logging setup, controlled by HEUDICONV_LOGLEVEL env var
 - added log message in a few spots around

Changes in logic:
 - allow empty subject id which then treats provided path
   not as a template but as a directory under which to look for DICOMs
 - if subject id is empty -- assess it from collected seqinfo
   relying on `get_session_subject_id` method in the heuristic file

Enhancemets:
 - some code duplication removed by providing more variables via
   parameters, etc
 - removed some variables in `convert` definition which were not used
 - added few parameters to be available during formatting
   - bids_subject_session_prefix, (e.g. sub-SUBJECT_ses-SESSION if ses defined)
   - bids_subject_session_dir (e.g. sub-SUBJECT/ses-SESSION if ses is defined)
```


bin/heudiconv
```diff
@@ -26,8 +26,40 @@ import tarfile
...
+# Rudimentary logging support.  If you want it better -- we need to have
+# more than one file otherwise it is not manageable
+logging.basicConfig(
+    format='%(levelname)s: %(message)s',
+    level=getattr(logging, os.environ.get('HEUDICONV_LOGLEVEL', 'INFO'))
+)
+lgr.debug("Starting the abomination")  # just to "run-test" logging
+
+
+SeqInfo = namedtuple(
```
yarikoptic: eventually I might consider using attrs module to use here (https://attrs.readthedocs.io/en/stable/) instead of namedtuple

matthew-brett:
Still no tests :( ?

yarikoptic:
We will get there ;-)  for now I have it not working at all for me... then I would make it work for me and solidify with the tests ;)

yarikoptic:
here is the first bird @matthew-brett :  2aea89ab4a0ab648cf7890d7b370f15dc7eee324  join the fun ;)

yarikoptic:
woohoo -- 9% coverage already @matthew-brett ! ;)

satra:
shall we also see if we can remove dependency on dcmstack and simply is nibabel? that would allow this to work with py3 as well.

yarikoptic:
It might simplify things indeed since we don't have that package for precise... But given that there is no tests, you night burn yourself ;-)
Is it possible to have dicoms with actual data being stripped so they would be small?

satra:
essentially all i am using dcmstack for are:
1. do grouping
2. extract additional metadata, but focuses on siemens only

@matthew-brett can possibly advise if nicom is ready for metadata across scanner types

i can easily rewrite the grouping function simply using pydicom or nicom, but testing beyond some trivial test cases gets hard. we could use the phantom data you supplied to run some tests on circle (cache the raw data so we don't have to download each time) and then do the change.

that will get us a few more tests, definitely more coverage and then we can start simplifying the dependencies.

yarikoptic:
Fwiw Within upcoming days I hope to collect more of phantom data
But if I could also single check if I have broken previously working setups, it would be great. Eg did you try that data of mine I shared before?

satra:
@yarikoptic - yes i did. that's the dataset i used here:

https://neurohackweek.github.io/lesson-dataflows/02-heudiconv/

yarikoptic:
Thanks @satra -- FYI playing with that data lead me to https://github.com/neurolabusc/dcm2niix/issues/45

matthew-brett:
Guys - happy to work with y'all on nibabel to make it useful for this purpose.

Brendan Moloney has expressed interest in moving dcmstack code over to nibabel, so maybe this will be a good engine for that to happen.

Did y'all see what I did to test against larger data in nibabel?    Basically I use a `nibabel-data` directory with submodules to contain optional data for testing, and there's  a decorator for test functions that checks for the presence of the needed data before running tests. See https://github.com/nipy/nibabel/blob/master/nibabel/tests/test_parrec_data.py for an example.

yarikoptic:
pushed inhere few more commits which if anything fix what I have broken.  I would appreciate review and/or trying on your cases and merge.  I have submitted another PR which sits on top of this one which does more evil and is WiP https://github.com/nipy/heudiconv/pull/32

yarikoptic:
woohoo, thank you Satra.  I wonder though if you had a chance to try it to see if I haven't broken anything.  There is WiP in #32 but that one would definitely break things for you I think :-( and needs to be brought back into "compatible" state

satra:
@yarikoptic - i'll also try the docker container to see if anything is amiss :)

@mgxd - can you try running the latest heudiconv on some of our data?

mgxd:
@satra still looking good after conversion

PR #32, opened 2016-10-10
WiP RF+ENH: further RF to support flexible heuristic to name sequences so they follow BIDS in their naiming

yarikoptic:
Sits on top of https://github.com/nipy/heudiconv/pull/31

It implements perspective heuristic which would allow simply to name sequences following BIDS convention.  More details available at https://docs.google.com/document/d/1R54cgOe481oygYVZxI7NHrifDyFUZAjOBwCTu7M7y48/edit?usp=sharing

TODOs
- [x] harmonize location (and may be naming) of stored "info" (now sticks out too much)
- [x] make it work in multiple sessions within the same study
- ~move derivatives~ (atm there is no derivatives)
- [x] make sure original dicoms are tarballed and stored in appropriate subdir of BIDS dataset
- [x] basic support for DataLad ;)
  - [x] annotate dicom tarballs and anatomicals in annex with `distribution-restrictions=sensitive`
- [x] made DICOM tarballs reproducible

one of the test DICOM datasets to try it on is http://datasets.datalad.org/?dir=/dicoms/dartmouth-phantoms/bids_test3


satra:
@yarikoptic - the heuristics has become very complicated now - requires a sufficiently enlightened python programmer. can we have the simpler version, which was relatively easy for people to learn?

also: 

1. we should not rely on dicomstack but just use pydicom + nibabel to group the dicoms
2. we should embed all the dicom metadata into the bids info json file. (currently bids is very limited w.r.t dicom metadata)

yarikoptic:
I didn't change any of heuristics IIRC (at least not yet) ;) 
I just allowed for what was not allowed before (e.g. sorting dicoms by study UIDs etc).
custom handling for BIDS files simply didn't do the right thing e.g. for fieldmaps etc

indeed currently it is a heap of changes, but once again -- I am just trying first to tune it so I could do what I need to accomplish, and later would hopefully look to catch back diversions from original implementation... which is tough as you know since there were not a single test of any kind.

yarikoptic:
@satra would you be so kind to contribute (I believe you could just push to this branch in my clone) the conda setup in .travis.yml so we could test again against py3s?

satra:
@yarikoptic - you mean like nipype with miniconda?

yarikoptic:
yeap. because that is what causes the trouble (pulls in scipy) to install via pip

satra:
@yarikoptic - let us know when you would like this merged? seems like this is growing quite a bit. should we have some documentation that makes this more general than dartmouth ? :)

satra:
@yarikoptic - ah i see i was supposed to send a pull-request for conda based install.

yarikoptic:
Hi @satra -- thanks for getting back to it! ;)  indeed conda setup could help

We have a bunch of other changes which we haven't dared to push in here yet, but even without them IIRC we did broke compatibility with previous use-cases for heudiconv.  Would need re-analysis to see what that was and how could I bring it back so I don't think it should be merged as is, unless you would confirm that this version still functions correctly on your use-cases ;)

We have tried to push all Dartmouth-specific changes into our heuristic file, but I guess we could generalize it even more.

bin/heudiconv
```diff
@@ -566,81 +605,172 @@ def convert(items, symlink=True, converter=None,
...
+            json.dump(json_, open(json_phasediffname, 'w'), indent=2)
+            os.chmod(json_phasediffname, 0o0440)
+
+        # phasediff one should contain two PhaseDiff's
+        #  -- one for original amplitude and the other already replicating what is there
+        # so let's load json files for magnitudes and
+        # place them into phasediff
+
+
+def embed_metadata_into_nifti(converter, is_bids, item_dicoms, outname,
```
yarikoptic: interesting... I have blindly kept this one, but then sanitized .json which it produced since it was way too heavy (removed all cs) and not neatly formatted, so I did reformatting with some really obnoxious use of `re` replacements...   here is that code:  https://github.com/dartmouth-pbs/heudiconv/blob/mvdoc-fixup-protocols/bin/heudiconv#L157

satra:
@yarikoptic - i really think we should converge on a common model while adapting to the needs of your approach. shall we do a quick hangout to go over the pieces and understand where the divergence has occurred? this PR has gotten quite big and we really should try to maintain one master.

yarikoptic:
@satra sure...  choose the best for you time tomorrow before 2pm and let me know ;)
meanwhile -- the easiest way to catch divergences is to run conversion with master and then with our PR branch I guess.  IIRC major one was me introducing full paths into heudiconv meta data iirc.

satra:
@yarikoptic - 11-12pm thu 9

yarikoptic:
added to the calendar and invited you an Matteo (with whom we tortured heudiconv)

yarikoptic:
ok @satra -- we have fixed it so it works with `-d ... -s subj` invocation in c3d27a3 .
I can't quite pin point the divergence in file names recording I had in mind -- may be it is fictional ;)  but another main divergence is the location to store subject and info directories.
-   To stay a bit more compliant with BIDS, I guess, we do not info into each subject directory but place into top level `.heudiconv` while maintaining the same ses/subj hierarchy
- Our code now also does not generate leading subject directory in the output "automagically" for the case when -s is actually provided ;)  I guess this could and should be fixed up

you can give it a try now on some dataset of yours and possibly catch other divergences

satra:
@yarikoptic - any chance we spend an afternoon hacking to merge these two branches before it gets completely out of hand. any chance you could merge with master (if possible)?

yarikoptic:
:+1: 
started to merge and resolve conflicts in heudiconv but only later realized how much and severity of them...  would indeed take some good time to resolve them all.  
there were no conflicts in `tests/` though - since none was added/changed in master :-1: 
But did you try this version before on any of your data/usecases?

satra:
@mgxd - did you try this branch on BANDA? if not can you try it on a participant or two?

bin/heudiconv
```diff
@@ -73,8 +165,57 @@ def save_json(filename, data):
         Dictionary to save in json file.
 
     """
+    # adds trailing whitespaces due to indent
+    # see https://bugs.python.org/issue16333
+    # seems to be fixed in 3.4, for now fixing manually
+    # not only for aestetics but also to help guaranteeing the
```
matthew-brett: "aesthetics"

bin/heudiconv
```diff
@@ -73,8 +165,57 @@ def save_json(filename, data):
         Dictionary to save in json file.
 
     """
+    # adds trailing whitespaces due to indent
```
matthew-brett: Bug remains for Python 2.7 on my laptop, fixed for 3.4, on my laptop.

Maybe

```
# Remove trailing whitespace incorrectly inserted when using indent
```

bin/heudiconv
```diff
@@ -73,8 +165,57 @@ def save_json(filename, data):
...
+    # seems to be fixed in 3.4, for now fixing manually
+    # not only for aestetics but also to help guaranteeing the
+    # same result across versions of Python
     with open(filename, 'w') as fp:
-        json.dump(data, fp, sort_keys=True, indent=4)
+        json_str = json.dumps(data, sort_keys=True, indent=4)
+        fp.write(json_str.replace(' \n', '\n'))
+
+
+def slim_down_info(j):
```
matthew-brett: More descriptive variable name than `j`?

bin/heudiconv
```diff
@@ -142,92 +283,133 @@ def find_files(regex, topdir=curdir, exclude=None, exclude_vcs=True, dirs=False)
 find_files.__doc__ %= (_VCS_REGEX,)
 
 
-def process_dicoms(fl, dcmsession=None, basedir=None, dcmfilter=None):
-    """Process list of dicoms and return  seqinfo and file group
+def group_dicoms_into_seqinfos(fl, flfilter=None, dcmfilter=None, grouping='studyUID'):
```
matthew-brett: Nice renaming, expansion of docstring.

bin/heudiconv
```diff
@@ -142,92 +283,133 @@ def find_files(regex, topdir=curdir, exclude=None, exclude_vcs=True, dirs=False)
...
+    """Process list of dicoms and return seqinfo and file group
+
+    `seqinfo` contains per-sequence extract of fields from DICOMs which
+    will be later provided into heuristics to decide on filenames
 
     Parameters
     ----------
     fl : list of str
```
matthew-brett: How about `filename_list` instead of `fl`?

bin/heudiconv
```diff
@@ -142,92 +283,133 @@ def find_files(regex, topdir=curdir, exclude=None, exclude_vcs=True, dirs=False)
...
-      corresponds to multiple sessions
-    basedir : str, optional
-      Base directory relative to which filenames are provided in `fl`
+    flfilter : callable, optional
+      Applied to each of fl. Should return True if file needs to be kept,
+      False otherwise. Used to filter fl
     dcmfilter : callable, optional
       If called on dcm_data and returns True, it is used to set
       series_id
+    grouping : str ('studyUID', 'accession_number') or None, optional
```
matthew-brett: I think the numpy documentation standard would be

```
grouping : {'studyUID', 'accession_number', None}, optional
```

bin/heudiconv
```diff
@@ -142,92 +283,133 @@ def find_files(regex, topdir=curdir, exclude=None, exclude_vcs=True, dirs=False)
...
+        what to group by: studyUID or accession_number
 
     Returns
     -------
-    seqinfo, filegrp : list of list, dict
+    seqinfo : list of list
```
matthew-brett: Nice improvement to docstring.

bin/heudiconv
```diff
@@ -142,92 +283,133 @@ def find_files(regex, topdir=curdir, exclude=None, exclude_vcs=True, dirs=False)
...
       `seqinfo` is a list of info entries per each sequence (some entry 
       there defines a key for `filegrp`)
-      `filegrp` is a dictionary with files groupped per each sequence.
+    filegrp : dict
+      `filegrp` is a dictionary with files groupped per each sequence
```
matthew-brett: grouped

bin/heudiconv
```diff
@@ -142,92 +283,133 @@ def find_files(regex, topdir=curdir, exclude=None, exclude_vcs=True, dirs=False)
...
     """
+    allowed_groupings = ['studyUID', 'accession_number', None]
+    if grouping not in allowed_groupings:
+        raise ValueError('I do not know how to group by {0}'.format(grouping))
+    per_studyUID = grouping == 'studyUID'
```
matthew-brett: Clearer leaving these guys as string comparisons?

bin/heudiconv
```diff
@@ -142,92 +283,133 @@ def find_files(regex, topdir=curdir, exclude=None, exclude_vcs=True, dirs=False)
...
-            del mw.series_signature['ICE_Dims']
-        except:
-            pass
-        try:
-            del mw.series_signature['SequenceName']
-        except:
-            pass
+            studyUID_ = mw.dcm_data.StudyInstanceUID
+        except AttributeError:
+            #import pdb; pdb.set_trace()
```
matthew-brett: Remove comment cruft?

bin/heudiconv
```diff
@@ -142,92 +283,133 @@ def find_files(regex, topdir=curdir, exclude=None, exclude_vcs=True, dirs=False)
...
+            lgr.info("File %s is missing any StudyInstanceUID" % filename)
+            studyUID_ = None
```
matthew-brett: Rename as `this_studyUID` or similar to distinguish from `studyID`?

satra:
@matthew-brett - there is a PR (https://github.com/dartmouth-pbs/heudiconv/pull/26) to this PR that brings this up to speed with master. while many of this comments would still persist, i think it would be better to review with that PR merged into this.

@yarikoptic - have you had a chance to test it?

matthew-brett:
@satra - sorry, I didn't realize this was out of date.  I'll try and update when your PR gets merged.  @yarikoptic - another plea from me - while it's still fresh in my mind.

yarikoptic:
Woohoo -- conflicts are gone and those few tests we have still pass! thanks @satra

matthew-brett:
Added a PR to add Python 3 testing - which fails:

https://github.com/dartmouth-pbs/heudiconv/pull/27

yarikoptic:
With the review and few trials -- seems to perform correctly with @satra 's merge of master (mucho gracias).  Left out TODOs and more could be handled later (planing to do some work at OHBM BrainHack) so might be worthwhile considering to get it merged? ;)

matthew-brett:
Any thoughts about merge here?

Yarik - did you have time to look at my comments?

I'd like to do some development here, but please let me know if there may be a significant delay in getting this merged, so I can concentrate on other stuff.

yarikoptic:
@matthew-brett what comments? I think I have merged your PR and no comments were left unanswered

matthew-brett:
Yarik - I didn't address my own comments in my PR.  Have a look at the current Files Changed tab - I can see various comments of mine there that I think aren't addressed - can you?

satra:
@yarikoptic - any chance this can be looked at and @matthew-brett's comments addressed?

bin/heudiconv
```diff
@@ -1,19 +1,25 @@
...
 
-This function uses DicomStack and mri_convert to convert Siemens
-TrioTim directories. It proceeds by extracting dicominfo from each
-subject and writing a config file $subject_id/$subject_id.auto.txt in
-the output directory. Users can create a copy of the file called
-$subject_id.edit.txt and modify it to change the files that are
-converted. This edited file will always overwrite the original file. If
-there is a need to revert to original state, please delete this edit.txt
-file and rerun the conversion
+This script uses DicomStack and mri_convert to convert DICOM directories.
```
yarikoptic: took your version and expanded... most probably will need to resolve conflicts with master by now ;) thanks

bin/heudiconv
```diff
@@ -58,9 +95,64 @@ SeqInfo = namedtuple(
...
+                    # subject/session.
+                    # So 'locator' is pretty much an assignment from StudyInstanceUID
+                    # into some place within hierarchy
+        'session',  # could be None
+        'subject',  # should be some ID defined either in cmdline or deduced
+    ]
+)
+
+
+class TempDirs(object):
```
yarikoptic: Valid concern!  But primarily for the case whenever such refactoring would (and if, since original idea was to have a single script IIRC) happen -- ATM it is just as fine imho, so just added a TODO comment on that

bin/heudiconv
yarikoptic: thanks, done

bin/heudiconv
yarikoptic: I see no reason for type (`_list`) suffix really -- making it `files` (noone works with open files in here, so no fear of confusing with file instances) and `flfilter` into `file_filter`

bin/heudiconv
yarikoptic: could be done imho either way.  this centralizes comparison to the string value into a variable.  Happen we decide to refactor anything and complicate the logic to decide e.g. on `per_studyUID` value -- will be just a single point to change instead of hunting for all the comparisons to a string value.  So I would leave it as is for now

bin/heudiconv
yarikoptic: thanks -- cleaned up here and elsewheres

bin/heudiconv
yarikoptic: ok, did to `file_studyUID`

satra:
@yarikoptic - LGTM - i'm going to file an issue about cluster/distributed computing support to be added back after this is merged.

PR #47, opened 2017-03-20
fix: change default, update readme

mgxd:
@satra we should also update the dicominfo columns in the readme, as there are now more


satra:
we should. also it would be nice to change seqinfo into a list of `namedtuple`

PR #50, opened 2017-03-24
enh: make output more bids like

mgxd:
* remove non-alphanumeric characters from subject name
* `sub-` should be added to output subject directory
* `TaskName` key to functional json
* use img.affine property

### DIFF
--- a/bin/heudiconv
+++ b/bin/heudiconv
@@ -669,16 +669,6 @@ def convert_dicoms(subjs, dicom_dir_template, outdir, heuristic_file, converter,
                 "atm we support processing only 1 subject/session at a time"
             sess, sid = session_subject_ids.pop()
 
-        #
-        # Annonimization
-        #
-        anon_sid = sid
-        if anon_sid_cmd is not None:
-            from subprocess import check_output
-            anon_sid = check_output([anon_sid_cmd, sid]).strip()
-            lgr.info("Annonimized sid %s into %s", sid, anon_sid)
-        if anon_outdir is None:
-            anon_outdir = outdir
 
         if is_bids and not sid.isalnum(): # alphanumeric only
             old_sid = sid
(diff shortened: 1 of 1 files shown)


bin/heudiconv
```diff
@@ -657,26 +668,36 @@ def convert_dicoms(subjs, dicom_dir_template, outdir, heuristic_file, converter,
                 "atm we support processing only 1 subject/session at a time"
             sess, sid = session_subject_ids.pop()
 
+
+        if is_bids and not sid.isalnum(): # alphanumeric only
+            old_sid = sid
```
satra: i would just raise an error and have the user respecify the sid

satra: `raise ValueError('{} must be alphanumeric'.format(sid))`

bin/heudiconv
```diff
@@ -367,34 +367,42 @@ def conversion_info(subject, outdir, info, filegroup, basedir=None, ses=None):
...
-        nifti_image.to_filename(niftifile)
-        return ds.NiftiWrapper(nifti_image).meta_ext.to_json()
-    orig_nii = nb.load(niftifile)
-    aff = orig_nii.get_affine()
-    ornt = nb.orientations.io_orientation(aff)
-    axcodes = nb.orientations.ornt2axcodes(ornt)
-    new_nii = stack.to_nifti(voxel_order=''.join(axcodes), embed_meta=True)
-    meta = ds.NiftiWrapper(new_nii).meta_ext.to_json()
-    meta_info = json.loads(meta)
+    import dcmstack as ds
```
satra: perhaps move the dcmstack import into `min_meta` clause

mgxd: I was thinking this to potentially bypass a dependency, but it's also imported in `process_dicoms`. I still agree just so we do one less import in the node

bin/heudiconv
mgxd: the problem is some of the dataset IDs have underscores and dashes, which would have to be changed in the dicom directories as well. If we just raise an error, that process will have to be done manually

satra: i'm not following it here, can you provide an example?

mgxd: I'll reference the phantom conversion - the subject ID is `PHANTOM1_3`, but having `sub-PHANTOM1_3_func...` would mess up BIDS's separation by underscore. Rather than throwing an error and forcing you to rename the directory where the dicoms are, we can just drop the underscore and throw a warning that it was renamed. Really, this just makes it easier to do the demonstration next week

mgxd: @satra does this sound reasonable?

satra: reasonable, yes

PR #54, opened 2017-04-19
Change how the expression used to find DICOM files is formatted.

tsalo:
This new format allows users to include both subject id and session in
the search expression. Session is optional.

Also, Atom automatically cleaned up some of the whitespace in the
README file.

### DIFF
--- a/bin/heudiconv
+++ b/bin/heudiconv
@@ -816,7 +816,7 @@ s3
                         dest='dicom_dir_template',
                         required=True,
                         help='''location of dicomdir that can be indexed with
-                        subject id {participant} and session {session}. 
+                        subject id {subject} and session {session}.
                         Tarballs (can be compressed) are supported
                         in addition to directory. All matching tarballs for a
                         subject are extracted and their content processed in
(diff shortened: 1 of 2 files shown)


tsalo:
I have no idea why codecov/patch failed. Neither of the tests should be affected by my changes, as far as I can tell.

satra:
@tsalo - since we are using subjects everywhere else, perhaps we can stick to `{subject}` for now just to reduce a semantic clash - sorry i hadn't looked at the code before that previous statement.

no worries about codecov - we need to fix that.

bin/heudiconv
```diff
@@ -816,8 +816,9 @@ s3
                         dest='dicom_dir_template',
                         required=True,
                         help='''location of dicomdir that can be indexed with
-                        subject id. Tarballs (can be compressed) are supported
-                        in additions to directory. All matching tarballs for a
+                        subject id {participant} and session {session}.
```
satra: one more location :)

tsalo: Good catch. Thanks.

satra:
@mgxd - can we update the tutorial when this gets merged? 

since this is definitely an API breaking change, @tsalo could we raise a deprecation exception and point the user to the correct the correct syntax?

tsalo:
@satra Are you talking about something like:

```python
if '%s' in dicom_dir_template:
    lgr.error("""Formatting DICOM search pattern with '%s' is deprecated. 
Please indicate participant ID with {subject} and (optionally) timepoint with {session}.""")
```

satra:
@tsalo - unless we support a period of both options, we should just raise an `Exception` :) i can't remember if lgr.error raises an exception by default. i don't think so.

so perhaps just `raise ValueError('message')`

PR #56, opened 2017-04-19
Add dcm_dir_name to dicominfo (+ uc_bids heuristic)

danlurie:
It's not uncommon when scanning participants to have to re-run a scan due to motion/confusion/claustrophobia/etc. In these cases, our lab adds "XX" to the name of the folder containing DICOMS for the failed/incomplete scan. I've added a field to `dicominfo` (`dcm_dir_name`) to store this information so these scans can be easily excluded from conversion using `heudiconv`. 

Not sure what the policy is on adding new heuristics to the main repo, but I've included mine for now.

### DIFF
--- a/heuristics/uc_bids.py
+++ b/heuristics/uc_bids.py
@@ -1,0 +1,51 @@
+import os
+
+def create_key(template, outtype=('nii.gz',), annotation_classes=None):
+    if template is None or not template:
+        raise ValueError('Template must be a valid format string')
+    return template, outtype, annotation_classes
+
+def infotodict(seqinfo):
+    """Heuristic evaluator for determining which runs belong where
+    
+    allowed template fields - follow python string module: 
+    
+    item: index within category 
+    subject: participant id 
+    seqitem: run number during scanning
+    subindex: sub index within group
+    """
+    t1w = create_key('anat/sub-{subject}_T1w')
+    t2w = create_key('anat/sub-{subject}_acq-{acq}_T2w')
+    flair = create_key('anat/sub-{subject}_acq-{acq}_FLAIR')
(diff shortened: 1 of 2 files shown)


heuristics/uc_bids.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    t1w = create_key('anat/sub-{subject}_T1w')
+    t2w = create_key('anat/sub-{subject}_acq-{acq}_T2w')
+    flair = create_key('anat/sub-{subject}_acq-{acq}_FLAIR')
+    rest = create_key('func/sub-{subject}_task-rest_acq-{acq}_run-{item:02d}_bold')
+
+    info = {t1w: [], t2w: [], flair: [], rest: []}
+
+    for idx, seq in enumerate(seqinfo):
+        x,y,z,n_vol,protocol,dcm_dir = (seq[6], seq[7], seq[8], seq[9], seq[12], seq[3])
```
satra: @danlurie - i just asked @mgxd to update existing examples to use named tuple. you can literally do: `seq.dim1` instead of `seq[6]` for example. if in the future column ordering changes the heuristics would still work.

danlurie: Oh nice! <3 namedtuple.

satra:
i don't have a strong feeling about replacing `unspecified` with a new name at this point. historically, the `unspecified` pieces were there to conform to the output from freesurfer's `mri_convert`. but now that we use dcm2niix as the only (for present) converter, that historical reference is mostly useless.

more generally though i'd like to allow seqinfo to literally reflect more details and standard tags from the dicom series. (for a different PR, but just seeding the idea amongst folks).

PR #73, opened 2017-08-11
BF: many fixes for _scans.tsv files

yarikoptic:
- assign metadata also to the ones under sessions
- n/a for empty values (as the bible says)
- correct time stamps per file
- having -ses_ suffix where applicable (BIDS manuscript also was fixed, validator works correctly)
- sanitize more fields in .json files (just in case)


yarikoptic:
any feedback or a merge would be very welcome @satra? ;-)

PR #88, opened 2017-09-22
New utility: heudiconv_infodump

danlurie:
Much of the time when I go to create a new heuristic, I'm not super familiar with the scans I need to convert, so I don't have a good idea of what information to use to set up my heuristics. For a while, I was opening the DICOM headers using `pydicom` and manually matching the header fields to the heudiconv `SeqInfo` items. I realized it would be much easier to just have `heudiconv` generate a `SeqInfo` object for me and then write it to disk, so that's what this script does.

Call `heudiconv_infodump` with a `glob`-compatible file path template of the same kind used to call the main `heudiconv` script. For each scan that `heudiconv` finds, `heudiconv_infodump` will write the `SeqInfo` array to disk in the current working directory.

I thought this might be helpful for other people too. If so, I can do a proper docstring, etc. Let me know.

### DIFF
--- a/bin/heudiconv_infodump
+++ b/bin/heudiconv_infodump
@@ -1,0 +1,27 @@
+#!/usr/bin/env python
+
+import os
+import sys
+import imp
+from glob import glob
+
+# Use the specified file-path template to glob DICOM files
+fpt = sys.argv[1]
+
+# Import heudiconv
+heudi_bin = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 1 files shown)


satra:
i've updated the README to better highlight the tutorial :)

yarikoptic:
Not sure how that change in readme addressed it.. but there is `--command  ls` or smth like that which is relevant and may be anything similar could be coded/extended

satra:
@yarikoptic - if you go through the tutorial, it tells you to use `-c none` to generate info files.

PR #100, opened 2017-10-06
Update heudiconv: Handle AttributeError exception of line 484: image_type = tuple(dcminfo.ImageType) 

yinglilu:


mgxd:
@yinglilu thanks for the fix - i've added this in with b1e17d1

Issue #99, opened 2017-10-05
Handle AttributeError exception of line 484: image_type = tuple(dcminfo.ImageType)


yinglilu:
Hi heudiconv team,

Is it possible to add try-except of line 484: image_type = tuple(dcminfo.ImageType)?

We have some DICOM files missing the imageType tag, which choke the heudiconv:
'image_type = tuple(dcminfo.ImageType)
  File "/opt/anaconda2/lib/python2.7/site-packages/dicom/dataset.py", line 256, in __getattr__
    "'{0:s}'.".format(name))
AttributeError: Dataset does not have attribute 'ImageType'.'

According the DICOM standard, MR images' imageType tag is a type 1 attribute, it would be great if add try-except though.

Regards,

YingLi

mgxd:
@yinglilu 

thanks for reporting this, we should be making many fields optional. 

Would you want to try submitting the PR?

mgxd:
closed with b1e17d1

Issue #98, opened 2017-10-03
TR, TE ValueError


bkraft4257:
I am using heudiconv to convert my own DICOM files created from a Siemens scanner.  I have encountered problems with some of the series that Siemens creates where the TR and TE values are strings and not floats.  This results in a ValueError. 

(python2) flip ➜  synergy heudiconv -d '{subject}/{session}/data/dicom/20170914/*/*.DCM' --minmeta -b -c none -f convertall.py  -o ./output -s syn054 -ss 1

INFO: Need to process 1 study sessions
INFO: PROCESSING STARTS: {'session': '1', 'outdir': '/Users/bkraft/projects/hugenschmidt/synergy/output/', 'subject': 'syn054'}
INFO: Processing 1527 dicoms
INFO: Analyzing 1527 dicoms

Traceback (most recent call last):
  File "/usr/local/bin/heudiconv", line 2168, in <module>
    main()
  File "/usr/local/bin/heudiconv", line 2160, in main
    return _main(args)
  File "/usr/local/bin/heudiconv", line 2023, in _main
    min_meta=args.minmeta)
  File "/usr/local/bin/heudiconv", line 1335, in convert_dicoms
    grouping=None,  # no groupping
  File "/usr/local/bin/heudiconv", line 476, in group_dicoms_into_seqinfos
    TR = float(dcminfo.RepetitionTime) / 1000.
ValueError: could not convert string to float:

I modified the code on line 476 and 480 in heudiconv.py from 

     except AttributeError:

to 

    except (AttributeError, ValueError):

This resolved the error and allowed heudiconv to complete. 

I apologize for not issuing a pull request. I am still learning how to use GitHub and the process of collaborating on open source projects. 

Bob

mgxd:
Hi Bob, thanks for the heads up - just curious, what is the value of `dcm.RepetitionTime` for your scan?

bkraft4257:
I  read the DICOM header with pydicom.  The TR and TE are just empty strings. 

>>> ds = dicom.read_file("000001.DCM")

(0018, 0023) MR Acquisition Type                 CS: '2D'
(0018, 0024) Sequence Name                       SH: 'tfi2d1_113'
(0018, 0025) Angio Flag                          CS: 'N'
(0018, 0050) Slice Thickness                     DS: '8'
(0018, 0080) Repetition Time                     DS: ''
(0018, 0081) Echo Time                           DS: ''
(0018, 0082) Inversion Time                      DS: '145'
[...]

Issue #87, opened 2017-09-16
Refactor heudiconv script into proper python package


satra:
the heudiconv script has grown really large with many functions, missing tests. we need to refactor it into a set of smaller utility pieces.

@mgxd - can you take this on? feel free to discuss the pieces under this issue.

mgxd:
[WIP branch](https://github.com/mgxd/heudiconv/tree/rf/package), will submit WIP PR once functionality is equal to current master

Issue #84, opened 2017-09-16
saving of hidden files that are not updated with changes to heuristic


poldrack:
hi - it appears that information is saved into the .heudiconv folder that will prevent the heuristic from being rerun on the same data, even if the heuristic file changes.  I spent several hours trying to debug a problem before realizing this.  saving info to a hidden folder seems like a suboptimal solution, but if you are going to do it, please clear it out whenever the heuristic changes, or at least save it within the newly generated subject directory so that when that directory is deleted (which is necessary in order to rerun the job) then the hidden directory would be deleted as well.

satra:
noted - we did a big merge with a large number of changes. time to clean up the changes from this.

yarikoptic:
This I think bitten me as well a few times. But I just found peace of mind with it as a feature ;-)

satra:
@poldrack - as we move forward with this codebase, we will be bringing in a few more things.

- change in `heuristic | heudiconv | environment` will trigger a regeneration. at present there are two primary sources of difference: heuristic, dcm2niix but changes could come from heudiconv itself as well.
- integration of dcmstack into nibabel and/or extraction of features via other means

i think the biggest drawback from a testing standpoint is the absence of sufficient varied dicom datasets.

yarikoptic:
I think that the problem is also due to trying to address two target use-cases:

- original two stage: 
  1. generate .edit.txt with mapping to be adjusted if needed
  2. run actual conversion
- single stage:
  1. convert what is given without given a chance to adjust anything

The "single stage" is kinda my use-case which I added later, while "original two stage" is what Satra+Michael envisioned/implemented originally.  And ATM it is  not clear when heudiconv performs in two stage or a single stage scenario.  
In my case, I am trying to avoid silent 're-conversion' since it could lead to spurious files laying around left from previous attempts with previous heuristic and inconsistent conversion altogether. That is why added explicit `--overwrite` flag to instruct when I explicitly want to overwrite previous results of the conversion, and otherwise to crash if any files tries to override any previous present file (be it a converted or a heuristic file).

I think we should sit down (during code rodeo) and outline major use-cases/workflows we would like to see supported, and make them clear(er) presented in API/documentation.

satra:
@yarikoptic - sounds like a good discussion topic for code rodeo.

yarikoptic:
[...]

PR #101, opened 2017-10-12
[WIP] RF into python package

mgxd:
As requested in #87 

Quick list of things still remaining:
* [x] BIDS-compatible subject names are not being output (still include nonalphanumeric chars)
* [x] If converted files already exist, will not rerun conversion (#84) - allow flag to control this
* [x] Logging verbosity reduced (is this a good thing?)
* [x] Reduce DICOM field dependencies - run with bare minimum information (#98, #99, #80)
* [x] Testing

Other things to keep in mind for in the futre:
Write out BIDS `IntendedFor` for fieldmaps if shim matches func/dwi/etc.
Re-add cluster distribution support

I've also left some comments while working on this - feel free to address while reviewing

@yarikoptic would appreciate some feedback if you can try running with some of your data!


heudiconv/utils.py
```diff
@@ -97,12 +97,12 @@ def dec(obj):
...
 def anonymize_sid(sid, anon_sid_cmd):
-    anon_sid = sid
     if anon_sid_cmd is not None:
         from subprocess import check_output
         anon_sid = check_output([anon_sid_cmd, sid]).strip()
-        lgr.info("Annonimized sid %s into %s", sid, anon_sid)
-    return anon_sid
+        lgr.info("Anonymized sid %s into %s", sid, anon_sid)
+        return anon_sid
+    return
```
yarikoptic: why it should return None???

yarikoptic:
Thanks @mgxd . I guess this (modularization) should have been done long ago, but I was stopped (can't find a reference now ;) ) - the reason was to have full copy of heudiconv (script) along with the converted data. 
Now indeed It might be a good moment since virtually no open PRs, and whatever changes I have (and didn't yet PRed) are minimal so I could find where to plug them in.  But it should be done quickly.  I would advise to not introduce any additional fixes/extensions (I see in TODOs) -- changes should really concern only with making it a module.
And please do not change/fix functionality unless 100% sure that it would be the right thing to do ;)  I will try to review in upcoming days

setup.py
```diff
@@ -6,64 +6,49 @@
...
             'heudiconv_heuristics':
-                findsome('heuristics', {'py'})
-        }
-    )
+                findsome('heuristics', ('py'))},
+        )
+
+
+if __name__ == '__main__':
+    main()
```
yarikoptic: rarely (if ever) saw such approach -- do you expect setup to be used as a module somehow?  if not, why bother with all the indentation/moving into main?

heudiconv/dicoms.py
```diff
@@ -0,0 +1,507 @@
...
+    embedfunc.inputs.min_meta = min_meta
+    if bids:
+        embedfunc.inputs.bids_info = load_json(op.abspath(outname_bids))
+    else:
+        embedfunc.inputs.bids_info = None
+    embedfunc.inputs.force = True
+    embedfunc.base_dir = tmpdir
+    cwd = os.getcwd()
+    try:
+        # MG - still a bug?
```
yarikoptic: I think, it is safe to remove this comment.  It was due to retransmission with fixed up record.

heudiconv/bids.py
```diff
@@ -0,0 +1,323 @@
...
+    for jsonfile in json_files:
+        json_ = load_json(jsonfile)
+        # sanitize!
+        for f1 in ['Acquisition', 'Study', 'Series']:
+            for f2 in ['DateTime', 'Date']:
+                json_.pop(f1 + f2, None)
+        # TODO:  should actually be placed into series file which must
+        #        go under annex (not under git) and marked as sensitive
+        # MG - Might want to replace with flag for data sensitivity
+        # related - https://github.com/nipy/heudiconv/issues/92
```
yarikoptic: "random" shifting of the Date should be done per dataset (so not exactly random - subsequent calls should reuse the same offset which shouldn't be shared later), and could be a nice (optional) feature indeed.  Let's keep this comment in mind, so later on some proper "--anonimize-dates" could get other values (e.g. "random-offset") not only the current "strip" ;)

heudiconv/bids.py
```diff
@@ -0,0 +1,323 @@
...
+            # that
+            raise ValueError("There must be no dates in .json sidecar")
+        #json.dump(json_, open(jsonfile, 'w'), indent=2)
+        save_json(jsonfile, json_, indent=2) # ensure this does same as above
+
+    # Load the beast
+    seqtype = op.basename(op.dirname(jsonfile))
+
+    # MG - want to expand this for other _epi
+    # possibly add IntendedFor automatically as well?
```
yarikoptic: thought about it, but not sure if we could easily pull it off since it is not 100% clear what file(s) fmap is pertinent to.  We could do matching by size/encoding directions etc, but it would be truly a heuristic ;)  may be should become some kind of an option ("--indend-fmaps-for-all-epis"? ;))

heudiconv/convert.py
```diff
@@ -0,0 +1,394 @@
...
+
+    shutil.copy(heuristic.filename, idir)
+    ses_suffix = "_ses-%s" % ses if ses is not None else ""
+    info_file = op.join(idir, '%s%s.auto.txt' % (sid, ses_suffix))
+    edit_file = op.join(idir, '%s%s.edit.txt' % (sid, ses_suffix))
+    filegroup_file = op.join(idir, 'filegroup%s.json' % ses_suffix)
+
+    # MG - maybe add an option to force rerun?
+    # related issue : https://github.com/nipy/heudiconv/issues/84
+    if op.exists(edit_file) and overwrite == True:
```
yarikoptic: I got to like Pythonic generic programming semantics, so I would have just said `and overwrite`... why bother with `overwrite == True`?

heudiconv/dicoms.py
```diff
@@ -0,0 +1,507 @@
...
+      Result tarball
+    """
+    # if tempdirs:
+    tmpdir = tempdirs(prefix='dicomtar')
+    # else:
+    #     tmpdir = mkdtemp(prefix='dicomtar')
+    outtar = out_prefix + '.dicom.tgz'
+    if op.exists(outtar): # and not global_options['overwrite']:
+        raise RuntimeError("File %s already exists, will not override, ensure "
+                           "unique scan names" % outtar)
```
yarikoptic: so "overwrite" mode is disabled now?

heudiconv/dicoms.py
```diff
@@ -0,0 +1,507 @@
...
+and there was a screw up for that A
+
+/mnt/btrfs/dbic/inbox/DICOM/2017/01/28/A000203
+        StudySessionInfo(locator='Wheatley/Beau/1007_personality', session=None, subject='sid000138') 16 sequences
+        StudySessionInfo(locator='Wheatley/Beau/1007_personality', session=None, subject='a000203') 2 sequences
+
+
+in that one though
+        """
+        # if global_options['overwrite'] and op.lexists(scaninfo):
```
yarikoptic: also overwrite mode is disabled...?

heudiconv/cli/run.py
```diff
@@ -0,0 +1,326 @@
...
+    if args.files and args.dicom_dir_template:
+        raise ValueError("Specify files or dicom_dir_template, not both")
+
+    if args.files and args.sids:
+        raise ValueError("")
+
+    if args.debug:
+        setup_exceptionhook()
+
+    # MG QUESTION - why not just use args.overwrite
```
yarikoptic: I guess I got tired of passing a dozen options deep inside the functions and decided to use this global set of options... see few points where I have noted that "overwrite" mode of operation was disabled.
That is partially why I have noted that for "make it a module" better to minimize amount of functional changes and concentrate on "partitioning" of the functionality

heudiconv/convert.py
mgxd: whoops...guess i was thinking verbose atm

heudiconv/dicoms.py
mgxd: nope - just forgot to add the argparse option to this

heudiconv/dicoms.py
mgxd: ...and here

setup.py
mgxd: I was mainly modeling after `nipype`'s approach - however I don't feel strong about either way, I can change this if you do

heudiconv/utils.py
mgxd: the problem is really [this line](https://github.com/mgxd/heudiconv/blob/rf/package/heudiconv/cli/run.py#L276) which forced `anon_sid` to be set, and disrupted naming in `prep_conversion`

heudiconv/dicoms.py
mgxd: It doesn't seem like did overwrite actually did anything here yet...

mgxd:
@yarikoptic 
The `monitor` tests seem to be failing, but I'm not sure if these were ever successfully running..
Besides that, most tests are passing, but I'm having some problems getting the `datalad` tests to run.

bin/heudiconv
```diff
@@ -1,2160 +0,0 @@
...
-so it seems that there is a filename collision so it tries to save into the same file name
-and there was a screw up for that A
-
-/mnt/btrfs/dbic/inbox/DICOM/2017/01/28/A000203
-        StudySessionInfo(locator='Wheatley/Beau/1007_personality', session=None, subject='sid000138') 16 sequences
-        StudySessionInfo(locator='Wheatley/Beau/1007_personality', session=None, subject='a000203') 2 sequences
-
-
-in that one though
-        """
```
yarikoptic: I have finally caught this issue and fixed for it in #105 . Sorry for causing more conflicts/more work

yarikoptic:
Sorry, we caused you some conflicts to resolve and adopt into your modularization effort.

Before anything,  Dear Father (@satra and @mih), could you please give your blessing upon heudiconv to become the right child of the Python ecosystem?  Its code base grew up forbidding sane management of everything within a single file.  If really needed to make it into a single file may be some approach (like of impressive project) to later glue them all up would work?  but IMHO it becomes less and less feasible/needed as time goes. 

Re monitor tests... well indeed, seems to be left there although should have been gone to tests/... may be while at it merge/migrate under tests https://github.com/dartmouth-pbs/heudiconv/pull/20 ?

Re datalad -- should work!  just point me to the specific failure(s) and we will get it resolved

mgxd:
Okay, [tests are now passing](https://travis-ci.org/mgxd/heudiconv).

One tweak I wanted to bring up was commit 59d71c7, which was dropped because `compress_dicoms` no longer raises a `RuntimeException`, but simply returns `None` when `overwrite` isn't set.

satra:
@mgxd - are you going to resolve the conflicts?

@yarikoptic - @mgxd is doing this with my blessing.

yarikoptic:
@satra great, thanks!  @mgxd - best to be uniform everywhere and throw an exception, may be a new dedicated one, if something is attempted to be overridden

heudiconv/bids.py
satra: technically the shim and encoding direction (at least one) has to remain the same. i think some of this could be done automatically. perhaps we can generate this as a function that people can reuse in the heuristic. in fact it may be useful, now that this is a proper python package to create a set of utility functions that simplify heuristics. also would be useful to ensure that json files can be changed via heuristics as well.

satra:
@yarikoptic, @mgxd - i think this should be merged as soon as we can before making additional changes. will help others to contribute if we have this structure in place.

so perhaps if tests pass, that's a good point for merger at the moment.

yarikoptic:
I will try to give it a try later today on some local "real case" data to see if anything changed in terms of behavior for me.  Indeed, should be merged as soon as possible, but we should also go through the code to identify possible issues.
Also, anyone could try with our heuristic and data shared for the QA sessions from http://datasets.datalad.org/?dir=/dbic/QA (`sourcedata/` contains tarballs with original dicoms).  I use `--bids` and `--datalad` options and typically go through one accession at a time

satra:
@yarikoptic - that should perhaps be a specific test run on travis. @mgxd - can you add yarik's suggestion as a test?

mgxd:
i'm having problems installing a datalad repo within a `/tmp` directory, any ideas @yarikoptic ? Runs fine in other partitions
```python
from datalad import api
import os
from tempfile import mkdtemp


subject = 'sub-sid000143'

tmpdir = mkdtemp()
os.chdir(tmpdir)
print(os.getcwd())
print('attempting to use datalad install')

api.install('///dbic/QA')
print('it has been called')

assert os.path.exists('./QA') == True
```

yarikoptic:
How does it fail? Which datalad version do you have?

mgxd:
Datalad version `0.9.1` - the script fails at the assertion, but only when the current directory is a temp directory - can you try this out and verify?

yarikoptic:
Thanks for your patience @mgxd!  Long(er) story: https://github.com/datalad/datalad/issues/1940

Quick workaround -- use full path(s) or if you sure using datalad, use its datalad.utils.chpwd instead of os.chdir.  Here is an adjusted script
```shell
$> cat test.py
from datalad import api
from datalad.utils import getpwd
import os
from tempfile import mkdtemp

subject = 'sub-sid000143'

tmpdir = mkdtemp()
os.chdir(tmpdir)
print(os.getcwd())
print(getpwd())
print('attempting to use datalad install')

# Due to https://github.com/datalad/datalad/issues/1940
# use full path.  Alternative could be to use datalad.utils.chpwd
api.install(path=os.path.join(tmpdir, 'QA'), source='///dbic/QA')
print('it has been called')

assert os.path.exists('./QA') == True
```

satra:
@yarikoptic - shall we merge?

yarikoptic:
oh damn, didn't test it yet (will do quickly now)... sure thing we can merge and fix up later.  Did you test it on your cases?

satra:
i'm assuming @mgxd did :)

yarikoptic:
ok -- got to try it to immediately run into the changed API:
```
$> heudiconv --datalad --bids -f /home/yoh/picts/mris/heudiconv/heuristics/dbic_bids.py -o /tmp sourcedata/sub-qa/ses-20170828
usage: heudiconv [-h] [--version]
                 [-d DICOM_DIR_TEMPLATE | --files [FILES [FILES ...]]]
                 [-s [SUBJS [SUBJS ...]]] [-c {dcm2niix,none}] [-o OUTDIR]
                 [-a CONV_OUTDIR] [--anon-cmd ANON_CMD] -f HEURISTIC_FILE [-p]
                 [-ss SESSION] [-b] [--overwrite] [--datalad] [--dbg]
                 [--command {treat-json,ls,populate-templates}]
                 [-g {studyUID,accession_number}] [--minmeta] [-q QUEUE]
                 [--sbargs SBATCH_ARGS]
heudiconv: error: unrecognized arguments: sourcedata/sub-qa/ses-20170828
```
so there is now argument to specify input directory/files?

mgxd:
@yarikoptic yes - I made `dicom dir template` and `files` mutually exclusive so wanted to have a unique flag for each

yarikoptic:
ok. I can work with that... ;)
So, I did conversion on one of the QA runs and the only difference which I could contest is 

```patch
--- sub-qa/func/sub-qa_task-rest_acq-p2Xs4X35mm_events.tsv        2017-11-03 16:37:37.134637912 -0400
+++ /home/yoh/.tmp/XTYhjr_sub-qa_task-rest_acq-p2Xs4X35mm_events.tsv      2017-11-03 16:54:25.253157100 -0400
@@ -1 +1 @@
-onset    duration        trial_type      response_time
\ No newline at end of file
+onset    duration        trial_type      TODO -- fill in rows and add more tab-separated columns if desired
```

I am ok to have `response_time` (we could even make it smart and not to add it if task==`rest` ;)). But I would like to keep the "TODO" entry as well -- this way people could just `git grep TODO` to find the files  they need to get tuned up (or removed).

Now, I will just do an "incremental" addition and if no gotchas - I will merge. Great job. If anything pops up - we could resolve it as we go.

yarikoptic:
btw, I didn't mention that I very much like that tool datalad (and git-annex) which allow me quickly add/drop data files from local drive which is nearly full ;-)

yarikoptic:
- subject is `None` in the commit message now
- heudiconv_monitor needs entry_point definition as well I guess
- `... yet to finish reviewing`

heudiconv/cli/run.py
```diff
@@ -0,0 +1,313 @@
...
+                             dicoms,
+                             sid,
+                             args.anon_cmd,
+                             args.converter,
+                             args.session,
+                             args.with_prov,
+                             args.bids)
+            continue
+
+        anon_sid = anonymize_sid(sid, args.anon_cmd) if args.anon_cmd else None
```
yarikoptic: here is where sid is reset to None if no anon_cmd.  I will return it back to `sid` value instead of None

satra:
woohoo!

heudiconv/cli/run.py
mgxd: @yarikoptic ah yeah I remember this change now -
if we define `anon_sid` as sid further down the pipeline:
https://github.com/nipy/heudiconv/blob/master/heudiconv/cli/run.py#L281

then conversion will use the pre bids cleaned name
https://github.com/nipy/heudiconv/blob/master/heudiconv/convert.py#L71

in retrospect there are better ways to handle it, but currently this will break bids compatibility with sids that contain non-alphanumeric characters

yarikoptic:
Please file a new issue, ideally with a test to demo, so we don't forget that I broke something to fix it for the past behavior

PR #102, opened 2017-10-14
BF: do not chmod files using hardcoded mode number

yarikoptic:
It was resulting in ignoring umask etc, and making files non-readable by others where they should have been.  Now it should be more consistent and only make files not writable, and if making writeable only for levels where they are also readable.

While at it also marked a test to be skipped if there is no datalad


yarikoptic:
if no objections, I would like to get it merged soon ;)

PR #105, opened 2017-10-19
BF: do not set readonly on the sidecar phasediff.json if not yet set read-only

yarikoptic:
Finally I got annoyed with that issue enough and figured it out ;)  Due to need to specify those EchoTime[12] "manually", we turned them "read-only" way too early, which lead the "embedder" to fail to embed later in the process.  Now tune up of EchoTime's would only change permissions if file is already set read-only


mgxd:
@yarikoptic I've included this in #101

PR #113, opened 2017-11-06
BF: bidify sid and remove lingering dicom tempdirs

mgxd:
* revert `anon_sid` default to `None` in `run.py`, but log `sid` in datalad message
  * currently, setting `anon_sid` is superceeding the BIDS cleaned `sid`
* remove lingering dicom tempdir when running with `-c none`
  * add pathlib as a requirement

### DIFF
--- a/heudiconv/cli/run.py
+++ b/heudiconv/cli/run.py
@@ -267,8 +267,8 @@ def process_args(args):
         # the outdir -> study_outdir datasets if not yet there
         if args.datalad:
             from ..external.dlad import prepare_datalad
-            anon_sid = sid if not anon_sid else anon_sid
-            dl_msg = prepare_datalad(anon_study_outdir, anon_outdir, anon_sid,
+            dlad_sid = sid if not anon_sid else anon_sid
+            dl_msg = prepare_datalad(anon_study_outdir, anon_outdir, dlad_sid,
                                      args.session, seqinfo, dicoms, args.bids)
 
         lgr.info("PROCESSING STARTS: {0}".format(
--- a/heudiconv/cli/run.py
+++ b/heudiconv/cli/run.py
@@ -255,7 +255,7 @@ def process_args(args):
                              args.bids)
             continue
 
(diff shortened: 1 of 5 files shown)


mgxd:
ping @satra @yarikoptic

heudiconv/cli/run.py
```diff
@@ -255,7 +255,7 @@ def process_args(args):
                              args.bids)
             continue
 
-        anon_sid = anonymize_sid(sid, args.anon_cmd) if args.anon_cmd else sid
+        anon_sid = anonymize_sid(sid, args.anon_cmd) if args.anon_cmd else None
```
yarikoptic: I thought I was the one who reverted it to be 'sid' as it used to be.  So, does it mean now that I have to provide sid even though it might have been deduced already from the data?  note that we are in the loop here going through extracted studies, possibly with different sids.  and logically -- if there is no annonimization command provided, it just means that sid is already good enough, not that it is None

mgxd: @yarikoptic You shouldn't need to pass any value, since sid is derived from `get_study_sessions`

an example of the problem I'm running into: a subject id is `PHANTOM_1_3`, and we don't want to anonymize it.
```
anon_sid = anonymize_sid(sid, args.anon_cmd) if args.anon_cmd else sid
```
will set `anon_sid = 'PHANTOM_1_3'`

then [downstream](https://github.com/nipy/heudiconv/blob/master/heudiconv/cli/run.py#L282) this `anon_sid` is passed into the conversion. 

[`sid` is made bids compliant](https://github.com/nipy/heudiconv/blob/master/heudiconv/convert.py#L69-L70) (PHANTOM13), but then because `anon_sid` was defined, [it is used for the conversion](https://github.com/nipy/heudiconv/blob/master/heudiconv/convert.py#L137).

This leads to complications with generated bids files, likely because we separate by underscores.

If this change is no-good, we can move the checks within the `prep_conversion` function to before the `anonymize_sid` call here, and it should have the same result (?)

PR #117, opened 2017-11-29
fix: update dockerfile and correctly install non-developer

mgxd:
fixes non-developer installation problems, as well as a fresh new dockerfile


mgxd:
we should also increase heudiconv's version to differentiate pre/post refactor, perhaps a 0.4.1 or 0.5

PR #129, opened 2018-01-05
Fixes from DBIC field-test site and code rodeo coding session (AKA reaching convergence)

yarikoptic:
Unfortunately we are still using the codebase which is not yet refactored as in the master (planing to look at switching may be while at code rodeo), some fixes piled up and provided here:
- no git-annex shared access should be configured.  That one is leading to weak data protection so not enabled (files could be then modified in place, ruining assumptions/checksums)
- few extensions/fixes in heuristics (for SE sequences for AP/PA fieldmap correction etc)
- allow to override the subject id

If possible, please merge, and not just pick the changes into master since otherwise it makes it a bit more difficult for me to track what was already adopted into master and what not.

# Further work during rodeo

- A change in heuristic will trigger a regeneration of the conversion table
- `--overwrite` guards are reincarnated -- if the converted files already exist and no `--overwrite` is provided, the run will error out

# Doing comparisons against our "production" version

Here I outline critical differences detected/resolved when checked
- [x] session got lost while trying on [///dicoms/dartmouth-phantoms/bids_test4-20161014/phantom-1](http://datasets.datalad.org/?dir=/dicoms/dartmouth-phantoms/bids_test4-20161014/phantom-1) which defined session in the scout sequence
- [x] possible to override the subject
- [x] reincrnate logging -- now it is just a NullLogger without any way to control from env vars like it did before
- [x] fix up for the tests
```
tests/test_heuristics.py:58: in test_dbic_bids_largely_smoke
    runner(args)
E   Failed: DID NOT RAISE <type 'exceptions.RuntimeError'>
```


yarikoptic:
I could swear that we (at code rodeo with @mgxd) ironed out that usecase and tests for validating that we do not override files but as travis and local testing shows -- they are still there, so I am a bit confused and need to address them.

yarikoptic:
@mgxd do you remember changes related to propagation of session information from the heuristic?  seems got completely lost (added description in the summary above)

yarikoptic:
@mgxd I have introduced fixes which seems have brought functioning to fulfill my usecase.   Yet to test a bit more but seems to do "The right thing". Please review the changes or the latest commits for changes. Ideally there should be tests for them, but I deferred that until later (#143)

yarikoptic:
@mgxd and @satra -- I think I am almost done with tune ups on top of master to bring heudiconv into operation consistent with our deployment of previous (released) version at DBIC.  It would be great if you review the changes and also test on your usecases before we merge.  I will also test more on our usecases since I still keep running into differences (I wish we had even more tests ;-) )

mgxd:
@yarikoptic should be able to try this out today

yarikoptic:
Great, thanks in advance @mgxd .  I hope to add few more tests, but so far so good in terms of functionality, so shouldn't change ;-)

yarikoptic:
Thanks! I will merge then and we will fix up the rest after

Cheers

PR #130, opened 2018-01-05
RF+ENH: Move heuristics/ to become a heudiconv.heuristics submodule

yarikoptic:
It is really inconvenient IMHO to need to specify heuristics files if it is a "standard" heuristic, so imho we could just keep them as submodules and ship along in an easy to find location etc

Ok, accomplished:
- [x] `dbic_bids` -> `reproin`
- [x] make possible to just specify heuristic name, e.g. `-f dbic_bids` and then heuristic would be looked up among the "standard" ones
- [x] also it might allow to present available heuristics in case if argument gives some name which isn't known.  e.g.
```shell
$> heudiconv --command heuristics              
- banda-bids: 
- bids_with_ses: 
- cmrr_heuristic: 
- convertall: 
- example: 
- multires_7Tbold: 
- reproin: (AKA dbic-bids) Flexible heuristic to establish BIDS DataLad datasets hierarchy
- studyforrest_phase2: 
- uc_bids: 
(dev) 3 5439 [1].....................................:Fri 02 Feb 2018 06:01:09 PM EST:.
hopa:~/picts/mris/heudiconv-master
$> heudiconv --command heuristic-info -f reproin | head -n 10
(AKA dbic-bids) Flexible heuristic to establish BIDS DataLad datasets hierarchy

Initially developed and deployed at Dartmouth Brain Imaging Center
(http://dbic.dartmouth.edu) using Siemens Prisma 3T

## Dataset ownership/location

Datasets will be arranged in a hierarchy similar to how study/exam cards are
arranged at the scanner console.  You should have
...
```
 and actually provide an additional (we have already a few) command `heuristics` which could also present that list.
Those descriptions could be specified in the first line of the corresponding file `__doc__`.

Thought first to run by you guys @nipy/team-heudiconv


yarikoptic:
- those heuristics which aren't to be used should be removed. if they are shipped along -- they should be "exposed" IMHO
- alternative sources for heuristics -- sure, could be another PR
- dbic -> repronim, yeap, will be here

yarikoptic:
ok, done -- sits on top of #147 ATM

```shell
$> heudiconv --command heuristics              
- banda-bids: 
- bids_with_ses: 
- cmrr_heuristic: 
- convertall: 
- example: 
- multires_7Tbold: 
- reproin: (AKA dbic-bids) Flexible heuristic to establish BIDS DataLad datasets hierarchy
- studyforrest_phase2: 
- uc_bids: 

$> heudiconv --command heuristic-info -f reproin | head -n 10
(AKA dbic-bids) Flexible heuristic to establish BIDS DataLad datasets hierarchy

Initially developed and deployed at Dartmouth Brain Imaging Center
(http://dbic.dartmouth.edu) using Siemens Prisma 3T

## Dataset ownership/location

Datasets will be arranged in a hierarchy similar to how study/exam cards are
arranged at the scanner console.  You should have
```

yarikoptic:
Ok, this over is done and ready for more review
@satra @mgxd

README.md
```diff
@@ -43,11 +43,10 @@ Debian-based systems we recommend to use
 
 ### From source
 
-You can clone this directory and do a `make install`
+You can clone this directory and use `pip install .` (with `--user`, `-e`
```
mgxd: this won't include `dcmstack`, perhaps `pip install -r requirements.txt`?

heudiconv/cli/run.py
```diff
@@ -148,9 +158,11 @@ def get_parser():
                         'DICOMs to anonymized IDs. Such command must take a '
                         'single argument and return a single anonymized ID. '
                         'Also see --conv-outdir')
-    parser.add_argument('-f', '--heuristic', dest='heuristic_file',
-                        required=True,
-                        help='python script containing heuristic')
+    parser.add_argument('-f', '--heuristic', dest='heuristic',
+                        # some commands might not need heuristic
```
mgxd: we can remove the comment and just raise an error if required but not specified

yarikoptic: or may be we would refactor to always have a command... who knows ;-)  comment was left primarily for myself I guess, since initially it was a required option.

README.md
yarikoptic: THANKS!  Added `dcmstack` into requires and thus would be included if `pip install .[all]`. Adjusted README to recommend ppl use `[all]` flavor. See 7f00b1a54ef38ae3cc7cc4cd059e10331ba97654

yarikoptic:
yes, release should come (very) soon!  May be we test drive it for a few days on real cases and then kick out 0.5 end of the next week or so, deal? ;)

heudiconv/info.py
```diff
@@ -24,6 +24,7 @@
 
 EXTRA_REQUIRES = {
     'tests': TESTS_REQUIRES,
+    'extras': ['dcmstack'],
```
mgxd: sadly not on pypi

how about adding a `dependency_links` option to `setuptools.setup` with the gitURL?

yarikoptic: eh, yesterday that I have seen travis to fail I remembered that patched dcmstack is needed.  we have it patched in neurodebian IIRC but indeed, we cannot just add it from pypi :-/  I will try to reach @moloney  again, hope that he is alright (previous attempts went without a reply)
I was not aware of `dependency_links`, is it present in olderish setuptools as well?

yarikoptic: meanwhile pushed a commit removing it from requires to get tests passing.  We could decide how to proceed on that one later (situation wouldn't be worse than before)

yarikoptic:
ok, since "approved" and passes tests again, will merge for a final round of tests

Issue #120, opened 2017-12-11
outtype 'dicom' creates invalid links


cni-md:
Running heudiconv with outtpye=('dicom') will create only links to the dicom files which are in /tmp/heudiconv/... (lines 881ff)

After conversion the tmp folder will be deleted and the links are invalid.
I replaced this pice of code with shutil.copyfile(filename, outfile)

mgxd:
@cni-md would you mind sharing your fix in a PR? 

related to #103

cni-md:
@mgxd  Is this still a problem? I see, that heudiconv was completely rewritten and I still use the "old" version. 

It is now in convert.py line 331:
# TODO: add option to enable hardlink?
I removed the if / else and replaced it with shutil.filecopy() 

I can write a PR, but the symlinks should work at least for the programmer ;)  And I do not know why this is working for him and not for me.

PR #135, opened 2018-01-19
quick and dirty symlink fix for #120

cni-md:
@mgxd Please check the itention of the original code.

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -324,10 +324,11 @@ def convert_dicom(item_dicoms, bids, prefix,
             outfile = op.join(dicomdir, op.basename(filename))
             if not op.islink(outfile):
                 # TODO: add option to enable hardlink?
-                if symlink:
-                    os.symlink(filename, outfile)
-                else:
-                    os.link(filename, outfile)
+#                if symlink:
+#                    os.symlink(filename, outfile)
+#                else:
+#                    os.link(filename, outfile)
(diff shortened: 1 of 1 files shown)


heudiconv/convert.py
```diff
@@ -324,10 +324,11 @@ def convert_dicom(item_dicoms, bids, prefix,
...
                 # TODO: add option to enable hardlink?
-                if symlink:
-                    os.symlink(filename, outfile)
-                else:
-                    os.link(filename, outfile)
+#                if symlink:
+#                    os.symlink(filename, outfile)
+#                else:
+#                    os.link(filename, outfile)
+                shutil.copyfile(filename, outfile)
```
yarikoptic: that would lead to huge performance impact on any sizeable dataset for no good reason really.  The fix here would be just to get a full path to the filename, not a relative path (I think that is what was broken, wasn't it?)

yarikoptic: oh, I misunderstood what issue we are solving.. #120 concerns about broken symlinks to dicom outputs, while I thought we are talking about failing "embedding" (works in my elderly version without a message at least)

cni-md: Yes, the link option was not working for me. On my system my dicom tar-ball is extracted to /tmp/..., and the conversion created absolute links in <outdir>/sub/func/... to this /tmp dir, but at the end the tmp dir is removed,my dicom files are gone, and I have only links to nowhere.

Issue #108, opened 2017-10-27
Error while using '--with-prov' flag


gregblt:
Hi,

When I'm trying to use heudiconv with the '-p' flag, I get the following error (it works properly without using this flag): 

INFO: Need to process 1 study sessions
INFO: PROCESSING STARTS: {'session': 'baseline', 'outdir': '/SPINE_BIDS_Test_server-side/nifti/', 'subject': '004'}
INFO: Processing 480 dicoms
INFO: Analyzing 480 dicoms
INFO: Generated sequence info with 3 entries
INFO: Doing conversion using dcm2niix
INFO: Converting /SPINE_BIDS_Test_server-side/nifti/sub-004/ses-baseline/anat/sub-004_ses-baseline_FLAIR (128 DICOMs) -> /SPINE_BIDS_Test_server-side/nifti/sub-004/ses-baseline/anat . Converter: dcm2niix . Output types: ('nii.gz',)
INFO: Executing node convert in dir: /tmp/heudiconvdcm23WAK6/convert
INFO: Running: dcm2niix -b y -z i -x n -t n -m n -f anat -o /tmp/heudiconvdcm23WAK6/convert -s n -v n /tmp/heudiconvdcm23WAK6/convert/1.3.12.2.1107.5.2.7.20107.4.0.1576580820461847.dcm
_**Traceback (most recent call last):
  File "/usr/local/bin/heudiconv", line 2197, in <module>
    main()
  File "/usr/local/bin/heudiconv", line 2189, in main
    return _main(args)
  File "/usr/local/bin/heudiconv", line 2052, in _main
    min_meta=args.minmeta)
  File "/usr/local/bin/heudiconv", line 1407, in convert_dicoms
    min_meta=min_meta)
  File "/usr/local/bin/heudiconv", line 911, in convert
    res = convertnode.run()
  File "/opt/conda/lib/python2.7/site-packages/nipype/pipeline/engine/nodes.py", line 372, in run
    self._run_interface()
  File "/opt/conda/lib/python2.7/site-packages/nipype/pipeline/engine/nodes.py", line 482, in _run_interface
    self._result = self._run_command(execute)
  File "/opt/conda/lib/python2.7/site-packages/nipype/pipeline/engine/nodes.py", line 613, in _run_command
    result = self._interface.run()
  File "/opt/conda/lib/python2.7/site-packages/nipype/interfaces/base.py", line 1130, in run
    prov_record = write_provenance(results)
[...]

PR #142, opened 2018-01-30
fix: various fixes left over from refactor

mgxd:
* missing function imports
* moved some variables to their proper scope
* some pep8 fixes as well

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -2,6 +2,7 @@ import os
 import os.path as op
 import logging
 import shutil
+import sys
 
 from .utils import (
     read_config,
@@ -66,6 +67,7 @@ def conversion_info(subject, outdir, info, filegroup, ses):
                 try:
                     files = filegroup[item]
                 except KeyError:
+                    PY3 = sys.version_info[0] >= 3
                     files = filegroup[(str if PY3 else unicode)(item)]
                 outprefix = template.format(**parameters)
                 convert_info.append((op.join(outpath, outprefix),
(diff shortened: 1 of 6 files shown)


heudiconv/convert.py
```diff
@@ -266,8 +268,8 @@ def convert(items, converter, scaninfo_suffix, custom_callable, with_prov,
                     tmpdir = tempdirs('dcm2niix')
 
                     # run conversion through nipype
-                    res = nipype_convert(item_dicoms, prefix, with_prov, bids,
-                                         tmpdir)
+                    res, prov_file = nipype_convert(item_dicoms, prefix, with_prov,
+                                                    bids, tmpdir)
```
yarikoptic: is that something new in nipype 1.0 that it would return both?  I am receiving just 
```
-> bids_outfiles = save_converted_files(res, item_dicoms, bids,
(Pdb) p res
<nipype.interfaces.base.InterfaceResult object at 0x7f1b4b5f3350>
```
and this line even test covered, so not quite sure how tests manage(d) to pass etc

heudiconv/convert.py
```diff
@@ -382,9 +387,18 @@ def nipype_convert(item_dicoms, prefix, with_prov, bids, tmpdir):
...
     convertnode.inputs.out_filename = op.basename(op.dirname(prefix))
-    convertnode.inputs.terminal_output = 'allatonce'
     convertnode.inputs.bids_format = bids
-    return convertnode.run()
+    eg = convertnode.run()
+
+    # prov information
+    prov_file = prefix + '_prov.ttl' if with_prov else None
+    if prov_file:
+        safe_copyfile(op.join(convertnode.base_dir,
```
yarikoptic: so we need a test which runs with prov -- this clause isn't covered

mgxd: yes, but currently i believe this option is broken with latest nipype (#108)

mgxd: so looks like prov <=1.5.0 is fine (nipy/nipype#2240), but we should complain if user's version is higher

yarikoptic: travis is probably fine/has an old one ;-)

PR #153, opened 2018-02-10
RF: minors -- "ls" should print, and comparator would use just heuristic name not path

yarikoptic:


yarikoptic:
ok, no objections has been heard, let's proceed

PR #155, opened 2018-02-13
Fixed spaces problem

pvelasco:
I think the two lines I modified are supposed to execute _inside_ loop, rather that at the end of the loop.  That is, execute for every "item", rather than just for the last one.


yarikoptic:
that is pity we don't have a dedicated unittests battery for the `convert` call and actually it seems only a single heuristic makes use of this callback functionality... we should add something, merging for now as is without dedicated test

yarikoptic:
thank you @pvelasco

PR #156, opened 2018-02-13
install python-pathlib on travis

yarikoptic:
we use it so better install via apt on travis


yarikoptic:
forget about it ... apparently no debian pkg at that elderly point and backport for elderly 12.04 would need a bit of work, so I wouldn't care

PR #159, opened 2018-02-21
Allow to override locator path from the cmdline

yarikoptic:
also I started to dislike the name "locator", so I am accepting suggestions.  
It is the path under output dir, which could be used to define a hierarchy of datasets, so in --datalad mode every level would become a super dataset.  in ReproIn heuristic we use "PI/StudentOrAssistant/Study" hierarchy.


yarikoptic:
ok, no objectsion == everyone loves it ;)

PR #160, opened 2018-02-22
ENH: reincarnate the paranoid test of a json tuned up with regular expressions

### LINKED ISSUE #104: bring back an assert for correctness of regex-tuned json for the paranoid Yarik
In 12e94ea8533d3b49c2c733a313f6baf507ce42e8 it was commented out.

@satra , was the issue that some floating point number changed its value/json serialization somehow?  do you happen to remember where/how did it happen, since I think it should've not happened (at least never happened to me).

that assert is kinda critical to have in one form or another to guarantee that ugly regex tuning of output didn't backfire

yarikoptic:
Fixes #104

### DIFF
--- a/heudiconv/utils.py
+++ b/heudiconv/utils.py
@@ -226,9 +226,16 @@ def json_dumps_pretty(j, indent=2, sort_keys=True):
     js_ = re.sub('  *("?[-+.0-9e]+"?)[ \n]*', r' \1', js_)
     # no spaces after [
     js_ = re.sub('\[ ', '[', js_)
-    j_ = json.loads(js_)
-    # Removed assert as it does not do any floating point comparison
-    #assert(j == j_)
+    # the load from the original dump and reload from tuned up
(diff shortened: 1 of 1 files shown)


yarikoptic:
@mgxd do you by any chance remember on which dataset you ran into such numeric difference?  now it should be all legit imho

satra:
@yarikoptic - can you try it one of your prisma scans without `--minmeta` ? we ran into it i believe with the `--minmeta` option on one of our scans.

yarikoptic:
ok, I could try **with** `--minmeta` since I actually didn't use it recently

But note that this one is different from before where I was comparing against original structure, not the one reloaded from the non-perturbed string dump

yarikoptic:
ok, ran on a sample dataset with this change with and without --minmeta, nothing crashed, so it must be solid ;)

once again -- if now assertion fails it would mean that regexes actually modify the data, which they shouldn't so a crash would be warranted.

PR #167, opened 2018-02-26
Make calls to import pydicom compatible with pydicom 1.0

pvelasco:
In the newly released pydicom 1.0 the name of the module is
now 'pydicom', so modify the code accordingly.

I'm also removing the import of pydicom and dcmstack inside
group_dicoms_into_seqinfos because I think it is redundant
(although I'm not 100% sure).

### DIFF
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -10,7 +10,12 @@ import csv
 from random import sample
 from glob import glob
 
-import dicom as dcm
+try:
+    # for pydicom < 1.0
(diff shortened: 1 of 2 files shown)


yarikoptic:
the situation is more grave here since dcmstack relies on old(er) name... while possibly a generic workaround is being cooked (https://github.com/pydicom/pydicom/pull/583) I will see if I could come up with not that painful butt-plug here

yarikoptic:
superseded by #169

Issue #150, opened 2018-02-08
Tests fail with nipype 1.0.0 because of Dcm2niix interface changes


mvdoc:
e.g., simple test

```python
def test_scout_conversion(tmpdir):
    args = (
        "-f reproin --files %s -o %s"
        % (TESTS_DATA_PATH, tmpdir)
    ).split(' ')
    print args
    runner(args)
```

results in

```terminal
(heudiconv) contematto@talete ~/github/heudiconv (fix/emptydir*) $ pytest -sv tests/test_heuristics.py::test_scout_conversion
======================================================================================================== test session starts ========================================================================================================
platform darwin -- Python 2.7.14, pytest-3.4.0, py-1.5.2, pluggy-0.6.0 -- /Users/contematto/anaconda2/envs/heudiconv/bin/python
cachedir: .pytest_cache
rootdir: /Users/contematto/github/heudiconv, inifile: pytest.ini
collected 1 item                                                                                                                                                                                                                    

tests/test_heuristics.py::test_scout_conversion ['-f', 'reproin', '--files', '/Users/contematto/github/heudiconv/tests/data', '-o', '/private/var/folders/gs/4v1cn8mj4n52qrr9r6xv4ckm0000gn/T/pytest-of-contematto/pytest-13/test_scout_conversion0']
INFO: Running heudiconv version 0.4
INFO: Analyzing 1 dicoms
INFO: Filtering out 0 dicoms based on their filename
INFO: Generated sequence info for 1 studies with 1 entries total
INFO: Processing sequence infos to deduce study/session
INFO: Study session for {'locator': 'Halchenko/Yarik/950_bids_test4', 'session': None, 'subject': 'phantom1sid1'}
INFO: Need to process 1 study sessions
INFO: PROCESSING STARTS: {'session': None, 'outdir': '/private/var/folders/gs/4v1cn8mj4n52qrr9r6xv4ckm0000gn/T/pytest-of-contematto/pytest-13/test_scout_conversion0/Halchenko/Yarik/950_bids_test4', 'subject': 'phantom1sid1'}
INFO: Processing 1 pre-sorted seqinfo entries
INFO: Processing 1 seqinfo entries
INFO: Doing conversion using dcm2niix
[...]

PR #168, opened 2018-03-01
BF(workaround): provide explicit output_dir for dcm2niix call

yarikoptic:
Fixes gh-150
Ref: https://github.com/nipy/nipype/issues/2474


satra:
@yarikoptic - i'll let @mgxd review this and whether any nipype fix is needed.

yarikoptic:
I would love to finally kick out 0.5 so the faster the better .. just will do more testing locally - change is minimal. and seems need to meld with #167 anyways to make tests hopefully pass ...

yarikoptic:
superseded by #169

PR #172, opened 2018-03-01
Small fixes: skip tests when resource is N/A, do not install "tests" package

yarikoptic:


yarikoptic:
to improve coverage, at some point we should add travis matrix which lacks optional dependencies and runs tests offline.

yarikoptic:
ok, removed back the patch for python 2.7.old (on nd 14.04) which can't do `exec` in this form.  I will just patch for that distribution specifically.
I think all the changes for the best and noone objected, so let's make it happen

PR #177, opened 2018-03-16
sty: readme formatting, link to condensed slides

mgxd:
* currently hosting new tutorial slides on my fork, but would be great if we could host on `nipy.github.io/heudiconv`

PR #178, opened 2018-03-19
BF: pydicom 1.0.2 provided custom exception while trying to import dicom

yarikoptic:
Should fix up testing failures encountered during #177 

We would need to kick out yet another bugfix release for that... hopefully the last one in dealing with pydicom changes... heh


yarikoptic:
I will merge it to get us into usable state but feel free to review/suggest changes

PR #183, opened 2018-03-27
BF: do not remove all spaces after a number

### LINKED ISSUE #182: assertion failure in json_dumps_pretty
I have a SeriesDescription that is causing the following assertion to fail:
assert j_just_reloaded == j_tuned

js['SeriesDescription'] is fine (it contains underscores) but js['global']['const']['SeriesDescription'] has spaces:
"SeriesDescription": "Trace:Nov 13 2017 14-36-14 EST"

And is altered to:
 "SeriesDescription": "Trace:Nov 132017 14-36-14EST"

The regex doing this is:
js_ = re.sub('  *("?[-+.0-9e]+"?)[ \n]*', r' \1', js_)

yarikoptic:
My whole approach there for "pretty printing" json is a hack and should be RFed away eventually whenever more reports would come in ;)
For now this should close #182

### DIFF
--- a/heudiconv/utils.py
+++ b/heudiconv/utils.py
@@ -200,7 +200,8 @@ def json_dumps_pretty(j, indent=2, sort_keys=True):
     # uniform no spaces before ]
     js_ = re.sub(" *\]", "]", js_)
     # uniform spacing before numbers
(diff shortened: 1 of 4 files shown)


yarikoptic:
blocked by #184

yarikoptic:
rebased on top of the master with #184 fix, let's see

PR #185, opened 2018-03-28
fix: use source_dir instead

mgxd:
The call to `Dcm2niix` was passing in `source_names`, which was essentially being treated as `source_dir` anyways under the hood.

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -390,12 +390,13 @@ def nipype_convert(item_dicoms, prefix, with_prov, bids, tmpdir):
 
     item_dicoms = list(map(op.abspath, item_dicoms)) # absolute paths
 
+    dicom_dir = op.dirname(item_dicoms[0]) if item_dicoms else None
+
     convertnode = Node(Dcm2niix(), name='convert')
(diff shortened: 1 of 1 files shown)


heudiconv/convert.py
```diff
@@ -390,12 +390,13 @@ def nipype_convert(item_dicoms, prefix, with_prov, bids, tmpdir):
 
     item_dicoms = list(map(op.abspath, item_dicoms)) # absolute paths
 
+    dicom_dir = op.dirname(item_dicoms[0]) if item_dicoms else None
+
     convertnode = Node(Dcm2niix(), name='convert')
     convertnode.base_dir = tmpdir
-    convertnode.inputs.source_names = item_dicoms
+    convertnode.inputs.source_dir = dicom_dir
```
yarikoptic: would it work with older nipype?

mgxd: yes

mgxd:
@yarikoptic any objections?

yarikoptic:
I have tested (tests) on local machine with nipype 1.0.1 -- seems to pass. Merging, thanks @mgxd !

PR #187, opened 2018-04-09
WiP: ReproIn heuristic - support usecase from @mih et al.

yarikoptic:
- allow/strip any leading CAPITALS followed by a column with the series/protocol name (see https://github.com/ReproNim/reproin/issues/14)
- consider `series_description` after `protocol_name` if it was not only empty, but didn't parse according to our spec (thus should "allow" for `protocol_name` to be a fancy "xnatDownload" as in https://github.com/ReproNim/reproin/issues/14)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ before_install:
   - pip install --upgrade virtualenv
   - virtualenv --python=python venv
   - source venv/bin/activate
+  - pip --version # check again since seems that python_requires='>=3.5' in secretstorage is not in effect
   - python --version # just to check
   - pip install -r dev-requirements.txt
   - pip install requests # below installs pyld but that assumes we have requests already
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -293,10 +293,14 @@ def get_formatted_scans_key_row(item):
                                             force=True))
     # we need to store filenames and acquisition times
     # parse date and time and get it into isoformat
-    date = mw.dcm_data.ContentDate
-    time = mw.dcm_data.ContentTime.split('.')[0]
-    td = time + date
-    acq_time = datetime.strptime(td, '%H%M%S%Y%m%d').isoformat()
+    try:
+        date = mw.dcm_data.ContentDate
+        time = mw.dcm_data.ContentTime.split('.')[0]
+        td = time + date
+        acq_time = datetime.strptime(td, '%H%M%S%Y%m%d').isoformat()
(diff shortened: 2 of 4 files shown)


mih:
If it wouldn't be 'WiP' I would have merged it in order to get a little further with https://github.com/datalad/datalad-neuroimaging/pull/17

yarikoptic:
bloody python3 zealots ...
```
SecretStorage requires Python '>=3.5' but the running Python is 2.7.14
```
so we could expect lots of fun with deployments of datalad if pip would not become smart to not even consider "new" versions for incompatible python. According to https://github.com/PLOS/allofplos/issues/20 it should be the case for pip >= 9. I think we get 10, and I guess that doesn't feature doesn't work since secret storage does have 
```python_requires='>=3.5',``` , so either virtualenv causes older pip to be used somehow (yet to check) or that feature in pip is buggy

**Comment by mih** [comment]
> we could expect lots of fun with deployments of datalad

Mission accomplished: https://github.com/datalad/datalad/issues/2442

**Comment by yarikoptic** [comment]
ok, tests passed and I hope that this doesn't ruin anything, merging

**Comment by mih** [comment]
FTR: With this PR, but without ContentDate/Time:

```
WARNING: Failed to get date/time for the content: Dataset does not have attribute 'ContentDate'.
  File "/home/mih/env/datalad3-dev/bin/heudiconv", line 11, in <module>
    load_entry_point('heudiconv', 'console_scripts', 'heudiconv')()
  File "/home/mih/hacking/heudiconv/heudiconv/cli/run.py", line 120, in main
    process_args(args)
  File "/home/mih/hacking/heudiconv/heudiconv/cli/run.py", line 330, in process_args
    overwrite=args.overwrite,)
  File "/home/mih/hacking/heudiconv/heudiconv/convert.py", line 198, in prep_conversion
    overwrite=overwrite,)
  File "/home/mih/hacking/heudiconv/heudiconv/convert.py", line 292, in convert
    save_scans_key(item, bids_outfiles)
  File "/home/mih/hacking/heudiconv/heudiconv/bids.py", line 237, in save_scans_key
    op.join(output_dir, 'sub-{0}{1}_scans.tsv'.format(subj, ses)), rows)
  File "/home/mih/hacking/heudiconv/heudiconv/bids.py", line 270, in add_rows_to_scans_keys_file
    data_rows_sorted = sorted(data_rows, key=lambda x: (x[1], x[0]))
TypeError: '<' not supported between instances of 'NoneType' and 'str'
```

Issue #188, opened 2018-04-09
excessive conversions


mgxd:
In cases where all DICOMs are lumped together in the same directory, the entire dataset is being converted multiple times, since we are looping across series https://github.com/nipy/heudiconv/blob/4eff1a663fb562dabd00fc3b470204c7f2097aaf/heudiconv/convert.py#L235 and yet passing in all DICOMs each time https://github.com/nipy/heudiconv/blob/4eff1a663fb562dabd00fc3b470204c7f2097aaf/heudiconv/convert.py#L393-L397 Introduced in #185, patch in progress.

chpiero:
Is it possible to have an updated version of the docker container? What I see is 16 days ago while the fix is 4 days ago. Also on git hub there is an old version of the convert.py file

chpiero:
Sorry but the fix doesn't work

mgxd:
closed in #207

PR #189, opened 2018-04-09
[WIP] fix: ensure partitioned convertion

mgxd:
addresses #188

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -270,18 +270,31 @@ def convert(items, converter, scaninfo_suffix, custom_callable, with_prov,
                                      prefix + scaninfo_suffix)
 
                 if not op.exists(outname) or overwrite:
-
-                    # first copy files select files
+                    # softlink select files
                     item_dicoms_dir = tempdirs('DICOMs')
+                    lgr.info(
+                        "Symlinking series %s/%s into %s",
+                        item_idx+1,
+                        len(items),
+                        item_dicoms_dir)
+
                     for dcm in item_dicoms:
-                        shutil.copy(
-                            dcm, op.join(item_dicoms_dir, op.basename(dcm)))
+                        try:
+                            os.symlink(
+                                dcm, op.join(item_dicoms_dir, op.basename(dcm)))
+                        except:  # unsupported
+                            shutil.copy(
+                                dcm, op.join(item_dicoms_dir, op.basename(dcm)))
 
                     tmpdir = tempdirs('dcm2niix')
                     # run conversion through nipype
                     res, prov_file = nipype_convert(
                         item_dicoms_dir, prefix, with_prov, bids, tmpdir)
 
+                    try:
+                        os.unlink(item_dicoms_dir)
+                    except:  # DICOMs were copied instead
+                        pass
                     tempdirs.rmtree(item_dicoms_dir)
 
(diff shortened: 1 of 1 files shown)


heudiconv/convert.py
```diff
@@ -271,16 +270,23 @@ def convert(items, converter, scaninfo_suffix, custom_callable, with_prov,
                                      prefix + scaninfo_suffix)
 
                 if not op.exists(outname) or overwrite:
-                    tmpdir = tempdirs('dcm2niix')
 
+                    # first copy files select files
+                    item_dicoms_dir = tempdirs('DICOMs')
+                    for dcm in item_dicoms:
```
yarikoptic: Please add some log message (I would have added info level above the loop) about this copying happening since otherwise it would be just "busy doing something" for a while

heudiconv/convert.py
```diff
@@ -271,16 +270,23 @@ def convert(items, converter, scaninfo_suffix, custom_callable, with_prov,
                                      prefix + scaninfo_suffix)
 
                 if not op.exists(outname) or overwrite:
-                    tmpdir = tempdirs('dcm2niix')
 
+                    # first copy files select files
```
yarikoptic: typo -  "copy selected files"?

heudiconv/convert.py
```diff
@@ -271,16 +270,23 @@ def convert(items, converter, scaninfo_suffix, custom_callable, with_prov,
...
+                    item_dicoms_dir = tempdirs('DICOMs')
+                    for dcm in item_dicoms:
+                        shutil.copy(
+                            dcm, op.join(item_dicoms_dir, op.basename(dcm)))
```
yarikoptic: curious: symlinking wouldn't be enough?
I have tested on `///dicoms/dartmouth-phantoms/bids_test3-20161011//bids_anat-T1w` -- seems have worked just fine.  so why not to try to symlink first and then if that fails (file system doesn't support it) - do full copy (I wish there was something like `cp --reflink=auto` in shutils to facilitate fast copy on COW filesystems)

mgxd:
@yarikoptic thank you! Symlinking sounds much better than my solution :+1:

yarikoptic:
My answer: The spiral of the evolution + 0% dedicated tests to assure that behavior?

satra:
but we are still grouping dicoms right? so we can implement that behavior, right?

mgxd:
I'm not entirely sure how this was happening before all these changes, but before we merged #185, we weren't using `dcm2niix` "correctly" - `dcm2niix` doesn't take a list of files as the input (but this is something we could perhaps implement in the Nipype wrapper)

```
usage: dcm2niix [options] <in_folder>
```
We were passing in the first element of `item_dicoms` as the `in_folder`

This new way will link all our grouped DICOMs to a `tmpdir`, and then that is passed in for conversion.

satra:
@mgxd - this is a very simple fix:

we just need to change this line:
https://github.com/nipy/nipype/blob/master/nipype/interfaces/dcm2nii.py#L402

to: 

`return spec.argstr % os.path.dirname(val[0])`

and then you don't need to send a directory as the first value.

the `Node` will link all the items to the local directory and the dirname will be local.

and then in heudiconv give the list of dicoms to source_names.

satra:
@mgxd - why was this closed? #207 should fix this though.

PR #202, opened 2018-04-28
Adding video tutorial to the README page

chrisfilo:


yarikoptic:
Thank you @chrisfilo and sorry that we didn't react in time.  We already have that tutorial listed below in https://github.com/nipy/heudiconv/blob/5ee4652efabe7c06d45737c901b51d266473c16f/README.md#other-tutorials     Since README is already long enough, not sure if we should fancy it with sizeable video screenshots.  So I will just close without merging for now

Issue #162, opened 2018-02-22
Support for multi-echo fMRI


emdupre:
In the current BIDS spec, we have support for multi-echo fMRI data sets, like https://openfmri.org/dataset/ds000258/.  However, heudiconv seems to be outputting the different echoes as individual runs, rather than recognizing that all echoes belong to the same run. This will make interfacing with downstream multi-echo tools like https://github.com/poldracklab/fmriprep/pull/891 problematic.

Is there a preferred path for tackling this issue? I don't have a multi-echo phantom, but am happy to contribute code.

yarikoptic:
checked with our physicist, he believes we do not have EPI multi-echo sequences, but we have some other double-echo one.  He will collect a sample phantom dataset on monday, but still would be neat to get proper (but minimlistic -- the smallest matrix, # of volumes, etc) EPI multi-echo phantom data ;)

yarikoptic:
here is a sample of multi-echo (non-EPI) sequence which collects both PD and T2w in one shot
http://datasets.datalad.org/?dir=/dicoms/dartmouth-phantoms/bids_test6-PD+T2w/ . You can get it via `datalad install ///dicoms/dartmouth-phantoms/bids_test6-PD+T2w/`
(note that I am only about to fix us not url unquoting that + in the url, so you might get "%2B" in the resultant directory name)

emdupre:
Pinging @rmarkello here who has been working on a solution for this use case.

rmarkello:
Thanks for providing an open test set, @yarikoptic! A couple of questions about implementation:

1. Would it be reasonable to provide individuals with information in the `dicominfo.tsv` files noting that the sequence is multi-echo (e.g., making the TE entry a list if multiple echo times exist)?
2. Is it fair to simply inject the `echo-<index>` key-value pair into the `template` provided in the heuristic file if a given sequence is multi-echo (and BIDS output is desired), or should an additional template field be provided for use in the heuristic under the assumption that users will specify it if they want it differentiated?

leej3:
Just wondering has anyone done anything further on this? I was hoping to use this functionality. Happy to help try and implement it.
If i run heudiconv on multi-echo scans from siemens i get the behaviour as described above (" heudiconv seems to be outputting the different echoes as individual runs"). If those numbers were incorporated in the template that would be sufficient. I don't know if they are guaranteed to correspond though. I haven't spotted how they are generated. 
[...]

PR #203, opened 2018-04-30
WIP: add multi-echo support

leej3:
not sure how to get echo  for passing to conversion_info yet. Once I do I think this works across all vendors though.

There may be an issue on GE scanners in that they don't correctly write to the dicom tag specified for echo number. I'll try to get this working for now though.

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -34,7 +34,7 @@ from .dicoms import (
 lgr = logging.getLogger(__name__)
 
 
-def conversion_info(subject, outdir, info, filegroup, ses):
+def conversion_info(subject, outdir, info, filegroup, ses,echo=None):
     convert_info = []
     for key, items in info.items():
         if not items:
@@ -58,6 +58,7 @@ def conversion_info(subject, outdir, info, filegroup, ses):
                     subject=subject,
                     seqitem=item,
                     subindex=subindex + 1,
+                    echo = echo,
                     session='ses-' + str(ses),
                     bids_subject_session_prefix=
                         'sub-%s' % subject + (('_ses-%s' % ses) if ses else ''),
@@ -176,7 +177,7 @@ def prep_conversion(sid, dicoms, outdir, heuristic, converter, anon_sid,
         assure_no_file_exists(edit_file)
         write_config(edit_file, info)
         save_json(filegroup_file, filegroup)
-
+    import pdb;pdb.set_trace()
     if bids:
         # the other portion of the path would mimic BIDS layout
         # so we don't need to worry here about sub, ses at all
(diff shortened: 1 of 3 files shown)


leej3:
fixes #162

heudiconv/convert.py
```diff
@@ -176,7 +177,7 @@ def prep_conversion(sid, dicoms, outdir, heuristic, converter, anon_sid,
         assure_no_file_exists(edit_file)
         write_config(edit_file, info)
         save_json(filegroup_file, filegroup)
-
+    import pdb;pdb.set_trace()
```
yarikoptic: we don't need this one do we? ;)

heudiconv/dicoms.py
```diff
@@ -194,6 +194,10 @@ def group_dicoms_into_seqinfos(files, file_filter, dcmfilter, grouping):
         else:
             sequence_name = 'Not found'
 
+        if dcminfo.get([0x18,0x86], None):
+            # Siemens scanners
+            echo = dcminfo[0x18,0x86].value
```
yarikoptic: yet to decide (ref: https://github.com/nipy/heudiconv/issues/162#issuecomment-385428448) either we should handle it manually at all or rely on annotation from dcm2niix.  That one would exist though only whenever `--bids` side car file is generated...  my points somewhat -- less we (re)implement of logic dcm2niix provides (and tests for) inside, the better

heudiconv/convert.py
```diff
@@ -34,7 +34,7 @@
 lgr = logging.getLogger(__name__)
 
 
-def conversion_info(subject, outdir, info, filegroup, ses):
+def conversion_info(subject, outdir, info, filegroup, ses,echo=None):
```
yarikoptic: not that we are all "in clear" but it is better to stay `pep8` friendly in the new code -- please add a space before `echo=` here and also where you call it below, but remove around ` = ` when passing as a keyword argument.

I would recommend configuring your IDE/editor to autohighlight pep8 gotchas -- pycharm does it quite beautifully

heudiconv/convert.py
leej3: ha. nope. definitely a work in progress :)

heudiconv/dicoms.py
leej3: I agree. if its trivial to implement we can do this approach. otherwise the multi-echo support will be implemented at a completely different stage of the heudiconv run.

heudiconv/convert.py
leej3: duly noted

leej3:
Update: in order to pass echo to conversion_info I would need to pass seqinfo_list. this is only available if reuse_conversion_table is False. 

With that said, the dicom tag is completely useless for the current session I am looking at. It doesn't remotely map to the true echo number so I vote to abandon this approach in favour of renaming the heudiconv output in accordance with the dcm2niix sidecar json. I'll have a quick look to see such an implementation is within my ken.

leej3:
Nope I'm struggling to get this working. Now when I run heudiconv I am not converting the multiple echos individually. I'm going to leave this be for the moment. Overall conclusion is that renaming the files based on echo number in the json side car created  by the the dcm2niix conversion would be the best way forward with this. I will upload a ME scan on GE to test that dcm2niix is not failing due to the assumption that the dicom tag 0x18,0x86 might actually contain useful information.

leej3:
sorry closed that by accident. but perhaps it should be closed...

thomshaw92:
@leej3 No! please don't close! 
We'd still love support for multi echo.
Cheers 👍

yarikoptic:
Other PRs referenced above provided alternative implementation

PR #204, opened 2018-05-01
Provide SeriesInstanceUID in SeqInfo

mih:
If one wants to implement a "deterministic" (as opposed to a heuristic), it is necessary that heudiconv provides a unique identifiers for an image series, so that conversion/naming parameters can be preconfigured specifically for a given series. Ideally that ID is unique not just within a session (e.g. series number), but globally unique (such as the `SeriesInstanceUID` used here).

Why do this? Because tools like datalad can perform a substantial part of the sorting and metadata inspection implemented in heudiconv, so their results would already be known before heudiconv sees DICOMs for the first time, and it could be instructed specifically what to do.

This PR enables the development of such deterministic heuristics.

Ping @bpoldrack -- PR to eliminate the upstream diff.

### DIFF
--- a/heudiconv/utils.py
+++ b/heudiconv/utils.py
@@ -38,7 +38,8 @@ seqinfo_fields = [
     'accession_number',      # 21
     'patient_age',           # 22
     'patient_sex',           # 23
-    'date'                   # 24
+    'date',                  # 24
+    'uid'                    # 25
  ]
 
 SeqInfo = namedtuple('SeqInfo', seqinfo_fields)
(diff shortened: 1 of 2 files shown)


heudiconv/utils.py
```diff
@@ -38,7 +38,8 @@
     'accession_number',      # 21
     'patient_age',           # 22
     'patient_sex',           # 23
-    'date'                   # 24
+    'date',                  # 24
+    'uid'                    # 25
```
yarikoptic: I know that is a "duplicate" and so far we were not really consistent, but may be we should keep `series_` prefix as in `series_id`, `series_description`, to make it clear that it is a property of the series (khe khe -- seqinfo) and not of the entire exam/study (as in `study_description`) .  Would pair nice(r) with aforemented `series_id`.  I would even add `study_uid` (although I think it might be a part of `series_id` ATM...?)

bpoldrack: Agree. Will do.

PR #207, opened 2018-05-05
fix multiple conversion #196

### LINKED ISSUE #196: heudiconv (--f reproin ?) gets confused by its own output
I run this, essentially sequentially converting two scans of the same subject that are supposed to go into the same dataset:

```
heudiconv -f reproin --files sourcedata/sub-02/ses-functional/dicoms -b -o . -l '' -s 02`
heudiconv -f reproin --files sourcedata/sub-02/ses-functional/dicoms -b -o . -l '' -s 02
```

The second call fails with

```
INFO: Running heudiconv version 0.5.dev1
INFO: Analyzing 5460 dicoms
<snip>
INFO: Converting sub-02/anat/sub-02_T1w (384 DICOMs) -> sub-02/anat . Converter: dcm2niix . Output types: ('nii.gz', 'dicom')
  File "/home/mih/env/datalad3-dev/bin/heudiconv", line 11, in <module>
    load_entry_point('heudiconv', 'console_scripts', 'heudiconv')()
  File "/home/mih/hacking/heudiconv/heudiconv/cli/run.py", line 120, in main
    process_args(args)
  File "/home/mih/hacking/heudiconv/heudiconv/cli/run.py", line 330, in process_args
    overwrite=args.overwrite,)
  File "/home/mih/hacking/heudiconv/heudiconv/convert.py", line 198, in prep_conversion
    overwrite=overwrite,)
  File "/home/mih/hacking/heudiconv/heudiconv/convert.py", line 303, in convert
    % (outname)
RuntimeError: was asked to convert into sub-02/anat/sub-02_T1w.nii.gz but destination already exists
```

Notice how it starts scanning 5k DICOMS (correct), but then only converts 384 (wrong). The reason is that with `-o .` it picks up the existing `.heudiconv/` from the first conversion and assumes that this is a 2nd-pass on the same conversion, and forgets that it started off on something entirely different.

While one could argue that I shouldn't be calling it like this, it should also not silently generate output that is unrelated to the given input args.

cni-md:
should be fix #196. With the original code all dicoms are fed to dcm2niix and not only one scan.

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -398,7 +398,7 @@ def nipype_convert(item_dicoms, prefix, with_prov, bids, tmpdir):
 
     convertnode = Node(Dcm2niix(), name='convert')
     convertnode.base_dir = tmpdir
-    convertnode.inputs.source_dir = dicom_dir
+    convertnode.inputs.source_names = item_dicoms
     convertnode.inputs.out_filename = op.basename(op.dirname(prefix))
 
     if nipype.__version__.split('.')[0] == '0':


satra:
@cni-md an @mgxd  - since @yarikoptic is asking for tests, can we add two tests - one for #196 and one for #188

yarikoptic:
Iirc it was some interface compatibility issue which came up with new nipype release

mgxd:
yes `source_names` isn't technically a `dcm2niix` input, but we decided to leave it to ease the handling of symlinking/copying files to the `Node` working directory in Nipype. 

let's merge for this now and i'll add a new test in a separate PR

PR #209, opened 2018-05-11
fix: various small fixes

mgxd:
Started out as `Dockerfile` fixes, evolved to showing heudiconv a little love

* update Dockerfile
  * add `util/gen-docker-image.sh` to easily generate new images with neurodocker
* `--overwrite` should not reuse conversion table
* if no valid argument is passed - don't print cryptic message, show help instead
* try to import latest pydicom first (> 1.0)

Classes #210

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 # pull request on our GitHub repository:
 #     https://github.com/kaczmarj/neurodocker
 #
-# Timestamp: 2018-02-02 16:32:55
+# Timestamp: 2018-05-11 21:21:11
 
 FROM debian:stretch
 
@@ -32,13 +32,6 @@ RUN apt-get update -qq && apt-get install -yq --no-install-recommends  \
     && chmod -R 777 /neurodocker && chmod a+s /neurodocker
 ENTRYPOINT ["/neurodocker/startup.sh"]
 
-RUN apt-get update -qq \
-    && apt-get install -y -q --no-install-recommends git \
-                                                     gcc \
-                                                     pigz \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-
 #------------------------
 # Install dcm2niix v1.0.20171215
 #------------------------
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -117,6 +118,20 @@ RUN echo '{ \
     \n      "debian:stretch" \
     \n    ], \
     \n    [ \
+    \n      "dcm2niix", \
+    \n      { \
+    \n        "version": "v1.0.20171215" \
```
yarikoptic: Hm... This one is old , better use more recent one, there were various fixes which are relevant

heudiconv/cli/run.py
```diff
@@ -237,6 +242,9 @@ def process_args(args):
     #
     # Load heuristic -- better do it asap to make sure it loads correctly
     #
+    if not args.heuristic:
+        raise RuntimeError("No heuristic specified - add to arguments and rerun")
+
```
yarikoptic: Love us great! But if only we could put it in writing (tests) to make sure it stays forever!

heudiconv/convert.py
```diff
@@ -115,7 +115,8 @@ def prep_conversion(sid, dicoms, outdir, heuristic, converter, anon_sid,
 
     # if conversion table(s) do not exist -- we need to prepare them
     # (the *prepare* stage in https://github.com/nipy/heudiconv/issues/134)
-    reuse_conversion_table = op.exists(edit_file)
+    # if overwrite - recalculate this anyways
+    reuse_conversion_table = op.exists(edit_file) and not overwrite
```
yarikoptic: I just feel we need a dedicated option to either reuse conversation table, otherwise we again keep pulling rope two ways back and forth... Eg what if someone might want to rerun conversion with new dcm2niix using his tuned conversion table?

mgxd: Perhaps a new option would be the best solution.. I ran into this problem when debugging a new heuristic, looks like we came across this bridge previously further down in the code :)

https://github.com/nipy/heudiconv/blob/9c3e82379ab70931b10e3b93bf96c3e30cb1d75a/heudiconv/convert.py#L128-L135

Dockerfile
mgxd: it is labelled as a pre-release so I was hesitant, but if you are vouching i'm for adding the latest and greatest

heudiconv/convert.py
mgxd: @yarikoptic how do you feel about forcing regeneration of the conversion table if the heuristic has changed?

yarikoptic: Ambivalent ;)  but isn't what we are doing already with above check based on file_md5sum?

heudiconv/convert.py
```diff
@@ -125,17 +126,15 @@ def prep_conversion(sid, dicoms, outdir, heuristic, converter, anon_sid,
     #  1. add a test
     #  2. possibly extract into a dedicated function for easier logic flow here
     #     and a dedicated unittest
-    if not reuse_conversion_table and \
-        op.exists(target_heuristic_filename) and \
-        file_md5sum(target_heuristic_filename) != file_md5sum(heuristic.filename):
+    if (op.exists(target_heuristic_filename) and
+        file_md5sum(target_heuristic_filename) != file_md5sum(heuristic.filename)):
+        # remake conversion table
```
mgxd: @yarikoptic at present if the edit file exists we would skip this check.

mgxd: @yarikoptic ping - the more I think about this, the more I like the idea of regenerating the conversion table if there has been a change to the heuristic used.

yarikoptic: I think we did have that intention originally but literally introduced a bug with the initial check (should have been `if  reuse_conversion_table`)... so let it be indeed

PR #218, opened 2018-06-13
BF+ENH: set dist-restrictions metadata correctly

### LINKED ISSUE #205: metadata is assigned incorrectly
I would blame DataLad API changing at some point and me for not having (apparently) a unittest within heudiconv to verify correct assignment of the metadata (to annotate sensitive files.  ATM they are assigned as two tags instead of `key=value`:
```shell
$> git annex metadata sub-sid000007_ses-1_T2w.nii.gz
metadata sub-sid000007_ses-1_T2w.nii.gz 
  lastchanged=2017-11-17@23-53-26
  tag=('distribution-restrictions', 'sensitive')
  tag-lastchanged=2017-11-17@23-53-26
ok
```

yarikoptic:
- Fixes #205 
- Should also now work with datalad 0.10 (set_metadata is a generator there now)
- Ideally we should have got a helper to "Fixup" metadata setting in already set incorrectly datasets

### DIFF
--- a/heudiconv/external/dlad.py
+++ b/heudiconv/external/dlad.py
@@ -1,4 +1,6 @@
+import inspect
 import os
+
 import os.path as op
 import logging
 from glob import glob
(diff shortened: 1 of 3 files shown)


yarikoptic:
eh, a slight coverage drop due to addition of a helper which I apparently ended up not using... Since we are  really in need of a release here, I will merge this one, and we could deal with fallout later if any

PR #223, opened 2018-06-26
BF: workaround for elderly buggy python 2.7 exec

### LINKED ISSUE #222: SyntaxError: unqualified exec [Centos 7 - Python 2.7.5]
Trying to install for the first time, getting the error below. A couple of other issues here mention this error, but the solution was for Ubuntu only. I also checked online and looks like this bug is fixed in python 2.7.9. The issue is that upgrading python can be problematic in linux, and I am not sure how heudiconv would work if installed on a specific python version whyle the standard `python` call will use the old version. Any idea how to make this work?

```
[root@mri dorian]# umask 0022
[root@mri dorian]# PREFIX_PATH=/APPS/
[root@mri dorian]# pip install --install-option="--prefix=$PREFIX_PATH" heudiconv[all]
/usr/lib/python2.7/site-packages/pip/_internal/commands/install.py:199: UserWarning: Disabling all use of wheels due to the use of --build-options / --global-options / --install-options.
  cmdoptions.check_install_build_global(options)
Collecting heudiconv[all]
  Using cached https://files.pythonhosted.org/packages/93/1b/3f866ddbf6af150ec94117ae5ddebf1c5584468d991345692fddf6224809/heudiconv-0.5.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-70IlXr/heudiconv/setup.py", line 22
        exec(infofile.read(), globals(), ldict)
    SyntaxError: unqualified exec is not allowed in function 'main' it contains a nested function with free variables

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-70IlXr/heudiconv/
```

yarikoptic:
Closes #222 
although it is up for a discussion if we would want to go that far ;-)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,9 @@ def main():
     # Get version and release info, which is all stored in heudiconv/info.py
     info_file = op.join(thispath, 'heudiconv', 'info.py')
     with open(info_file) as infofile:
-        exec(infofile.read(), globals(), ldict)
+        # exec(infofile.read(), globals(), ldict)
+        # Workaround for python 2.7 prior 2.7.8
+        eval(compile(infofile.read(), '<string>', 'exec'), globals(), ldict)
 
 
     def findsome(subdir, extensions):


yarikoptic:
Well, since it is compatible with all our pythons may be we indeed should use it for a few years ;-)

PR #225, opened 2018-06-27
revert back to the master of the stock dcmstack - should be good now

yarikoptic:
if all looks kosher -- let's pester Brendan to release ;-)


yarikoptic:
oh -- still needs python3 patches... uff

yarikoptic:
no longer needed - installing pip version now

PR #227, opened 2018-07-04
Fix links to heuristcs in README.md

mslw:
Since heuristics were moved one level lower in commit b1b7b60d, two links in README had to be updated.

This is an absolutly minor change. Feel free to discard it if a rewrite of readme before next release is coming anyway.


mgxd:
@mslw thanks for the patch!

PR #229, opened 2018-07-10
Allow absolute path for dcm2niix to use the compiled version of it without installing

ofgulban:
Hi,

This is a really great tool, so thanks to all the contributors.

I have came across a scenario where the user could not install dcm2niix (due to system permission rules of the institute). Upon consideration of many options, it was clear that the only practical way was to use heudiconv with compiled version of dcm2niix. I have made minimal changes to the code to accomplish this goal while not changing the default behavior. For example now the users can use:
```heudiconv -d /path/to/{subject}/*dcm -o /path/to/output -f /path/to/convertall.py -c /path/to/dcm2niix -s s1``` (see that depending on the absolute path any compiled version of dcm2niix can be used, which is now consistent with the rest of the syntax).

I think this would also free heudiconv from the necessitiy of manually installing dcm2niix as suggested in the readme. If this PR is accepted I think the readme and the related suggestions of manually installing dcm2niix can be updated.

Btw, I took the liberty of fixing PEP8 and PEP257 issues where possible in the files that I have edited. I have also made the help instructions in command line interface to have more consistent capitalization and punctuation, I think it looks a bit nicer. I have kept these as separate commits just in case anyway.

### DIFF
--- a/heudiconv/cli/run.py
+++ b/heudiconv/cli/run.py
@@ -18,14 +18,14 @@ INIT_MSG = "Running {packname} version {version}".format
 
 
 def is_interactive():
-   """Return True if all in/outs are tty"""
-   # TODO: check on windows if hasattr check would work correctly and add value:
-   return sys.stdin.isatty() and sys.stdout.isatty() and sys.stderr.isatty()
+    """Return True if all in/outs are tty."""
+    # TODO: check on windows if hasattr check would work correctly and add
+    # value:
+    return sys.stdin.isatty() and sys.stdout.isatty() and sys.stderr.isatty()
 
 
 def setup_exceptionhook():
-    """
-    Overloads default sys.excepthook with our exceptionhook handler.
+    """Overloads default sys.excepthook with our exceptionhook handler.
 
     If interactive, our exceptionhook handler will invoke pdb.post_mortem;
     if not interactive, then invokes default handler.
@@ -45,8 +45,9 @@ def setup_exceptionhook():
 
 
 def process_extra_commands(outdir, args):
-    """
-    Perform custom command instead of regular operations. Supported commands:
+    """Perform custom command instead of regular operations.
+
+    Supported commands:
     ['treat-json', 'ls', 'populate-templates']
 
     Parameters
@@ -55,6 +56,7 @@ def process_extra_commands(outdir, args):
         Output directory
     args : Namespace
         arguments
+
     """
     if args.command == 'treat-jsons':
         for f in args.files:
@@ -78,7 +80,8 @@ def process_extra_commands(outdir, args):
     elif args.command == 'populate-templates':
         heuristic = load_heuristic(args.heuristic)
         for f in args.files:
-            populate_bids_templates(f, getattr(heuristic, 'DEFAULT_FIELDS', {}))
+            populate_bids_templates(
+                f, getattr(heuristic, 'DEFAULT_FIELDS', {}))
     elif args.command == 'sanitize-jsons':
         tuneup_bids_json_files(args.files)
     elif args.command == 'heuristics':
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mgxd** [comment]
@ofgulban if you add the `dcm2niix` executable to your PATH before running heudiconv, you should be able to run heudiconv normally

**Comment by ofgulban** [comment]
@mgxd thanks a lot for the suggestion and the quick reply. I have tried various .bashrc tricks similar to your suggestion but nothing worked. Upon following heudiconv code through nipype core it seemed to me that the default call to dcm2niix is set to the system-wise installation path. However, to my knowledge, this path can not be edited without having appropriate user privilages (and the system where this problem occured is a very restricted one)

This was the reason why I have solved the issue by making use of an already available nipype functionality (see 9ba7e23 ). I have also thought that this small convenience improvement would be more straightforward for the general users.

**Comment by mgxd** [comment]
@ofgulban 
[this snippet of code](https://github.com/nipy/nipype/blob/master/nipype/utils/filemanip.py#L835-L874) is used within nipype to locate executables based on commands. If you write the following command in a terminal prior to running heudiconv, do you still run into problems?

`export PATH=$PATH:</path/to/folder/with/dcm2niix/executable>`

If so, could you ensure the executable is given the proper permissions (chmod +x dcm2niix)? Let me know if this doesn't work

**Comment by yarikoptic** [comment]
or if you need to overload default dcm2niix (too old?) then prepend that path: 
```
export PATH=</path/to/folder/with/dcm2niix/executable>:$PATH
```

**Comment by yarikoptic** [comment]
In general, I think it is not really needed (due to above) but if it would be of help to someone I do not mind **but** I would prefer to avoid changing private (`_cmd`) attributes in external libraries -- it begs for future problems with compatibility/maintenance.  So let's first figure out if it is really needed ;-)

**Comment by ofgulban** [comment]
@mgxd & @yarikoptic thanks for your suggestions. I have tried:
```
export PATH="$PATH:</home/faruk/mricrogl_lx/dcm2niix>"

echo $PATH
/home/faruk/miniconda2/bin:/home/faruk/miniconda2/bin:/usr/lib/fsl/5.0:/usr/local/bin:/usr/bin:/bin:</home/faruk/mricrogl_lx/dcm2niix>:</home/faruk/mricrogl_lx/dcm2niix>
```
The suggested command worked but heudiconv did not work. I have also used ```chmod +x dcm2niix``` which did not solve the issue. I can execute dcm2niix by itself that is not the problem. The problem is heudiconv can not make use of the user environment `PATH`.

I might lack the correct terminology to describe this but to me it seems that nipype default dcm2niix is set to run system-wise (allusers) installation of dcm2niix as a process that is separate from the user's environment. Therefore any user environment variable I have changed does not effect heudiconv behaviour. This was my main reason to introduce `_cmd` assignment.

Just to note again, this specific computer has restricted user permissions, I cannot edit paths for all-users (e.g. from /etc/profile) or build dcm2niix. @yarikoptic there is no old dcm2niix because it was never installed in this computer. This is why I was using already compiled version of dcm2niix.

If you have more suggestions I am happy to continue trying.

**Comment by yarikoptic** [comment]
oh -- remove those `<` and `>`, they were used just to signal where to place things ;-)
When you type `which dcm2niix` after that export, it should point to the dcm2niix of yours

**Comment by ofgulban** [comment]
Sorry, my bad. 
```
export PATH="$PATH:/home/faruk/mricrogl_lx/dcm2niix"

echo $PATH
/home/faruk/miniconda2/bin:/home/faruk/miniconda2/bin:/usr/lib/fsl/5.0:/usr/local/bin:/usr/bin:/bin:/home/faruk/mricrogl_lx/dcm2niix
```

returns the same error as before:
```
IOError: No command "dcm2niix" found on host w110er. Please check that the corresponding package is installed.
```

I have also tried to use alias:
```
alias dcm2niix="/home/faruk/mricrogl_lx/dcm2niix"
```

Which makes it possible to call dcm2niix without the absolute path, but this again does not make heudiconv run dcm2niix.

**Comment by ofgulban** [comment]
Aaah, ok :D

I was including dcm2niix in the path, when I have changed it to:

```
export PATH="$PATH:/home/faruk/mricrogl_lx"
```

it worked (with heudiconv too)

ofgulban:
Cool, thanks @mgxd and @yarikoptic . If you would like to have the `_cmd` parts of this PR I can create a separate PR for that. Might be useful to add a line about this to readme too. I have got fixated on the dcm2niix has to be installed part I guess.

mgxd:
hey @ofgulban , if you'd like to open the PEP updates in another PR, that would be great! I'm gonna close this one, since we can achieve the same goal without modifying `Dcm2niix`'s `_cmd` attribute

PR #238, opened 2018-08-02
Just a minor for reproin

yarikoptic:
just want to make sure no magical breakage of tests with this irrelevant changes ... and will merge


yarikoptic:
tested in #239 that travis misbehavior is not the fault of this PR ;)

PR #239, opened 2018-08-02
empty - just to test if it is a genuine new pytest regression leading to failed travis in other PRs

yarikoptic:


yarikoptic:
so something is up with pytest indeed

PR #245, opened 2018-08-16
added recurse_dicom_dir to heudiconv

tjhendrickson:
Added recurse_dicom_dir argument (--r, --recurse_dicom_dir) to run.py and logic to parser.py, in order so sub-directories are automatically recursed within the dicom_dir_template

### DIFF
--- a/heudiconv/cli/run.py
+++ b/heudiconv/cli/run.py
@@ -137,6 +137,8 @@ def get_parser():
                        '(can be compressed) are supported in addition to '
                        'directory. All matching tarballs for a subject are '
                        'extracted and their content processed in a single pass')
+    group.add_argument('-r', '--recurse_dicom_dir', dest='recurse_dicom_dir',
+                       help='automatically recurse dicomdir sub directories to find all dicom files')
     group.add_argument('--files', nargs='*',
                        help='Files (tarballs, dicoms) or directories '
                        'containing files to process. Cannot be provided if '
--- a/heudiconv/cli/run.py
+++ b/heudiconv/cli/run.py
@@ -63,9 +63,8 @@ def process_extra_commands(outdir, args):
         heuristic = load_heuristic(args.heuristic)
         heuristic_ls = getattr(heuristic, 'ls', None)
         for f in args.files:
-            study_sessions = get_study_sessions(
-                args.dicom_dir_template, [f], heuristic, outdir,
-                args.session, args.subjs, grouping=args.grouping)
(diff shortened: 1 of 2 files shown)


heudiconv/cli/run.py
```diff
@@ -137,6 +137,8 @@ def get_parser():
                        '(can be compressed) are supported in addition to '
                        'directory. All matching tarballs for a subject are '
                        'extracted and their content processed in a single pass')
+    group.add_argument('-r', '--recurse_dicom_dir', dest='recurse_dicom_dir',
+                       help='automatically recurse dicomdir sub directories to find all dicom files')
```
yarikoptic: note that it is relevant only in the case of using a template... I think the better and somewhat easier solution is just to support '**' globbing and demand python 3.5 for that.  `recursive` kwarg to glob could be set if '**' is detected in the glob pattern

tjhendrickson: Hmm. I was under the impression that heudiconv was under python 2.7 support, thus this is the reason I chose to do it the way I did. Wouldn't switching to python 3.5 but more burdensome  overall than using perhaps a sub-optimal way to recurse through the dicom directories? 

Am I missing something?

mgxd: another option would be integrating [glob2](https://github.com/miracle2k/python-glob2) to support `**` globbing in py2/3

tjhendrickson: I am in a bit of a time crunch for this. Could we merge the '--recurse_dicom_dir'?

mgxd: @tjhendrickson tests are not passing yet, but perhaps we don't need to add the additional argument - we can just make the recursive behavior "smart" in that if the `dicom_dir_template` contains `**`, we recurse. If not, we go about with a normal `glob` call

tjhendrickson: Oh I see. I can look into this tomorrow.

tjhendrickson:
Fingers crossed that the checks pass!

tjhendrickson:
My changes did not push for some reason...weird. Before I do I realize that I need to include glob2, is there a way or a file I should modify in order to install glob2?

mgxd:
@tjhendrickson the `os.walk` implementation you previously integrated seems like a better solution (just to avoid adding additional dependencies) - I'd recommend writing a helper function that either uses `os.walk` or `glob` to fetch the matching files in `dicom_dir_template`

mgxd:
@tjhendrickson are you still interested in this PR? You should be able to achieve this behavior with the `--files <directory>` flag, which will search all sub-directories of `<directory>`

mgxd:
@tjhendrickson going to close this since it has gone stale - feel free to open if you wish to keep working on it/

Issue #249, opened 2018-08-24
BIDS Validator ERROR - [Code 83] ECHOTIME1_2_DIFFERENCE_UNREASONABLE


bmelo:
When I use the BIDS validator (http://incf.github.io/bids-validator/) on a dataset converted with heudiconv v0.5.1 (dcm2niix v1.0.20180622), I'm getting this error message:

> **[Code 83] ECHOTIME1_2_DIFFERENCE_UNREASONABLE**
> The value of (EchoTime2 - EchoTime1) should be within the range of 0.0001 - 0.01.

This dataset has a field mapping sequence with two magnitude images (Siemens MR), converted using this key:
```python
def create_key(template, outtype=('nii.gz'), annotation_classes=None):
    if template is None or not template:
        raise ValueError('Template must be a valid format string')
    return (template, outtype, annotation_classes)

fmap_mag    = create_key('sub-{subject}/fmap/sub-{subject}_magnitude')
```
For some subjects, EchoTime of magnitude1.nii.gz is higher than EchoTime of magnitude2.nii.gz. But BIDS specification (v1.1.1, section 8.3.5.1) says: 

> ... EchoTime1corresponds to the shorter echo time and EchoTime2 to the longer echo time. Similarly _magnitude1 image corresponds to the shorter echo time and the OPTIONAL _magnitude2 image to the longer echo time.

**Obs.:** Another point related to this conversion, the BIDS specification template (same section) has not json files for magnitude images, but heudiconv is producing the ".._magnitude1.json" and ".._magnitude2.json" files.

Is it a problem with heudiconv? I can help you figure it out.

yarikoptic:
https://github.com/nipy/heudiconv/blob/master/heudiconv/bids.py#L144 points to the fact that someone the shorter one is the 2nd file although seems to be not the case for me... also from https://github.com/nipy/heudiconv/blob/master/heudiconv/convert.py#L445  where we take output from nipype node for dcm2niix, I wonder now if the order of entries there is guaranteed (sorted) anyhow... if you could check (are comfortable with python's pdb? if not - that is ok, I could guide or add more of .debug msgs to the code) at that point in two "differing" cases -- is it the matter of sorting files within Python side (that fmap_e2 comes before fmap_e1) or it is the output of dcm2niix that EchoTime is swapped and e2 has a shorter one?

bmelo:
Hi @yarikoptic, thank you! 

[...]

PR #250, opened 2018-08-24
Trying to solve issue #249.

bmelo:
Only tested with a small sample.

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -442,7 +442,7 @@ def save_converted_files(res, item_dicoms, bids, outtype, prefix, outname_bids, 
     from nipype.interfaces.base import isdefined
 
     bids_outfiles = []
-    res_files = sorted(res.outputs.converted_files)
+    res_files = res.outputs.converted_files
 
     if not len(res_files):
         lgr.debug("DICOMs {} were not converted".format(item_dicoms))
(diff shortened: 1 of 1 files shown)


yarikoptic:
so it did resolve the #249 issue for you on your data?

yarikoptic:
FWIW since tests did pass in #251 it means that sorting the filename at those points did cause some kind of change in behavior in our pipeline.

bmelo:
Sorry, I did a very minimalist test and it passed. Now I saw the mistake and I'll fix this in a few minutes.

yarikoptic:
according to the coverage report it seems to do change logic : https://codecov.io/gh/nipy/heudiconv/pull/250/src/heudiconv/convert.py  not sure yet if for the better.  So, to start with would be nice if you could test on your "full" case.  I will test on mine later on to see if that is somehow changes things unexpectedly

bmelo:
Hi @yarikoptic , I tested with 4 different datasets (3 Siemens, 1 Philips) and using T1w, bold, and fmap sequences. Heudiconv converted everything without errors.
In one of these datasets, I need to export all fmap sequences (82 sessions) again, and the validator does not report any error. 

For now, I'm going to have to use this version because error #249 is recurring in our environment. Thank you!

yarikoptic:
I thought we had merged this one awhile back... I was wrong.
Since we only heard confirmations that it resolved the issue, I will just merge this for now as is.

PR #251, opened 2018-08-24
Minor fixups to the reproin heuristic local cases

yarikoptic:
but largely to check if the failures in #250 are indeed triggered by introduced sorting or we have a more generic problem


mgxd:
@yarikoptic has this gone stale?

yarikoptic:
I guess... Will resurrect if needed

Issue #80, opened 2017-09-13
Siemens Csa headers


kasbohm:
Hi, running heudiconv on Siemens (Avanto/Skyra/Prisma) data, I get:
`Ignoring xxxx.IMA since not quite a «normal» DICOM: Dataset does not have attribute ‘ProtocolName’.`

I guess this is caused by the attribute not being present at the normal location (0018 1030) in our Siemens DICOMS, but exists as tProtocolName in the private Csa header. Did I miss a switch when calling heudiconv, or does some logic need to be added? nibabel.nicom.csareader? 

Cheers!

neurolabusc:
I updated dcm2niix so if ProtocolName is not in the DICOM header it use the ProtocolName in the CSA header for the BIDS sidecar. This is a hail mary, and only influences the BIDS sidecar: the protocol name will still be empty with regards to the output filename (e.g. `-f %p`). Whether this helps you depends on whether heudiconv uses dcmstack or dcm2niix to determine the protocol name.

satra:
@neurolabusc - thanks. at this point we would probably have to make this change on our side inside our grouping function. but eventually we should consider where the metadata get extracted.

satra:
@mgxd - can you add this as part of your refactor?

PR #254, opened 2018-08-31
Workaround for protocol name in private siemens csa headers #80

kasbohm:
#80 
Tested on Siemens Avanto, Prisma, Skyra. I am unsure performance takes a hit, but the changes should not affect files where protocolName is not missing.

1. If ProtocolName is missing, set to '', and then see if we can find it in csa headers.
2. Prevent dcminfo.get('SeriesDescription') to cause abort if missing


mgxd:
hi @kasbohm - sorry for the late response - searching the CSA header is a good fallback for many fields, not just `ProtocolName`, so I've started a branch from your fork adding this behavior as a function (#271). I've preserved your commits in that branch, so I'm going to close this for now, thanks for your contribution!

PR #263, opened 2018-10-15
MAINT: add dcmstack to deps

mgxd:
now that its been added to pypi

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,1 @@
 .[all]
-# This is a dcmstack branch with changes for Python 3
-# sent PR to main repo, TODO: check if merged
-# https://github.com/ghisvail/dcmstack/pull/1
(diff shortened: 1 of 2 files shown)


yarikoptic:
yeap, dcmstack 0.7 was released with python3 support! woohoo!

requirements.txt
```diff
@@ -1,5 +1 @@
 .[all]
-# This is a dcmstack branch with changes for Python 3
-# sent PR to main repo, TODO: check if merged
-# https://github.com/ghisvail/dcmstack/pull/1
-git+git://github.com/mvdoc/dcmstack@bf/importsys
```
yarikoptic: then we probably above should demand `dcmstack >= 0.7` (at least for python3)

yarikoptic: pushed that change in 2c239d9ddb009f4013badecf2c14d5e47154d5c4

PR #264, opened 2018-10-23
Few small fixes from using ReproIn @ DBIC

yarikoptic:
- allow for a space to be used as a separator within `locator`
- provide details while throwing error on detecting multiple session markers

above will be merged locally/pushed, here is the next change:

- Reproin: make __dups indexes be per series (by default, not tunable ATM)

TODO:
- [x] test on some real case with lots of _dups


yarikoptic:
Changes are trivial so assume no objections (need to push a separate PR on top)

yarikoptic:
Crap - pushed before hitting merge!  I will merge/push manually so you could observe a clean diff

PR #265, opened 2018-10-23
Make randstr field in _scans files deterministic

### LINKED ISSUE #240: do not modify existing `randstr` fields in _scans.tsv
ATM those get repopulated each time even if there is no other change.  That leads to spurious changes/commits without need.

yarikoptic:
Closes #240 

generate  randstr's in _scans files deterministically based on the content of the DICOM header (UIDs) so it would be difficult if not impossible to guess that random string. That should make those _scans.tsv reproducible even if done from scratch

### DIFF
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -292,13 +292,9 @@ def get_formatted_scans_key_row(dcm_fn):
         [ISO acquisition time, performing physician name, random string]
 
     """
-    from heudiconv.external.dcmstack import ds
-    mw = ds.wrapper_from_data(dcm.read_file(dcm_fn,
-                                            stop_before_pixels=True,
-                                            force=True))
+    dcm_data = dcm.read_file(dcm_fn, stop_before_pixels=True, force=True)
     # we need to store filenames and acquisition times
     # parse date and time and get it into isoformat
-    dcm_data = mw.dcm_data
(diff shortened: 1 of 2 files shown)


yarikoptic:
@mgxd @satra - any feedback/concerns about this way... ?

heudiconv/bids.py
```diff
@@ -291,25 +292,27 @@ def get_formatted_scans_key_row(item):
...
         acq_time = datetime.strptime(td, '%H%M%S%Y%m%d').isoformat()
     except AttributeError as exc:
         lgr.warning("Failed to get date/time for the content: %s", str(exc))
         acq_time = None
     # add random string
-    randstr = ''.join(map(chr, sample(k=8, population=range(33, 127))))
+    # But let's make it reproducible by using all UIDs
+    # (might change across versions?)
+    randcontent = u''.join(
+        [getattr(dcm_data, f) for f in sorted(dir(dcm_data))
```
satra: if the `getattr` returns `None` this will fail. here is a suggestion, if `None` is a possibility.

```suggestion
        [getattr(dcm_data, f) or '' for f in sorted(dir(dcm_data))
```

yarikoptic:
Cool, I need to learn how to do suggestions!  Thanks @satra

PR #268, opened 2018-11-01
maint: docker build

mgxd:
* updates Dockerfile in hopes of fixing [failing Docker builds](https://hub.docker.com/r/nipy/heudiconv/builds/)


yarikoptic:
And I wondered where is the change to the script calling neurodocker... Here it is https://github.com/nipy/heudiconv/blob/e4a55513be3711ef1cdbd2f476110619401dc893/utils/gen-docker-image.sh

PR #269, opened 2018-11-01
ENH: use base conda env + conda install numpy and scipy

kaczmarj:
This PR proposes using the very latest version of neurodocker, which allows the use of miniconda's base environment. This PR also installs scipy and numpy from conda to take advantage of openblas.


utils/gen-docker-image.sh
```diff
@@ -4,11 +4,13 @@ set -eu
...
-docker run --rm kaczmarj/neurodocker:0.4.1 generate docker -b neurodebian:stretch -p apt \
+image="kaczmarj/neurodocker:master@sha256:9f7d58f6977cfcd4dd5d1a2e70be4124417206b716d51b7d9a182820157f1bd3"
+
+docker run --rm $image generate docker -b neurodebian:stretch -p apt \
     --dcm2niix version=v1.0.20180622 method=source \
     --install git gcc pigz liblzma-dev libc-dev git-annex-standalone \
     --copy . /src/heudiconv \
-    --miniconda create_env=neuro conda_install="python=3.6 traits>=4.6.0" activate=True \
+    --miniconda use_env=base conda_install="python=3.6 traits>=4.6.0 scipy numpy nomkl" \
       pip_install="/src/heudiconv[all]" \
```
mgxd: could you also make this an editable install so we don't have two copies of the same source code?

mgxd:
thanks @kaczmarj

PR #271, opened 2018-11-08
enh: fallback to private siemens header

### LINKED ISSUE #80: Siemens Csa headers

mgxd:
Closes #80 
Builds off #254 with thanks to @kasbohm

## Additions
* [x] Allow parsing of private CSA header (Siemens)
* [x] Use fallback private CSA field for ProtocolName if not specified

## To consider
* Removing `heudiconv.external.dcmstack` now that [properly handled](https://github.com/moloney/dcmstack/commit/c2ef2f18e85c0463f9bc36e2c9efaaf9a54a6fda)

## To do
* [x] Testing? (data available?)


neurolabusc:
@yarikoptic - I do not have any raw DICOMs like this, but you can create your own with the command `dcmodify -ea "(0018,1030)" *.dcm`. Here is [an example DICOM](https://www.dropbox.com/s/dxi6qci4c2tt1i2/no_protocol_name.zip?dl=0) created by using dcmodify on an image from the [dcm_qa](https://github.com/neurolabusc/dcm_qa) repository.

yarikoptic:
Thank you @neurolabusc for the hint. As much as I like fake/simulated etc data for quick testing, I love real data to possibly "enjoy" the breath of diversion from a common practice/standard whenever a single case is detected ;)

kasbohm:
Thanks, @mgxd  and @yarikoptic for moving the issue forward, much appreciated. We have a few thousand datasets that look like this acquired at different Siemens scanners (Avanto/Skyra/Prisma), and I believe also at least one of the sites we run a multi-site study with has the same issue. Kind of surprised you had not seen it previously. If needed I can try to dig up a dataset for testing, let me know.

mgxd:
thanks for the recommendation @neurolabusc - handy command! and @kasbohm, I've tested with one subject missing `ProtocolName` and conversion looked good - maybe @yarikoptic would take you up on your offer and add to datalad? I'm going to merge this in now

PR #272, opened 2018-11-12
RF: move tests under heudiconv/ module

### LINKED ISSUE #171: move tests inside the heudiconv module
possibly also spreading them around where applicable under sub-modules (e.g. test_heuristics.py could go under heudiconv/heuristics/tests) to facilitate testing of individual components of the system where applicable

yarikoptic:
- to ease maintenance/distribution
- to facilitate testing of installed instances

Closes #171


yarikoptic:
Let it be then!

PR #273, opened 2018-11-12
BF: requires nipype at least 0.12 where Dcm2niix interface was added

yarikoptic:
Some older neurodebian systems still have prior releases of nipype


yarikoptic:
Just a fact, so will merge for now

PR #283, opened 2018-12-04
added decoding of bytes literal strings returned by subprocess during anonymization

### LINKED ISSUE #246: Usage of -o and -a flags is a bit confusing
I'd like to convert dicoms into a BIDS hierarchy, but anonymize the subjectIDs along the way.

When I do this, I provide parameters like this:
-o ./data/BIDS -a ./data/anonBIDS

The result is that all of the anonymized data goes into the anonBIDS directory, but that directory is missing all of the .heudiconv files, while the BIDS directory contains nothing *but* the .heudiconv files. 

Is this expected behavior? It seems not ideal. I think ideally I'd still retain the .heudiconv conversion info, but just with anonymized subject IDs.

### LINKED ISSUE #281: anon-cmd interprets function returns as bytes literal strings
When I call a python script using `anon-cmd`, the new subject id that the python script returns gets interpreted as a bytes literal string by heudivonv so the files end up reading like `sub-b'test001'_task-rest_bold.nii.gz` and `task-rest_bold.json`and `sub-b_scans.tsv`. Does it make sense to change line 108 in `utils.py` to be `return check_output(cmd).strip().decode()` or should I be passing the strings from python to the shell environment differently?

Thanks!

daeh:
Three changes related to the anonymization.
1 - If `subprocess.check_output` returns a bytes literal string it is decoded to utf-8. Fixes #281 
2 - Many anonymization use cases are likely to involve parsing tabular files with a python script, for which pandas is a common strategy. Including the pandas package in the docker file of the vanilla build saves people from needing to roll their own.
3 - The subject info in the .heudiconv directory is now anonymized if an anon_sid is present. Fixes #246


Dockerfile
```diff
@@ -94,6 +94,7 @@ RUN export PATH="/opt/miniconda-latest/bin:$PATH" \
            'traits>=4.6.0' \
            'scipy' \
            'numpy' \
+           'pandas' \
```
yarikoptic: I am a bit uncertain about this one.  Although indeed pandas is a popular robust way to parse .csv, it is quite a heavy dependency.  Would be useful to know how much (relatively) does it blow up the size of the image?

heudiconv/utils.py
```diff
@@ -105,7 +105,14 @@ def dec(obj):
 def anonymize_sid(sid, anon_sid_cmd):
     from subprocess import check_output
     cmd = [anon_sid_cmd, sid]
-    return check_output(cmd).strip()
+    shell_out = check_output(cmd).strip()
+    if isinstance(shell_out, bytes):
+        anon_sid = shell_out.decode()
+    elif isinstance(shell_out, str):
+        anon_sid = shell_out
```
yarikoptic: we still support python2, there `bytes is str` and barely anywhere code has `unicode` in mind for python2.  I wonder if then we would start encountering unicode where we don't expect.
It would probable be worth adding a test for using this functionality, which we do not have seems to me at all.
Would you mind for now just to extend `heudiconv/tests/test_regression.py::test_conversion` with the additional option for heudiconv invocation providing some basic anonimization command?  it would help to make sure that this code works across pythons.
If brave enough, in subsequent commit try to make that function return some proper utf-8 encoded sid, and see if things still work ;)

daeh: Added a pytest with the `--anon-cmd` flag and a variable passed from bash. Also fixed `anonymize_sid()` so that it works with both py2 and 3 (tested with 2.7.15 and 3.6.6).

Dockerfile
daeh: It's 1.599GB without pandas and 1.791GB with pandas. To me that's definitely worth the convenience but totally your call.

mgxd: @daeh could you add it to [`utils/gen-docker-image.sh`](https://github.com/nipy/heudiconv/blob/master/utils/gen-docker-image.sh) as well? We use that script to generate this Dockerfile

heudiconv/utils.py
```diff
@@ -103,9 +103,22 @@ def dec(obj):
 
 
 def anonymize_sid(sid, anon_sid_cmd):
+    import sys
     from subprocess import check_output
+    
     cmd = [anon_sid_cmd, sid]
-    return check_output(cmd).strip()
+    shell_return = check_output(cmd)
```
mgxd: How about condensing this into:
```suggestion
   kwargs = {'encoding': 'utf-8'} if sys.version_info[0] >= 3 else {}
   return check_output(cmd, **kwargs).strip()
```

daeh: definitely better ++

mgxd: looks like `encoding` wasn't added to `check_output` til 3.6 😓 - for the sake of backwards compatibility let's go with your original approach (decoding after subprocess call)

Dockerfile
daeh: ++ thanks, missed that

yarikoptic:
Thanks!

heudiconv/tests/test_regression.py
```diff
@@ -19,16 +19,17 @@
...
+def test_conversion(tmpdir, subject, heuristic, anon_cmd):
     tmpdir.chdir()
     try:
         datadir = fetch_data(tmpdir.strpath, subject)
     except IncompleteResultsError as exc:
         pytest.skip("Failed to fetch test data: %s" % str(exc))
     outdir = tmpdir.mkdir('out').strpath
 
-    args = gen_heudiconv_args(datadir, outdir, subject, heuristic)
+    args = gen_heudiconv_args(datadir, outdir, subject, heuristic, anon_cmd)
```
yarikoptic, line 32: in the light of #509 I now wonder if it all worked out correctly in this PR, since if I get it right - with the anonymization command we should have no longer got file names formatted (below) according to `subject` but rather to anonymized version of it

PR #290, opened 2018-12-21
Initial appveyor setup config

yarikoptic:
Some users tries to use heudiconv on windows, which is not entirely impossible. But we do not know how not ready we are.


yarikoptic:
brute force attempt failed, not time/interest to work on it. If someone would like to pick up -- go ahead!

PR #292, opened 2019-01-04
v0.5.2 Release

yarikoptic:
If tests pass, I will release unless objections are expressed

- [x] install test data


mgxd:
is #200 good to go? if so, would be a nice addition :)

Issue #286, opened 2018-12-05
concurrent datalad operations?


mgxd:
When concurrently converting across multiple participants while using `--datalad`, heudiconv terminates before completion due to git `index.lock` files and the dataset is not put under version control.

Perhaps [OpenNeuroOrg/datalad-service](https://github.com/OpenNeuroOrg/datalad-service) can be some inspiration

yarikoptic:
Yeap, a known limitation of the git and even you greater degree of datalad. No easy way around unfortunately ATM.
OpenNeuro datalad backend unlikely to have this issue since unlikely had multiple datalad invocations dinner on parallel in the same dataset, and they don't manage a hierarchy of the datasets.
They only workaround is to not use --datalad mode ATM and perform datalad add later for the entire result. This would though cause some files lacking their metadata tags

satra:
one possibility may be to stage the info into separate uuid branches and then merge it later in one go after conversion.

PR #304, opened 2019-02-05
FIX: SLURM queue support

### LINKED ISSUE #302: Issues w/ slurm
Greetings, 

heudiconv version 0.5.3. 

I am trying to run a basic heudiconv on one subject for testing w/ slurm (running on two nodes, partition name = heudiconv) to create my heuristic. 

1) Command as follows: 

heudiconv -d /srv/lab/fmri/mft/dicom4bids/data/{subject}*/* -f /srv/lab/fmri/mft/dicom4bids/data/convertall.py -s sub01 -q heudiconv -o /srv/lab/fmri/mft/dicom4bids/data/output/

This yields: 
INFO: Running heudiconv version 0.5.3
INFO: Need to process 1 study sessions
  File "/usr/local/anaconda3/envs/fmri/bin/heudiconv", line 11, in <module>
    sys.exit(main())
  File "/usr/local/anaconda3/envs/fmri/lib/python3.5/site-packages/heudiconv/cli/run.py", line 125, in main
    process_args(args)
  File "/usr/local/anaconda3/envs/fmri/lib/python3.5/site-packages/heudiconv/cli/run.py", line 296, in process_args
    study_outdir,
UnboundLocalError: local variable 'study_outdir' referenced before assignment

2) Now, adding "-l unknown" to the above command executes, but does not create a .heudiconv folder: 
INFO: Running heudiconv version 0.5.3
INFO: Need to process 1 study sessions
WARNING: Skipping unknown locator dataset

3) Curiously, removing the -q flag and the -l flag executes the command properly and creates .heudiconv under /output:
INFO: Running heudiconv version 0.5.3
INFO: Need to process 1 study sessions
INFO: PROCESSING STARTS: {'session': None, 'outdir': '/srv/lab/fmri/mft/dicom4bids/data/output/', 'subject': 'sub01'}
INFO: Processing 4976 dicoms
INFO: Analyzing 4976 dicoms
INFO: Generated sequence info with 10 entries
INFO: PROCESSING DONE: {'session': None, 'outdir': '/srv/lab/fmri/mft/dicom4bids/data/output/', 'subject': 'sub01'}

4) Anyone else experienced these issues? How come that adding -q breaks the locator inference? I still believe that a working scheduler integration for heudiconv will be extremely valuable w/ increasing data sets and I am happy to contribute. Perhaps a starting point would be to clarify what the -l flag actually requires. In the help menu for -l, it says: study path under outdir. If provided, it overloads the value provided by the heuristic. If --datalad is enabled, every directory within locator becomes a super-dataset thus establishing a hierarchy. Setting to "unknown" will skip that dataset. 

Thanks in advance!

mgxd:
Closes #302.

Since this was forgotten/neglected after #87, re-enables support for SLURM submission.

`queue_conversion` was refactored to read command line arguments, and pass them to the SLURM call script. This will help with less maintenance as options are changed or added, with the trade-off of requiring a command-line submission.

`--sbargs` was renamed to `--queue-args` to allow for additional queue support down the line.

### DIFF
--- a/heudiconv/cli/run.py
+++ b/heudiconv/cli/run.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 import os
 import os.path as op
 from argparse import ArgumentParser
@@ -215,12 +217,11 @@ def get_parser():
     parser.add_argument('--dcmconfig', default=None,
                         help='JSON file for additional dcm2niix configuration')
     submission = parser.add_argument_group('Conversion submission options')
-    submission.add_argument('-q', '--queue', default=None,
-                            help='select batch system to submit jobs to instead'
-                                 ' of running the conversion serially')
-    submission.add_argument('--sbargs', dest='sbatch_args', default=None,
-                            help='Additional sbatch arguments if running with '
-                                 'queue arg')
+    submission.add_argument('-q', '--queue', choices=("SLURM", None),
+                            default=None,
+                            help='batch system to submit jobs in parallel')
+    submission.add_argument('--queue-args', dest='queue_args', default=None,
+                            help='Additional queue arguments')
     return parser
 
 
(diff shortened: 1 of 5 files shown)


satra:
does this submit each participant as a separate job? and if so, does it handle datalad integration?

i do think the simplest option is to create separate branches for each heudiconv conversion in parallel and then merging the branches together as part of the parent process. or using a lockfile (see the lockfile in pydra task as an example). the call to datalad should only happen under a lock.

mgxd:
yes, it will submit each participant-session pair as a separate job. it will allow `--datalad` args to be passed, but will probably run into the same issue as in #286 - I didn't want to take that on yet.

I think both of those options you stated are promising, though i'm leaning slightly to the lock file one.

mgxd:
since this wasn't working to begin with, let's merge it in

PR #306, opened 2019-02-06
fix: embedding issue

mgxd:
Closes #291.

Fixes an issue where dcmstack output was not indexed properly (due to the change in https://github.com/moloney/dcmstack/commit/3815ea225e641af7c6c3a4c454b7dcd1bd2ede66). This "silent" error was avoided when using the `--minmeta` flag.

### Changes
* `stack` is set as the first iterated item of dcmstack's `parse_and_stack()`
* Parameter `force` of `embed_nifti()` was removed and automatically applied within.
* Test for various uses of `embed_nifti()`

PR #308, opened 2019-02-13
RF: embed_nifti to extract_more_metadata_for_nifti

yarikoptic:
That functionality was not used as far as I see it. More detail in the CHANGELOG change

Was reviewing #306 (better later than never) and just couldn't make it why we still have that additional "make nifti from dicoms" functionality if not actually used, or was I wrong?  Since nothing is embedded, I renamed that function, and ideally need to rename `embed_metadata_from_dicoms`.
So I wondered, what prov info we are collecting in there while running that function? may be `embed_metadata_from_dicoms` should just be melded with `extract_more_metadata_for_nifti`? @satra ?

### DIFF
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -353,33 +353,34 @@ def compress_dicoms(dicom_list, out_prefix, tempdirs, overwrite):
     return outtar
 
 
-def embed_nifti(dcmfiles, niftifile, infofile, bids_info, min_meta):
+def extract_more_metadata_for_nifti(dcmfiles, niftifile, infofile, bids_info=None, min_meta=False):
     """
 
-    If `niftifile` doesn't exist, it gets created out of the `dcmfiles` stack,
-    and json representation of its meta_ext is returned (bug since should return
-    both niftifile and infofile?)
-
-    if `niftifile` exists, its affine's orientation information is used while
-    establishing new `NiftiImage` out of dicom stack and together with `bids_info`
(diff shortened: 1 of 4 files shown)


heudiconv/dicoms.py
```diff
@@ -460,7 +456,7 @@ def embed_metadata_from_dicoms(bids, item_dicoms, outname, outname_bids,
     embedfunc = Node(Function(input_names=['dcmfiles', 'niftifile', 'infofile',
                                            'bids_info', 'min_meta'],
                               output_names=['outfile', 'meta'],
```
mgxd:
```suggestion
                              output_names=['meta'],
```

heudiconv/dicoms.py
```diff
@@ -392,20 +393,15 @@ def embed_nifti(dcmfiles, niftifile, infofile, bids_info, min_meta):
...
         new_nii = stack.to_nifti(voxel_order=''.join(axcodes), embed_meta=True)
-        meta = ds.NiftiWrapper(new_nii).meta_ext.to_json()
-
-    meta_info = None if min_meta else json.loads(meta)
+        meta_file = ds.NiftiWrapper(new_nii).meta_ext.to_json()
+        meta_info = json.loads(meta_file)
+    else:
+        meta_info = {}
 
     if bids_info:
```
mgxd: We should also clean up this check:
- L408 - 412 should be removed
- L414 as well

PR #333, opened 2019-04-17
Add an examplar of dcm_file path in seqinfo

bpinsard:
Add an examplar of a dicom file in seqinfo to allow access more information in infotodict.

### DIFF
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -91,6 +91,7 @@ def create_seqinfo(mw, series_files, series_id):
         date=dcminfo.get('AcquisitionDate'),
         series_uid=dcminfo.get('SeriesInstanceUID'),
         time=dcminfo.get('AcquisitionTime'),
+        example_dcm_file_path=series_files[0],
     )
     return seqinfo
 
--- a/heudiconv/parser.py
+++ b/heudiconv/parser.py
@@ -97,7 +97,7 @@ def get_extracted_dicoms(fl):
         for tm in tmembers:
             tm.mode = 0o700
         # get all files, assemble full path in tmp dir
-        tf_content = [m.name for m in tmembers if m.isfile()]
+        tf_content = sorted([m.name for m in tmembers if m.isfile()])
         # store full paths to each file, so we don't need to drag along
         # tmpdir as some basedir
         sessions[session] = [op.join(tmpdir, f) for f in tf_content]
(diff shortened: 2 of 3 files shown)


satra:
@bpinsard - i think this is a good start.  it would be nice to add a test, an example heuristic that uses the dicom info from the file.

@yarikoptic @mgxd what do you think?

more generally would it be useful to pass paths to all files in a group? obviously we can't store this in the dicominfo.tsv, but we are storing this in the filegroup.json, so could we simply pass that along with seqinfo?

yarikoptic:
I am worrying of storing full paths there, since then it would prevent any relocation etc.
This sounds like a workaround for some specific use-case of missing information in the seqinfo. So what was it @bpinsard ?

yarikoptic:
@bpinsard could you please followup with description of the problem/use-case you are trying to address here?

bpinsard:
At the acquisition center we store information in the following dicom tags:
- ReferringPhysicianName
- StudyDescription
- PatientName
Which I use to find the subject_id,  as well as research group, study, to build the locator in infotoids.

And I also extract information from:
- `dcm.csa_header['tags']['PhaseEncodingDirectionPositive']['items'][0]`
- `dcm.dcm_data.InPlanePhaseEncodingDirection`
- `dcm.dcm_data.ImageComments`
to identify the sequence, if it is a sbref, and phase-encoding direction more reliably than from the ProtocoleName or else.

I don't know if this is relevant to other people. If not we can close that.

yarikoptic:
hold on -- [we are already storing a filename and a directory for the file](https://github.com/nipy/heudiconv/blob/master/heudiconv/dicoms.py#L205):
```
        info = SeqInfo(
            total,
            op.split(series_files[0])[1],
            series_id,
            op.basename(op.dirname(series_files[0])),
```

But why don't we also extend the SeqInfo record with found to be useful additional fields to contain the ones you desire so there is no need to re-load the dicom?

bpinsard:
SeqInfo does include filename and the folder but the later is only the name of the last folder in the path that contains the dicom, not the full path. Is there a specific reason for that? If we remove the `basename` call, I think this leaves the opportunity to reread a single dicom per series, even if not optimal, and extract any kind of information without creating a very large SeqInfo structure.

yarikoptic:
> SeqInfo does include filename and the folder but the later is only the name of the last folder in the path that contains the dicom, not the full path.

oh, right -- I was too fast in misreading that code.  Thanks for the explanation ;)

> Is there a specific reason for that?

probably again just to not leak absolute paths there, and provide minimal sufficient (someone might want to use filename, some name of the immediate directory with dicoms) information for the heuristic.

yarikoptic:
I believe #581 / #637 (whatever comes first) will provide the alternative solution to this which would not need to leak the entire filename in the metadata

PR #339, opened 2019-05-10
RF+BF: make code a bit more pythonic, and change comparison > 1 not >=1 for deciding multi-echo

yarikoptic:
AFAIK a single echo time is not making it multi-echo, or does it?

I might go more through that code, so this PR will be evolving unless rapidly merged ;)

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -506,19 +506,14 @@ def save_converted_files(res, item_dicoms, bids, outtype, prefix, outname_bids, 
         #   series. To do that, the most straightforward way is to read the
         #   echo times for all bids_files and see if they are all the same or not.
 
-        # Check for echotime information
-        echo_times = set()
+        # Check for varying echo times
+        echo_times = set(
+            load_json(b).get('EchoTime', None)
+            for b in bids_files
+            if b
+        )
 
-        for bids_file in bids_files:
-            if bids_file:
-                # check for varying EchoTimes
-                echot = load_json(bids_file).get('EchoTime', None)
-                if echot is not None:
-                    echo_times.add(echot)
-
-        # To see if the echo times are the same, convert it to a set and see if
-        # only one remains:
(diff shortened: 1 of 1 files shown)


heudiconv/convert.py
```diff
@@ -506,19 +506,14 @@ def save_converted_files(res, item_dicoms, bids, outtype, prefix, outname_bids,
...
-            if bids_file:
-                # check for varying EchoTimes
-                echot = load_json(bids_file).get('EchoTime', None)
-                if echot is not None:
-                    echo_times.add(echot)
-
-        # To see if the echo times are the same, convert it to a set and see if
-        # only one remains:
-        is_multiecho = len(echo_times) >= 1 if echo_times else False
+        is_multiecho = len(echo_times) > 1
```
yarikoptic: FWIW:
original condition for `>=1` was added in 266de4a9a5d755455f348cb66ef025337a8d3b9a changing previous, IMHO also incorrect test:
```patch
         # To see if the echo times are the same, convert it to a set and see if
-        #   only one remains:
-        multiecho = False
-        if echo_times:
-            multiecho = len(set(echo_times)) == 1
+        # only one remains:
+        is_multiecho = len(echo_times) >= 1 if echo_times else False
```
which gives me some more confidence that my change is correct or my understanding is completely flawed ;)

yarikoptic:
attn @pvelasco - I am here adjusting the logic for deciding if sequence multi-echo, please review

pvelasco:
Hi @yarikoptic,
You are right: if `len(echo_times)==1`, then `is_multiecho` should be `false`.  So the changes introduced in [3250283](https://github.com/nipy/heudiconv/commit/3250283ff7be3f478b35b1fb4e504cec42403ddc#diff-a48009f9efe6cc64ffdb74d06e831296) were wrong.
Your changes should fix it.

yarikoptic:
thanks @pvelasco and @mgxd. Let's proceed then ;)

PR #343, opened 2019-05-14
Reproin fix for Philips 

cni-md:
As discussed here [https://github.com/ReproNim/reproin/issues/32#issue-439932420]( https://github.com/ReproNim/reproin/issues/32#issue-439932420)

I modified the heuristics to remove the Philips WIP prefix and a substitution for (date) to {date} as session label.
The later will only work if heudiconv is called with --file ..., not if the -d flag is used.

- [x] adjust docstring to mention `()` -> `{}` replacement and WIP stripping
- [x] add to the tests to assure that stripping/mapping works

### DIFF
--- a/heudiconv/heuristics/reproin.py
+++ b/heudiconv/heuristics/reproin.py
@@ -888,8 +888,9 @@ def parse_series_spec(series_spec):
 
         # sanitize values, which must not have _ and - is undesirable ATM as well
         # TODO: BIDSv2.0 -- allows "-" so replace with it instead
-        value = str(value).replace('_', 'X').replace('-', 'X')
-        value = str(value).replace('(', '{').replace(')', '}') # for Philips
+        value = str(value) \
+            .replace('_', 'X').replace('-', 'X') \
+            .replace('(', '{').replace(')', '}')  # for Philips
 
         if key in ['ses', 'run', 'task', 'acq']:
             # those we care about explicitly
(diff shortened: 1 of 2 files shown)


yarikoptic:
thank you!!! I added 2 TODOs in description we need to do before merging

cni-md:
@yarikoptic 
My testfile
https://cloud.lin-magdeburg.de/s/jpj7GDjnx8weWrf

I guess one of the dicoms is sufficient. It has the "WIP " and the -ses_(date) in the name.

yarikoptic:
well, for the tests of this PR real data is not even needed (I will push to this PR shortly), but shouldn't hurt ;-)   I could share it from http://datasets.datalad.org/?dir=/dicoms e.g. under `cni-md/` subdatasets or any alike, if you don't mind? then we could add additional tests

BTW I see
```
WARNING: Could not figure out where to stick 1 sequences: ['102-WIP MPR - anat-SmartBrain_ses-(date)_acq-32channel']
```
so that is some post-processed (MPR package?) version of the file.  Do you care to store them as well? we could in principle stick them in as  `_proc-MPR` may be (one more use case/motivation to get back to https://github.com/bids-standard/bids-specification/pull/105)

cni-md:
You can use the data if you like. 

The SmartBrain is an autoalign localizer and has some MPR reconstructions. I do not want to store them. It was just the shortest MR protocoll I have.

yarikoptic:
ok, I have adjusted docstring so we could merge (when tests pass) I think

Issue #340, opened 2019-05-13
Question regarding batch processing


stilley2:
### Summary

The [docs](https://heudiconv.readthedocs.io/en/latest/usage.html#batch-jobs) say that batch conversion with multiple parallel calls to heudiconv is possible. However, it's not clear what this means for files in the top level of the output directory (e.g. participants.tsv, task*.json). It would seem to me that there would be race conditions updating these files.

I skimmed through the code, and a potential workaround seems to be to ignore all the top level files after a batch run and repopulate them manually. This requires that the lower level json files are not altered based on the top level files. In other words, this requires that the subjects subdirectories do _not_ rely on inheritance from the top level directory. Is this an accurate assumption?

Assuming my understanding of all of this is correct, it seems the easiest way to allow batch processing is to include a flag that prevents writing of top level files, and a subcommand that can generate these after all the batches have complete. Thoughts?

Thanks!

### Platform details:

Choose one:
- [x] Local environment

- [ ] Container

- Heudiconv version: 

0.5.4

mgxd:
Hi @stilley2 - thank you for the report.

>It would seem to me that there would be race conditions updating these files.

Yep, this is a less common but totally possible pitfall with this approach. Though this could be remedied by creating a lockfile whenever modifying these top-level files to halt concurrent access and then removing it to allow the next process to proceed. 

Would you have any interest in adding this feature?

stilley2:
Sure I'll take a stab at in. In the meantime, is my assumption that the subject level files are not altered when the top level files are generated correct? That will allow me to proceed with the current version and just ignore the top level files.

stilley2:
I looked into file locking, and it seems tricky to come up with a cross platform solution that works on all operating systems and on distributed filesystems such as those in a cluster environment. I think the best plan is to just add a flag that turns off writing these top-level files, and then the user can use the `populate-templates` command after the batch processing. I've created #344 to do this.

satra:
@stilley2 - try this one: https://pypi.org/project/filelock/

stilley2:
@satra, I just tested that on a cluster environment, and it only worked when the processes were on the same node. So on the system I have access to this would not work with SLURM when requesting more than one node. I understand this is likely filesystem dependent, but I don't think we want to rely on such behavior.

satra:
@stilley2 - thank you for testing. did you use `SoftFileLock` ?

stilley2:
[...]

PR #344, opened 2019-05-14
Add --skiptop flag to prevent writing top level bids files

stilley2:
This is useful in batch processing to prevent potential race
conditions. The batch processing example has also been
modified to use this flag.

See #340

### DIFF
--- a/heudiconv/cli/run.py
+++ b/heudiconv/cli/run.py
@@ -138,6 +138,7 @@ def main(argv=None):
         args.bids_options = args.bids
         args.bids = True
     else:
+        args.bids_options = []
         args.bids = False
     if args.debug:
         lgr.setLevel(logging.DEBUG)
--- a/heudiconv/cli/run.py
+++ b/heudiconv/cli/run.py
@@ -18,7 +18,8 @@ lgr = logging.getLogger(__name__)
 
 INIT_MSG = "Running {packname} version {version}".format
 
-BIDS_OPTIONS = [('notop', 'Skip creating of top-level bids files. '
+BIDS_OPTIONS = [('help', 'Display this help message'),
+                ('notop', 'Skip creating of top-level bids files. '
                           'Useful when running in batch mode to prevent '
                           'possible race conditions.')]
 
@@ -113,7 +114,7 @@ def help_bids():
 def main(argv=None):
     parser = get_parser()
     args = parser.parse_args(argv)
-    if args.help_bids:
+    if 'help' in args.bids:
         help_bids()
         sys.exit(1)
     # exit if nothing to be done
@@ -212,7 +213,7 @@ def get_parser():
     parser.add_argument('-b', '--bids', nargs='*',
                         metavar=('BIDSOPTION1', 'BIDSOPTION2'),
                         help='flag for output into BIDS structure. '
-                        'Can also take bids specific options. Use --help-bids '
+                        'Can also take bids specific options. Use --bids help '
                         'for more information.')
     parser.add_argument('--overwrite', action='store_true', default=False,
                         help='flag to allow overwriting existing converted files')
@@ -253,8 +254,6 @@ def get_parser():
                             help='Additional queue arguments passed as '
                             'single string of Argument=Value pairs space '
                             'separated.')
-    parser.add_argument('--help-bids', action='store_true', dest='help_bids',
-                        help='Display bids specific help.')
     return parser
 
 
--- a/heudiconv/cli/run.py
+++ b/heudiconv/cli/run.py
@@ -18,11 +18,6 @@ lgr = logging.getLogger(__name__)
 
 INIT_MSG = "Running {packname} version {version}".format
 
-BIDS_OPTIONS = [('help', 'Display this help message'),
-                ('notop', 'Skip creating of top-level bids files. '
-                          'Useful when running in batch mode to prevent '
-                          'possible race conditions.')]
-
 
 def is_interactive():
    """Return True if all in/outs are tty"""
@@ -103,20 +98,9 @@ def process_extra_commands(outdir, args):
     return
 
 
-def help_bids():
-    sys.stderr.write('bids specific options can be passed after the bids flag.\n'
-                     'For example, "--bids notop".\n'
-                     'The currently supported options are:\n')
-    for option, helpstr in BIDS_OPTIONS:
-        sys.stderr.write('{}: {}\n'.format(option, helpstr))
-
-
 def main(argv=None):
     parser = get_parser()
     args = parser.parse_args(argv)
-    if 'help' in args.bids:
-        help_bids()
-        sys.exit(1)
     # exit if nothing to be done
     if not args.files and not args.dicom_dir_template and not args.command:
         lgr.warning("Nothing to be done - displaying usage help")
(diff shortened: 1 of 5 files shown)


heudiconv/cli/run.py
```diff
@@ -222,6 +222,10 @@ def get_parser():
                             help='Additional queue arguments passed as '
                             'single string of Argument=Value pairs space '
                             'separated.')
+    parser.add_argument('--skiptop', action='store_true',
+                        help='Skip creating of top-level bids files. '
+                        'Useful when running in batch mode to prevent '
+                        'possible race conditions.')
```
yarikoptic: our --help is already LONG enough and I expect that only more options would come to alter BIDS handling behavior.  What if we "convert" `--bids` option into something more useful?  
Especially since I think we are no longer using positional args, I think it should be easy to do:

<details>
<summary>Here is a prototype showing that we could use `args='*'` mode:</summary>

```shell
~/proj/repronim/reproin > cat /tmp/argparse_.py          
__version__ = '1'
from argparse import ArgumentParser
docstr = ("""Example:
         heudiconv -d rawdata/{subject} -o . -f heuristic.py -s s1 s2 s3""")
parser = ArgumentParser(description=docstr)
parser.add_argument('--version', action='version', version=__version__)

parser.add_argument('-b', '--bids', nargs='*',
                    choices=['notop', 'nostriptimes'],
                    help="""\
                    flag for output into BIDS structure.  Additional options which
                    could be specified:  notop - no top level metadata-files, 
                    nostriptimes - do not strip times from side-car files""")

out = parser.parse_args()
print(out)


~/proj/repronim/reproin > python /tmp/argparse_.py --help
usage: argparse_.py [-h] [--version]
                    [-b [{notop,nostriptimes} [{notop,nostriptimes} ...]]]

Example: heudiconv -d rawdata/{subject} -o . -f heuristic.py -s s1 s2 s3

optional arguments:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
  -b [{notop,nostriptimes} [{notop,nostriptimes} ...]], --bids [{notop,nostriptimes} [{notop,nostriptimes} ...]]
                        flag for output into BIDS structure. Additional
                        options which could be specified: notop - no top level
                        metadata-files, nostriptimes - do not strip times from
                        side-car files

~/proj/repronim/reproin > python /tmp/argparse_.py       
Namespace(bids=None)                               

~/proj/repronim/reproin > python /tmp/argparse_.py -b
Namespace(bids=[])

~/proj/repronim/reproin > python /tmp/argparse_.py -b notop
Namespace(bids=['notop'])

~/proj/repronim/reproin > python /tmp/argparse_.py -b notop -b nostriptimes
Namespace(bids=['nostriptimes'])

~/proj/repronim/reproin > python /tmp/argparse_.py -b notop nostriptimes 
Namespace(bids=['notop', 'nostriptimes'])

~/proj/repronim/reproin > python /tmp/argparse_.py -b notop nostriptimes nonsense
usage: argparse_.py [-h] [--version]
                    [-b [{notop,nostriptimes} [{notop,nostriptimes} ...]]]
argparse_.py: error: argument -b/--bids: invalid choice: 'nonsense' (choose from 'notop', 'nostriptimes')

```
</details>

so we can discriminate when it is not specified (`None` -- current `False`) for when specified but without options (`[]` -- current True).  So, if instead of carrying around a bool `bids` we just carry `bids` a list of options with an empty list for `False`, `['no-options']` if no options were provided, and otherwise just a full list of options, we could later add other options and remain compatible while testing `if bids:` and check for options `if bids and 'notop' not in bids:` where needed.

stilley2: Ok I restructured it with your suggestion. Internally, I set args.bids to either True of False and add args.bids_options, but from the point of view of the user it works as you suggested.

heudiconv/cli/run.py
```diff
@@ -222,6 +253,8 @@ def get_parser():
                             help='Additional queue arguments passed as '
                             'single string of Argument=Value pairs space '
                             'separated.')
+    parser.add_argument('--help-bids', action='store_true', dest='help_bids',
+                        help='Display bids specific help.')
```
yarikoptic: why not to just add "help" option to `--bids` and react on it?  part of the goal in my suggestion was not to expand the cmdline API options list more ;-)

heudiconv/convert.py
```diff
@@ -80,7 +80,7 @@ def conversion_info(subject, outdir, info, filegroup, ses):
 
 def prep_conversion(sid, dicoms, outdir, heuristic, converter, anon_sid,
                    anon_outdir, with_prov, ses, bids, seqinfo, min_meta,
-                   overwrite, dcmconfig):
+                   overwrite, dcmconfig, bids_options):
```
yarikoptic: that is what I was hoping to avoid -- needing to eventually add/adjust signature of every function which now gets `bids` but might eventually need to get access to  `bids_options`...

but may be making it more obvious at the cost of some duplication would be better than my more obscure suggestion... Let's see what others thing

heudiconv/cli/run.py
stilley2: Ah yes that's a good ideal. Will do.

heudiconv/convert.py
stilley2: Hmm I see what you're saying. I wanted to keep bids as a boolean because I think it's more obvious to people and prevents us having to check every usage of args.bids. (e.g., if bids is an empty list, we would want to calculate bids stuff, but `if args.bids:` would no longer work.) Therefore a change is required in either case, and I feel the `bids` + `bids_options` method is more explicit and clear. I'm curious what others thing.

mgxd: Yeah, I'd like to avoid additional parameters to the signatures whenever possible. Though if we're changing the variable type + intention, it may be worthwhile to rename it to something more readable - I like `bids_options` or `bids_opts`. Most of the current checks can be ported rather easily
```python
if args.bids_opts is not None:  # if args.bids
   if ['notopdirs'] in args.bids_opts:  # special case
       ...
```

heudiconv/cli/run.py
```diff
@@ -98,9 +103,20 @@ def process_extra_commands(outdir, args):
     return
 
 
+def help_bids():
```
mgxd: This can be added to the argument's `desc` field instead

heudiconv/cli/run.py
```diff
@@ -112,6 +128,19 @@ def main(argv=None):
         random.seed(args.random_seed)
         import numpy
         numpy.random.seed(args.random_seed)
+    # Ensure only supported bids options are passed
+    allowed_options = [option for option, _ in BIDS_OPTIONS]
```
mgxd: argparse's `choices` will check if each input is valid, so this isn't needed

heudiconv/cli/run.py
stilley2: Refactored as per @mgxd's suggestion and just included all the help in the argparse help field

heudiconv/convert.py
stilley2: Refactored to use bids_options internally

heudiconv/cli/run.py
stilley2: I added bids options information to the `help` field (as far as I can tell `desc` is only parser level).

heudiconv/cli/run.py
stilley2: Done.

yarikoptic:
a possible idea to followup with https://github.com/nipy/heudiconv/issues/349

yarikoptic:
I think not

Issue #144, opened 2018-02-01
Assertions and other 


mfalkiewicz:
Dear heudiconv developers,
  thank you for coming up with this awesome project. I am trying to use heudiconv now to convert massive datasets, often with several thousands of subjects. As you know, such datasets are often messy. Running a pass through the data with heudiconv often takes several hours. Unfortunately, heudiconv seems to crash on almost every inconsistency it encounters. For instance:

> INFO: PROCESSING STARTS: {'session': None, 'outdir': '/home/homeGlobal/mfalkiewicz/largefs/dataset_conversion/1000Gehirne/', 'subject': '997128'}
INFO: Processing 3376 dicoms
INFO: Analyzing 3376 dicoms
Traceback (most recent call last):
  File "/usr/bin/heudiconv", line 2197, in <module>
    main()
  File "/usr/bin/heudiconv", line 2189, in main
    return _main(args)
  File "/usr/bin/heudiconv", line 2052, in _main
    min_meta=args.minmeta)
  File "/usr/bin/heudiconv", line 1364, in convert_dicoms
    grouping=None,  # no groupping
  File "/usr/bin/heudiconv", line 425, in group_dicoms_into_seqinfos
    assert studyUID == file_studyUID
AssertionError

This error popped out after waiting for 3 hours and processing of 17 subjects (the actual processing started after 3 hours - data for ~1500 subjects). I think the default behavior in this setting should be dumping the information about failed subject to a text file and continuation for other subjects. 

The other problem that I encountered are DICOM folder templates. heudiconv crashes when it encounters a folder instead of a dicom. I think in this situation it should simply continue the exectution.

Unfortunately, I do not have time to get deeply into the logic of the above mentioned assertion, nor the exact mechanisms of scanning files through dicoms. I would be very greatful if you could fix these features. This software is extremely useful!

yarikoptic:
run with --dbg flag so you end up in pdb upon failure and could check details on what is going on?

mgxd:
@mfalkiewicz 

can you share the command you are using and how your data is set up? we may be able to optimize this and make the conversion much faster and more efficient

yarikoptic:
knock knock @mfalkiewicz -- without further info we wouldn't be able to reproduce or figure really out why is this happening and would need just to close this issue unresolved

mfalkiewicz:
My apologies guys, I will provide the details in the coming days.

mfalkiewicz:
Hello,
[...]

PR #345, opened 2019-05-15
Add PDT2 as valid multi echo sequence type

nicholst:
As suggested by @mgxd in https://github.com/nipy/heudiconv/issues/144, adding PDT2 as a multiecho modality label... WIP, still testing the mod.

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -572,7 +572,7 @@ def save_converted_files(res, item_dicoms, bids, outtype, prefix, outname_bids, 
                 echo_number = fileinfo.get('EchoNumber', 1)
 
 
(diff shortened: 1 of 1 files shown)


nicholst:
I can confirm that this small change works as intended... PD-T2 pairs are now labeled appropriately with `echo-1` and `echo-2` instead of having a 1 or 2 appended at the end.

yarikoptic:
related: https://github.com/nipy/heudiconv/pull/339 since adjusting multi-echo logic, but I think should be "compatible" ;)

mgxd:
@nicholst thanks for the patch!

@yarikoptic since this is a small addition, i'll just merge it in now

PR #348, opened 2019-05-18
Acquire a lock prior to top level bids file modification

stilley2:
Uses SoftLockFile from py-filelock. Potentially addresses #340, although I'm not sure if SoftFileLock is guaranteed to work on distributed filesystems.

Currently the timeout is hardcoded, but if #344 gets merged then we can use the new method of specifying bids options to make the timeout configurable.

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -1,3 +1,4 @@
+from filelock import SoftFileLock
 import os
 import os.path as op
 import logging
@@ -32,6 +33,8 @@ from .dicoms import (
     compress_dicoms
 )
 
+LOCKFILE = 'heudiconv.lock'
+LOCKFILE_TIMEOUT = 10
 lgr = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


heudiconv/convert.py
```diff
@@ -202,14 +205,16 @@ def prep_conversion(sid, dicoms, outdir, heuristic, converter, anon_sid,
         clear_temp_dicoms(item_dicoms)
 
     if bids:
```
mgxd: We should log a warning to users if this file exists before entering the context manager, maybe something like:

> Existing lockfile found in {path to LOCKFILE} - waiting the lock to be released. To set a timeout limit, set the HEUDICONV_FILELOCK_TIMEOUT environmental variable to a value in seconds. If this process hangs, it may require a manual deletion of the {LOCKFILE}.

heudiconv/convert.py
```diff
@@ -1,3 +1,4 @@
+from filelock import SoftFileLock
```
yarikoptic: Note: I personally not familiar with `filelock` (used [fasteners](https://github.com/harlowja/fasteners) when needed), but [it](https://github.com/benediktschmitt/py-filelock) seems to be alive, so ok

yarikoptic:
@stilley2 - would you have time in the near future to finalize this PR, i.e. to address @mgxd comments?  It seems that more ppl are running into #362 so we better have it addressed, and locking is the way to go, so thank you for your contribution!

heudiconv/convert.py
```diff
@@ -210,14 +212,23 @@ def prep_conversion(sid, dicoms, outdir, heuristic, converter, anon_sid,
...
-        populate_bids_templates(anon_outdir,
-                                getattr(heuristic, 'DEFAULT_FIELDS', {}))
+        lockfile = op.join(anon_outdir, LOCKFILE)
+        if op.exists(lockfile):
+            lgr.warning("Existing lockfile found in {0} - waiting for the "
+                        "lock to be released. To set a timeout limit, set "
+                        "the HEUDICONV_FILELOCK_TIMEOUT environmental variable "
+                        "to a value in seconds. If this process hangs, it may "
+                        "require a manual deletion of the {0}.".format(lockfile))
+        timeout = os.environ.get("HEUDICONV_LOCKFILE_TIMEOUT", -1)
```
mgxd:
```suggestion
        timeout = os.getenv("HEUDICONV_LOCKFILE_TIMEOUT", -1)
```

PR #350, opened 2019-05-20
sty: cleanup codebase

mgxd:
Some linting for the package:
* black autolinter
* flake8


yarikoptic:
eh, sorry I have missed this one @mgxd ...  but there is always a danger of doing any of such PRs while there are active other PRs - would cause conflicts, pains, etc.  If you think it is still worthwhile, let's first push toward having other active PRs reach merge/closed state and then do such a big cleanup

mgxd:
totally agree - got stale on my end.

I'll relint after some changes come this way (mainly #359)

PR #351, opened 2019-05-22
fix: dcm2niix outname

mgxd:
This changes the `dcm2niix` outname from an erroneous substring to the heuristic prefix. Specifically, it will change the call from
```
dcm2niix -b y -f func ...
```
to
```
dcm2niix -b y -f /path/to/bids/sub-1/func/sub-1_task-blah_bold ...
```

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -418,7 +418,7 @@ def nipype_convert(item_dicoms, prefix, with_prov, bids, tmpdir, dcmconfig=None)
     convertnode = Node(Dcm2niix(from_file=fromfile), name='convert')
     convertnode.base_dir = tmpdir
     convertnode.inputs.source_names = item_dicoms
-    convertnode.inputs.out_filename = op.basename(op.dirname(prefix))
+    convertnode.inputs.out_filename = prefix
 
     if nipype.__version__.split('.')[0] == '0':
         # deprecated since 1.0, might be needed(?) before


yarikoptic:
Did you forget `y` in your example after `-b`? Just making sure I am not missing something

yarikoptic:
This suggests that you will be writing into a directory nearby original file... That one can be read only

What problem are you trying to fix?

yarikoptic:
Oh, scrape my comment about original directory... But knowing the problem/reason would help. I would have also tried to look into history of may be commit which changed that line tells why only the basename was considered. I would guess - to not leave junk around if conversion fails etc

mgxd:
@yarikoptic this will accomplish 2 things:

1) make it clear to users which key is currently being converted
2) avoid dcm2niix having to add prefixes to outputs with the same name if multiple outfiles are generated
(`-o func` could generate `func.nii` `func1.nii` etc)

yarikoptic:
Re 2. -- sorry I don't get it since I think dcm2niix doesn't add prefixes but suffixes, and does it regardless either you provide it with just relative path or full path or directory for output:
```
(git-annex)hopa:…mouth-phantoms/bids_test6-PD+T2w[master]005-anat-PD+T2w_acq-tse-tra-3mm
$> dcm2niix -b y -f something . 
Chris Rorden's dcm2niiX version v1.0.20190410  (JP2:OpenJPEG) GCC8.3.0 (64-bit Linux)
Found 70 DICOM file(s)
Warning: interpolated protocol 'anat-PD+T2w_acq-tse-tra-3mm' may be unsuitable for dwidenoise/mrdegibbs. ./000001.dcm
slices not stacked: echo varies (TE 9.4, 94; echo 1, 2). Use 'merge 2D slices' option to force stacking
Convert 35 DICOM as ./something_e1 (640x640x35x1)
Compress: "/usr/bin/pigz" -b 960 -n -f -6 "./something_e1.nii"
Convert 35 DICOM as ./something_e2 (640x640x35x1)
Compress: "/usr/bin/pigz" -b 960 -n -f -6 "./something_e2.nii"
Conversion required 2.147344 seconds (0.196041 for core code).
dcm2niix -b y -f something .  5.91s user 0.17s system 282% cpu 2.153 total
1 11283.....................................:Thu 23 May 2019 11:25:50 AM EDT:.
(git-annex)hopa:…mouth-phantoms/bids_test6-PD+T2w[master]005-anat-PD+T2w_acq-tse-tra-3mm
$> dcm2niix -b y -f /tmp/out/something . 
Chris Rorden's dcm2niiX version v1.0.20190410  (JP2:OpenJPEG) GCC8.3.0 (64-bit Linux)
Found 70 DICOM file(s)
Warning: interpolated protocol 'anat-PD+T2w_acq-tse-tra-3mm' may be unsuitable for dwidenoise/mrdegibbs. ./000001.dcm
slices not stacked: echo varies (TE 9.4, 94; echo 1, 2). Use 'merge 2D slices' option to force stacking
Convert 35 DICOM as ./tmp/out/something_e1 (640x640x35x1)
Compress: "/usr/bin/pigz" -b 960 -n -f -6 "./tmp/out/something_e1.nii"
Convert 35 DICOM as ./tmp/out/something_e2 (640x640x35x1)
Compress: "/usr/bin/pigz" -b 960 -n -f -6 "./tmp/out/something_e2.nii"
Conversion required 1.765559 seconds (0.144816 for core code).
1 11284.....................................:Thu 23 May 2019 11:27:10 AM EDT:.
(git-annex)hopa:…mouth-phantoms/bids_test6-PD+T2w[master]005-anat-PD+T2w_acq-tse-tra-3mm
$> dcm2niix -b y -f /tmp/out/ .         
Chris Rorden's dcm2niiX version v1.0.20190410  (JP2:OpenJPEG) GCC8.3.0 (64-bit Linux)
Found 70 DICOM file(s)
Warning: interpolated protocol 'anat-PD+T2w_acq-tse-tra-3mm' may be unsuitable for dwidenoise/mrdegibbs. ./000001.dcm
slices not stacked: echo varies (TE 9.4, 94; echo 1, 2). Use 'merge 2D slices' option to force stacking
Convert 35 DICOM as ./tmp/out/_e1 (640x640x35x1)
Compress: "/usr/bin/pigz" -b 960 -n -f -6 "./tmp/out/_e1.nii"
Convert 35 DICOM as ./tmp/out/_e2 (640x640x35x1)
Compress: "/usr/bin/pigz" -b 960 -n -f -6 "./tmp/out/_e2.nii"
Conversion required 1.748011 seconds (0.145586 for core code).
dcm2niix -b y -f /tmp/out/ .  5.88s user 0.11s system 341% cpu 1.752 total
1 11285.....................................:Thu 23 May 2019 11:27:40 AM EDT:.
(git-annex)hopa:…mouth-phantoms/bids_test6-PD+T2w[master]005-anat-PD+T2w_acq-tse-tra-3mm
$> dcm2niix --version
Chris Rorden's dcm2niiX version v1.0.20190410  (JP2:OpenJPEG) GCC8.3.0 (64-bit Linux)
 Error: invalid option '--version (null)'
Example output filename: '/func.nii.gz'
```
what am I missing?

mgxd:
erm by "prefix" I meant suffix, and this change should decrease the amount of collisions. Is there any advantage to having `-f anat` instead of `-f ./full/path/to/output`?

And I think the added clarity that 1) gives is enough to make the change.

yarikoptic:
Having intermediate results stored in a temporary place, not in the final location/naming helps to 
- avoid junk from crashed runs interfering (we would crash unless `--overwrite` is given)
- cleanup by removing stale `tmp/` directories

> this change should decrease the amount of collisions.

how so? I thought that `func` is getting generated in the temporary directory.  So collisions of what?

yarikoptic:
as for `1.` -- which is imho indeed valuable, I wonder if we could take "git-annex" approach to urls (when addurl) and some other tools -- make a name out of the path, e.g. `prefix.lstrip(os.sep).replace(os.sep, '_')`, may be ideally first stripping away the dataset topdir from the prefix first.  Then they can see what is being produced and get an idea later on from looking at left over tmp directories what has not completed correctly

mgxd:
>Having intermediate results stored in a temporary place, not in the final location/naming

they still would be generated in a temp dir (we're still setting the nipype node's `base_dir`)

It would just generate a path that emulates the final path, i.e. 

```
# invocation
heudiconv ... -o /code/heudiconv/localconv/multiecho

# current file generated after convertnode.run
/tmp/dcm2niix_5p2gey2/convert/func_e1.nii.gz

# proposed file generated after convertnode.run
/tmp/dcm2niixqk4u57yr/convert/code/heudiconv/localconv/multiecho/sub-ME/func/sub-ME_task-test_run-1_bold_e1.nii.gz
```

yarikoptic:
thanks for the explanation @mgxd  -- makes full sense! I was misguided by the full path as prefix I guess thinking it would be the full output path. 
I also dug through the history to make sure that original use of basename on dir there to see that may be there was some specific reason noted... but it seems it was not the case in 527f9f3f  so this should only make it better ;)  Thanks!

PR #354, opened 2019-06-18
BF: avoid cached heuristic discrepancies

### LINKED ISSUE #353: Cached heuristic name discrepancy
### Summary
When reconverting and using a different heuristic filename from the cache, the conversion table (`filegroup.json`) is still reused. This results in `heudiconv` ignoring the new heuristic and defaulting to the grouping from the previous heuristic. Relevant code excerpt:

https://github.com/nipy/heudiconv/blob/312cd2e5db4968b1d71920aa681ff2795efc3fab/heudiconv/convert.py#L125-L133
...and saving...
https://github.com/nipy/heudiconv/blob/312cd2e5db4968b1d71920aa681ff2795efc3fab/heudiconv/convert.py#L154

**Proposed solution** - save used heuristic in cache simply as `heuristic.py` to avoid any naming discrepancies.

mgxd:
Closes #353

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -123,12 +123,20 @@ def prep_conversion(sid, dicoms, outdir, heuristic, converter, anon_sid,
     # ref: https://github.com/nipy/heudiconv/issues/84#issuecomment-330048609
     # for more automagical wishes
     target_heuristic_filename = op.join(idir, 'heuristic.py')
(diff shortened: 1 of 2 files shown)


mgxd:
build isn't displaying, but this should be good to go - https://travis-ci.org/nipy/heudiconv/builds/555892061

PR #358, opened 2019-07-03
BF: ReproIn - Support pipolar fieldmaps by providing them with _epi not _magnitude

### LINKED ISSUE #357: Support pepolar fmaps
Not yet sure if might be reproin heuristic specific. ATM for SE sequences intended to be pepolar fieldmaps https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#case-4-multiple-phase-encoded-directions-pepolar we are getting files with `_magnitude` suffix, whenever it should be `_epi`.

Attn @amanelis17

yarikoptic:
There is also internal small ReproIn RF to minimize code duplication for common
functionality.

This change might result in slightly different naming due to now enforced order
of _dir- to come before the leftover BIDS suffixes.  So if previously there were
some other BIDS _key-value pairs which we did not treat in ReproIn, and they
were listed BEFORE _dir- - they might now be added AFTER it.  Some subsequent
PR should probably add treatment for other BIDS _key's (e.g. _ce) and place them
in the order listed within BIDS specification

Closes #357 

Unfortunately I believe we do not have any phantom SE scan with _dir- to provide some basic integration test for this fix ATM.

### DIFF
--- a/heudiconv/heuristics/reproin.py
+++ b/heudiconv/heuristics/reproin.py
@@ -551,7 +551,10 @@ def infotodict(seqinfo):
             if not dcm_image_iod_spec:
                 raise ValueError("Do not know image data type yet to make decision")
             seqtype_label = {
-                'M': 'magnitude',  # might want explicit {file_index}  ?
+                # might want explicit {file_index}  ?
+                # _epi for pipolar fieldmaps, see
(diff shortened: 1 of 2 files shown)


yarikoptic:
Let's consider it ok - fixes could come later

PR #360, opened 2019-07-17
Fixed possible TypeError on sorting None object

darrencl:
Fixing possible error on `convert.py`  on sorting `None`. Workaround is to return `[None] * len(res_files)` instead of `sorted([None] * len(res_files))` upon `else` condition.


yarikoptic:
Thanks, LGTM

mgxd:
thanks!

PR #364, opened 2019-08-07
BF:Change nipype required version to >=1.0.0

hbraunDSP:
HeuDiConv requires the function nipype.utils.filemanip.which(), which first appears in nipype release 1.0.0

### DIFF
--- a/heudiconv/info.py
+++ b/heudiconv/info.py
@@ -11,7 +11,7 @@ tarballs into collections of NIfTI files following pre-defined heuristic(s)."""
 REQUIRES = [
     'nibabel',
     'pydicom',
-    'nipype>=0.12.0',
+    'nipype>=1.0.0',
     'pathlib',
     'dcmstack>=0.7',
 ]


mgxd:
Thank you for the fix @hbraunDSP!
Looks like we're also running into nipy/nipype#2980, so we'll have to patch that in as well.

@yarikoptic - are you okay with dropping support for 0.x nipype? I am.

heudiconv/info.py
```diff
@@ -11,7 +11,7 @@
 REQUIRES = [
     'nibabel',
     'pydicom',
-    'nipype>=0.12.0',
+    'nipype>=1.0.0',
     'pathlib',
     'dcmstack>=0.7',
```
effigies: This should fix the test. `lxml` did not set a `python_requires` when they broke 3.4 compatibility.

```suggestion
    'dcmstack>=0.7',
    'lxml<4.4.0; python_version == "3.4"',
```

effigies: Oh, looks like @mgxd got here first.

yarikoptic: > @yarikoptic - are you okay with dropping support for 0.x nipype? I am.

yes -- let's look forward!

re lxml -- should we just stop carrying about 3.4?  Even Debian stable has 3.7 nowadays ;)

mgxd: I'm all for removing 3.4 from our testing suite!!!

PR #366, opened 2019-08-17
FIX: Determine multi-echo echo numbers from EchoTimes 

### LINKED ISSUE #347: Error if EchoNumber tag missing with multi-echo sequence
### Summary

While PR #345 successfully treats PD+T2 pairs as multi-echo, we now get runtime errors when such acquisitions are missing the `EchoNumber` tag:
```
*Runtime**Error**: was asked to copy X but destination already exists: Y
```
Using the `--overwrite` option eliminates the error but only leaves us with one echo.

Presently, multi echo support depends on `EchoNumber` tag, but given there are DICOM files in the wild that lack this tag we need a backup logic, presumably based on `EchoTime`.  I can see that this requires a modification to the logic of `save_converted_files` in [convert.py](https://github.com/nipy/heudiconv/blob/master/heudiconv/convert.py)... something like 

```
EchoNumberBackup = [x+1 for x in sorted(range(len(echo_times)), key=echo_times.__getitem__)]
```

and then [assignment of `echo_number`](https://github.com/nipy/heudiconv/blob/master/heudiconv/convert.py#L572) will need to check for empty value and use `EchoNumberBackup` when missing.  I can almost see what to do but it's just a bit beyond my very elementary Python.

### Platform details:

Choose one:
- [x] Local environment

Python 3.6.6

- Heudiconv version: 

v0.5.4 + PR #345 patch

tsalo:
Closes #347. If the EchoNumber tag is missing from the sequence metadata, this will use the EchoTimes from the associated multi-echo files to determine echo number.

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -518,11 +518,11 @@ def save_converted_files(res, item_dicoms, bids_options, outtype, prefix, outnam
         #   echo times for all bids_files and see if they are all the same or not.
 
         # Check for varying echo times
-        echo_times = set(
+        echo_times = sorted(list(set(
             load_json(b).get('EchoTime', None)
             for b in bids_files
             if b
-        )
+        )))
 
         is_multiecho = len(echo_times) > 1
 
@@ -574,8 +574,11 @@ def save_converted_files(res, item_dicoms, bids_options, outtype, prefix, outnam
             # For multi-echo sequences, we have to specify the echo number in
             # the file name:
             if bids_file and is_multiecho:
-                # Get the EchoNumber from json file info.  If not present, it's echo-1
-                echo_number = fileinfo.get('EchoNumber', 1)
+                # Get the EchoNumber from json file info.  If not present, use EchoTime
+                if 'EchoNumber' in fileinfo.keys():
+                    echo_number = fileinfo['EchoNumber']
+                else:
+                    echo_number = echo_times.index(fileinfo['EchoTime']) + 1
 
 
                 supported_multiecho = ['_bold', '_epi', '_sbref', '_T1w', '_PDT2']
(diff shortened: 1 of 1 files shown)


heudiconv/convert.py
```diff
@@ -574,9 +574,11 @@ def save_converted_files(res, item_dicoms, bids_options, outtype, prefix, outnam
             # For multi-echo sequences, we have to specify the echo number in
             # the file name:
             if bids_file and is_multiecho:
-                # Get the EchoNumber from json file info.  If not present, it's echo-1
-                echo_number = fileinfo.get('EchoNumber', 1)
-
+                # Get the EchoNumber from json file info.  If not present, use EchoTime
+                if 'EchoNumber' in fileinfo.keys():
+                    echo_number = fileinfo['EchoNumber']
+                else:
```
mgxd: We should ensure echotime is available
```suggestion
                elif fileinfo.get('EchoTime'):
```
and just set `echo_number` to None if neither are defined.

tsalo: I was thinking that, since this line should only be reached if the data are multi-echo, it should raise an error if EchoTime is unavailable. What was happening before (in #347) was that the real cause of the error (missing EchoNumber field) was being obscured because the value was defaulting to 1. The error I was getting there was from copying files, which was hard to interpret.

That said, maybe defaulting to `None` will make the resulting error clearer, while defaulting to `1` made it confusing.

mgxd: Either way is fine to me - let's keep it as you have it. Could you merge with master - the CI problem should be fixed now

tsalo:
I'm not sure what's causing the Py2.7 tests to fail, but it looks like it's happening for ReproIn tests specifically.

mgxd:
Ugh.. it looks like there were some side effects to the mass refactors included in the latest nipype release (1.2.1). I'll look into this - your fixes look good.

PR #369, opened 2019-09-11
enh: add telemetry support

satra:


heudiconv/cli/run.py
```diff
@@ -99,6 +99,16 @@ def process_extra_commands(outdir, args):
...
+    import etelemetry
+    try:
+        latest = etelemetry.get_project("nipy/heudiconv")
+    except RuntimeError as e:
+        print("Could not check for version updates: ", e)
+    else:
+        if latest and 'version' in latest:
+            print("Your version: {0} Latest version: {1}".format(__version__,
+                                                                 latest["version"]))
+
```
yarikoptic: not sure if this would be the right place to inform. I would like `heudiconv --version` output remain containing only the current version of the heudiconv

heudiconv/cli/run.py
```diff
@@ -99,6 +99,16 @@ def process_extra_commands(outdir, args):
 
 
 def main(argv=None):
+    import etelemetry
+    try:
+        latest = etelemetry.get_project("nipy/heudiconv")
+    except RuntimeError as e:
+        print("Could not check for version updates: ", e)
+    else:
+        if latest and 'version' in latest:
```
yarikoptic: shouldn't it compare `__version__` to the `latest` somehow?

heudiconv/cli/run.py
```diff
@@ -99,6 +99,16 @@ def process_extra_commands(outdir, args):
 
 
 def main(argv=None):
+    import etelemetry
+    try:
+        latest = etelemetry.get_project("nipy/heudiconv")
+    except RuntimeError as e:
+        print("Could not check for version updates: ", e)
```
yarikoptic: is there a >90% guarantee that `etelemetry` wouldn't crash for some other (IO/Syntax/whatever) reason? ;-)  I would have caught all other exceptions (eg. `except Exception as e`) and issued a warning but kept going

heudiconv/info.py
```diff
@@ -14,6 +14,7 @@
     'nipype>=1.0.0',
     'pathlib',
     'dcmstack>=0.7',
+    'etelemetry'
```
yarikoptic: please add trailing `,` right away to minimize next diff ;)

heudiconv/cli/run.py
```diff
@@ -99,6 +99,16 @@ def process_extra_commands(outdir, args):
...
+    else:
+        if latest and 'version' in latest:
+            print("Your version: {0} Latest version: {1}".format(__version__,
```
yarikoptic: also I wonder if it should use `lgr.info` (or `.warning`) instead of a pure `print`

satra:
@mgxd and @yarikoptic - could we merge this and cut a release?

yarikoptic:
> @mgxd and @yarikoptic - could we merge this and cut a release?

Let's merge

Have you tried it "in a wild"? (would be nice to have some testing done before releasing)

satra:
a few other packages are sending signals. we need to make an interface to summarize the signals for the developer.

PR #373, opened 2019-09-18
ENH: Support multi-echo EPI phase data

### LINKED ISSUE #368: Support for complex multi-echo fMRI
### Summary

Functional phase data (from magnitude+phase reconstruction of fMRI data) is not currently supported as a multi-echo data type. I know that there are plans for a more robust form of support for multi-volume sequences in #346, but in the meantime I think that `_phase` needs to be added to the [list of supported multi-echo data types](https://github.com/nipy/heudiconv/blob/5cfb6f2835f76b8c9d6f19c45e8be0759535900f/heudiconv/convert.py#L583).

### Platform details:

Choose one:
- [X] Local environment: OSX Mojave, Python 3.6.2

- [ ] Container

- Heudiconv version: latest, as of 5cfb6f2

tsalo:
Closes #368. This solution will likely be overwritten by a larger refactor of how multi-volume data are handled, but in the meantime this should allow heudiconv to convert complex fMRI data.

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -580,7 +580,7 @@ def save_converted_files(res, item_dicoms, bids_options, outtype, prefix, outnam
                 else:
                     echo_number = echo_times.index(fileinfo['EchoTime']) + 1
 
-                supported_multiecho = ['_bold', '_epi', '_sbref', '_T1w', '_PDT2']
+                supported_multiecho = ['_bold', '_phase', '_epi', '_sbref', '_T1w', '_PDT2']
                 # Now, decide where to insert it.
                 # Insert it **before** the following string(s), whichever appears first.
                 for imgtype in supported_multiecho:


yarikoptic:
2.7 tests failure is probably unrelated, filed separately: https://github.com/nipy/heudiconv/issues/374

I will publish your phantom data (thanks again) to datalad as soon (I hope ;)) as I get access to it, so some rudimentary regression testing could be added.  Especially in the light of possible future RFing we do not want to loose this particular functionality, do we? ;)

yarikoptic:
I am uploading actual data files to http://datasets.datalad.org/?dir=/dicoms/salo/MEEPI ATM.  It is indeed quite huge.  I wonder if we could just just a select subset of dicom files for a minimalistic test?

effigies:
Just as a side note, if you're collecting lots of DICOMs over here, they might be useful to stress test `nibabel.nicom`.

yarikoptic:
well, I am collecting some. There is more in dcm2niix repos: see dcm_qa* submodules of https://github.com/rordenlab/dcm2niix/ 
Of cause I would welcome (re)use of any of those for testing!  Let me know if anything I could do to facilitate that even more ;)

tsalo:
@yarikoptic 612_Estimation_task1_6ET_MB4_SBRef could be removed (although it's not big) and volumes from phase/mag run could be removed, as long as the number of volumes was kept the same between 613_Estimation_task1_6ET_MB4 and 614_Estimation_task1_6ET_MB4. Would you like me to cut them down to like 50 volumes each?

effigies:
@yarikoptic I think the main thing would be to try to go through them systematically and see if nibabel interprets them correctly, which is going to require either good documentation of each or an experienced eye. For the ones that don't, we can pull them down by datalad or any other method for an extended test, and start fixing them.

But I can stop spamming your thread. I'll open an issue on nibabel.

yarikoptic:
@tsalo we could remove or just selectively get ones for testing and specify to heudiconv to process

tsalo:
@yarikoptic Sorry for the delay. [test_multiecho](https://github.com/nipy/heudiconv/blob/665d901ea4ba1ce69df7aa580b4cd04a1edaac49/heudiconv/tests/test_regression.py#L56) seems like a good base for me to adapt. I have no clue how to select only a subset of dicoms for conversion though.

yarikoptic:
I guess I will just get this one merged as is, without testing for now... hands are too full, and progress must not stop! ;)

PR #379, opened 2019-10-03
ENH: identify invalid json

### LINKED ISSUE #378: enhance explicitness about what json files heudiconv fails to read
### Summary
I was helping a user troubleshoot a problem with heudiconv, as it was tripping up on an invalid json file, but the user only thought heudiconv was touching files in the subject directory inside the bids directory, and not looking in other places inside the bids directory (which it does).

A clue that could have gotten this user on track faster would be to print which file was the invalid json file.

To help, the `load_json` function could be changed to something like:

```
try:
    with open(filename, 'r') as fp:
            data = json.load(fp)
except json.JSONDecodeError:
    print("{fname} is not a valid json file".format(fname=filename)
    raise
```
Thoughts on this?
If positive I can open a pull request.

### Platform details:

Choose one:
- [ ] Local environment

- [x] Container

- Heudiconv version: 0.5.4

jdkent:
fixes #378 

This change prints out what json file was invalid.

### DIFF
--- a/heudiconv/tests/test_utils.py
+++ b/heudiconv/tests/test_utils.py
@@ -1,10 +1,12 @@
+import json
 import os
 import os.path as op
 from heudiconv.utils import (
     get_known_heuristics_with_descriptions,
     get_heuristic_description,
     load_heuristic,
-    json_dumps_pretty)
+    json_dumps_pretty,
+    load_json)
 
 import pytest
 from .utils import HEURISTICS_PATH
@@ -59,3 +61,16 @@ def test_json_dumps_pretty():
     # just the date which reveals the issue
     # tstr = 'Mar  3 2017 10:46:13 by eja'
     assert pretty({'WipMemBlock': tstr}) == '{\n  "WipMemBlock": "%s"\n}' % tstr
+
+
+def test_load_json(tmp_path, capsys):
+    # test invalid json
+    content = "I'm Jason Bourne"
+    fname = "invalid.json"
+    json_file = tmp_path / fname
+    json_file.write_text(content)
+
+    with pytest.raises(json.JSONDecodeError):
+        load_json(json_file)
+        captured = capsys.readouterr()
+        assert fname in captured.out
--- a/heudiconv/tests/test_utils.py
+++ b/heudiconv/tests/test_utils.py
@@ -2,10 +2,7 @@ import json
 import os
 import os.path as op
 
-try:
-    from json.decoder import JSONDecodeError
-except ImportError:
-    JSONDecodeError = ValueError
+from ..utils import JSONDecodeError, create_tree, save_json
 
 from heudiconv.utils import (
     get_known_heuristics_with_descriptions,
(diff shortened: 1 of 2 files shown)


heudiconv/tests/test_utils.py
```diff
@@ -1,10 +1,18 @@
+import json
 import os
 import os.path as op
+
+try:
+    from json.decoder import JSONDecodeError
+except ImportError:
+    JSONDecodeError = ValueError
```
yarikoptic: so this is to work on PY2.  I would have made it explicit.
We don't use/require `six` at runtime so doomed to use `sys.version_info[0] > 2` or define PY2, PY3 e.g. within `heudiconv.utils`.
That would help later on to find/prune python2-specific code.

heudiconv/tests/test_utils.py
```diff
@@ -58,4 +66,27 @@ def test_json_dumps_pretty():
...
\ No newline at end of file
+    assert pretty({'WipMemBlock': tstr}) == '{\n  "WipMemBlock": "%s"\n}' % tstr
+
+
+def test_load_json(tmp_path, capsys):
+    # test invalid json
+    icontent = u"I'm Jason Bourne"
+    ifname = "invalid.json"
+    invalid_json_file = tmp_path / ifname
+    invalid_json_file.write_text(icontent)
```
yarikoptic: FWIW, we have `create_tree` helper, so all above could be expressed as
```python
create_tree(str(tmp_path), {'invalid.json': u"I'm Jason Bourne"})
```
I believe

heudiconv/tests/test_utils.py
```diff
@@ -58,4 +66,27 @@ def test_json_dumps_pretty():
...
+
+    with pytest.raises(JSONDecodeError):
+        load_json(str(invalid_json_file))
+        captured = capsys.readouterr()
```
yarikoptic: I believe all the checks should be brought out of the context manager, or otherwise would not actually be executed since above `load_json` would raise an exception (to be caught by `pytest.raises`)

heudiconv/tests/test_utils.py
```diff
@@ -58,4 +66,27 @@ def test_json_dumps_pretty():
...
+        assert ifname in captured.out
+
+    # test valid json
+    vcontent = {"secret": "spy"}
+    vfname = "valid.json"
+    valid_json_file = tmp_path / vfname
+
+    with open(str(valid_json_file), "w") as vj:
+        json.dump(vcontent, vj)
```
yarikoptic: `create_tree` here as well

heudiconv/utils.py
```diff
@@ -14,6 +14,11 @@
 from glob import glob
 from subprocess import check_output
 
+try:
+    from json.decoder import JSONDecodeError
+except ImportError:
+    JSONDecodeError = ValueError
```
yarikoptic: explicit PY2 check here please (see above comment)

heudiconv/tests/test_utils.py
yarikoptic: actually -- you do the same in `heudiconv.utils`, so do it there and just import here the exception

heudiconv/utils.py
```diff
@@ -172,10 +177,15 @@ def load_json(filename):
...
     data : dict
     """
-    with open(filename, 'r') as fp:
-        data = json.load(fp)
-    return data
+    try:
+        with open(filename, 'r') as fp:
+            data = json.load(fp)
+    except JSONDecodeError:
+        print("{fname} is not a valid json file".format(fname=filename))
```
yarikoptic: use `lgr.error` not pure print

heudiconv/tests/test_utils.py
jdkent: would `create_tree` also be appropriate if the contents are a dictionary, not a string? (I used `save_json` instead)

jdkent:
Thank you for the helpful review! I believe I've covered your suggestions, with the deviation of using [save_json instead of create_tree](https://github.com/nipy/heudiconv/pull/379#discussion_r331198949)

yarikoptic:
cool, thanks!

PR #380, opened 2019-10-03
MNT: Add Python support metadata to package

effigies:
Got the currently supported versions from `.travis.yml`. This will advertise better and keep `pip > 9.0.0` from installing on unsupported Python installations.

Also added in some basic classifiers.

Inspired by going to comment on a PR and not knowing what language features I could assume.


yarikoptic:
cool, thanks!

effigies:
Just a note that you might want to sync up your long description with README, as the [PyPI entry](https://pypi.org/project/heudiconv/) is pretty sparse. You can also add `long_description = file:README.rst` to your `setup.cfg` (see [this gist](https://gist.github.com/effigies/c9f4194034ee218bab1668bfd7851cfc) for more), if you don't want to sync.

Issue #393, opened 2019-11-27
slurm tests error out on which call


yarikoptic:
```
============================================================================================================= ERRORS ==============================================================================================================
____________________________________________________ ERROR at setup of test_queue_no_slurm[--files /home/yoh/picts/mris/heudiconv-master/heudiconv/tests/data/01-fmap_acq-3mm] ____________________________________________________
Error evaluating 'skipif' expression
    /usr/bin/sbatch
     ^
SyntaxError: invalid syntax
_______________________________________________ ERROR at setup of test_queue_no_slurm[-d /home/yoh/picts/mris/heudiconv-master/heudiconv/tests/data/{subject}/* -s 01-fmap_acq-3mm] _______________________________________________
Error evaluating 'skipif' expression
    /usr/bin/sbatch
     ^
SyntaxError: invalid syntax
```
Environment was produced via
```
$> deactivate ; rm -rf venvs/dev3 && virtualenv --system-site-packages --python=python3 venvs/dev3 && source venvs/dev3/bin/activate && pip install -e '.[all]'
zsh: command not found: deactivate
Already using interpreter /usr/bin/python3
Using base prefix '/usr'
New python executable in /home/yoh/picts/mris/heudiconv-0.5.x/venvs/dev3/bin/python3
Also creating executable in /home/yoh/picts/mris/heudiconv-0.5.x/venvs/dev3/bin/python
Installing setuptools, pkg_resources, pip, wheel...done.
Obtaining file:///home/yoh/picts/mris/heudiconv-0.5.x
Requirement already satisfied: nibabel in /usr/lib/python3/dist-packages (from heudiconv==0.5.4) (2.5.1)
Requirement already satisfied: pydicom in /usr/lib/python3/dist-packages (from heudiconv==0.5.4) (1.2.1)
Collecting nipype>=0.12.0
  Using cached https://files.pythonhosted.org/packages/09/3e/5d28e2d049458afe7cc2401a38e30ea01fc0f9bea6494456524d65a02d65/nipype-1.3.1-py2.py3-none-any.whl
Processing /home/yoh/.cache/pip/wheels/f9/b2/4a/68efdfe5093638a9918bd1bb734af625526e849487200aa171/pathlib-1.0.1-cp37-none-any.whl
Processing /home/yoh/.cache/pip/wheels/74/f3/63/abd014d755b6dcacb074366b2494e488da8200d5b43ff37617/dcmstack-0.7.0-cp37-none-any.whl
[...]

PR #394, opened 2019-12-01
BF: fix check for the sbatch (SLURM) being available

yarikoptic:
Partial fix for #393. A proper fix should check if SLURM is actually
usable ATM


yarikoptic:
D`oh, test was intended to be ran when no slurm available, I got it wrong, will redo

PR #407, opened 2019-12-24
BF: fix logging string + move etelemetry under try/except

### LINKED ISSUE #405: something (etelemetry?) causes logger to puke a traceback upon startup
```shell
bids@rolando:~$ heudiconv -o /tmp/cmrr -l '' --files .... -f cmrr-test.py 
  File "/usr/lib/python2.7/logging/__init__.py", line 868, in emit
    msg = self.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 741, in format
    return fmt.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 465, in format
    record.message = record.getMessage()
  File "/usr/lib/python2.7/logging/__init__.py", line 329, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Logged from file run.py, line 255
INFO: Running heudiconv version 0.6.0 latest Unknown
...
```

yarikoptic:
Eventually some generalization of https://github.com/ReproNim/reproman/pull/493
should become part of etelemetry, and thus avoiding redoing it here. Thus fix is
minimal

Closes #405

### DIFF
--- a/heudiconv/cli/run.py
+++ b/heudiconv/cli/run.py
@@ -248,11 +248,11 @@ def process_args(args):
 
     outdir = op.abspath(args.outdir)
 
-    import etelemetry
     try:
+        import etelemetry
         latest = etelemetry.get_project("nipy/heudiconv")
     except Exception as e:
-        lgr.warning("Could not check for version updates: ", e)
+        lgr.warning("Could not check for version updates: %s", str(e))
         latest = {"version": 'Unknown'}
 
     lgr.info(INIT_MSG(packname=__packagename__,
--- a/heudiconv/tests/test_main.py
+++ b/heudiconv/tests/test_main.py
@@ -1,6 +1,9 @@
 # TODO: break this up by modules
 
-from heudiconv.cli.run import main as runner
+from heudiconv.cli.run import (
+    main as runner,
+    process_args,
+)
 from heudiconv import __version__
 from heudiconv.utils import (create_file_if_missing,
                              set_readonly,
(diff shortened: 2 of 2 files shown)


heudiconv/cli/run.py
```diff
@@ -248,11 +248,11 @@ def process_args(args):
 
     outdir = op.abspath(args.outdir)
 
-    import etelemetry
     try:
+        import etelemetry
         latest = etelemetry.get_project("nipy/heudiconv")
     except Exception as e:
-        lgr.warning("Could not check for version updates: ", e)
+        lgr.warning("Could not check for version updates: %s", str(e))
```
satra:
```suggestion
        lgr.warning("Could not check for version updates: %s", str(e))
        lgr.warning(("Could not check for version updates: ", e)) 
```

heudiconv/cli/run.py
```diff
@@ -248,11 +248,11 @@ def process_args(args):
 
     outdir = op.abspath(args.outdir)
 
-    import etelemetry
     try:
+        import etelemetry
```
satra: this can be outside, unless etelemetry is not a dependency.

yarikoptic: It could be. But if it fails to import for some reason - whole process crashes. on Debian we don't have it yet, I have to patch - this would about needing to patch. So IMHO a little cost to ensure absent side effects and ease packaging where it is not available yet

heudiconv/cli/run.py
yarikoptic: this would log it as a tuple.  Also docstring says:

```
In [4]: log.warning?
Signature: log.warning(msg, *args, **kwargs)
Docstring:
Log 'msg % args' with severity 'WARNING'.

To pass exception information, use the keyword argument exc_info with
a true value, e.g.

logger.warning("Houston, we have a %s", "bit of a problem", exc_info=1)
```
so it expects a string as the first argument, not a tuple.  It actually works if I pass a tuple, but e.g. that would have caused datalad, which provides custom format which expects str, to puke.  So I would keep it as it was

yarikoptic:
kept running into this, will merge as is, further enhancements go go as PRs

PR #415, opened 2020-02-17
BF: use datalad 0.12, drop PY2, include workaround for dcmstack installation

yarikoptic:
0.12 series introduced a good number of API changes. To avoid per-version special handling, I am just boosting minimal requirement to be datalad 0.12.2.

I also dropped bothering with annex.thin setting since IMHO it did nothing really in a typical use case.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 .[all]
+git+git://github.com/moloney/dcmstack@master
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 # vim ft=yaml
 language: python
 python:
-  - 2.7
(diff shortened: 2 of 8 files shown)


requirements.txt
```diff
@@ -1 +1,2 @@
 .[all]
+git+git://github.com/moloney/dcmstack@master
```
mgxd:
```suggestion
git+git://github.com/moloney/dcmstack@fa24702#egg-dcmstack
```

mgxd: You may also have to install this first - looks like its getting overwritten on install
https://travis-ci.org/nipy/heudiconv/jobs/652087701#L718

PR #416, opened 2020-02-18
MAINT: Drop Python 2 support, update dcmstack + datalad

mgxd:
Replaces #415


mgxd:
let's get this in so we're green again

PR #418, opened 2020-02-19
Fixed the call to embed_metadata_from_dicoms

### LINKED ISSUE #417: `'TaskName'` not set in json files for multi-echo data
### Summary
When extracting multi-echo EPI data, `bids/validator` gives the following error:
```
1: [ERR] You have to define 'TaskName' for this file. (code: 50 - TASK_NAME_MUST_DEFINE)
  ./sub-SMSMEEPI/func/sub-SMSMEEPI_task-TASK_acq-normal_run-01_echo-1_phase.nii.gz
  ./sub-SMSMEEPI/func/sub-SMSMEEPI_task-TASK_acq-normal_run-01_echo-2_phase.nii.gz
  [...]
```
I found that the `'TaskName'` field is added to the json files in `embed_nifti`, which is called from `embed_metadata_from_dicoms`.  The problem is that, for multi-echo data, the call to `embed_metadata_from_dicoms` 
https://github.com/nipy/heudiconv/blob/756fd80bdbc696505189cbb5c33002ee6caae2d8/heudiconv/convert.py#L331
doesn't get executed because `len(bids_outfiles) > 1`.

I'm working on a fix.

### Platform details:

Choose one:
- [ ] Local environment

- [X] Container: nipy/heudiconv:master

- Heudiconv version: 0.6.0

pvelasco:
So that it works for multi-echo series.

Fixes #417

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -321,16 +321,21 @@ def convert(items, converter, scaninfo_suffix, custom_callable, with_prov,
                         % (outname)
                     )
 
-        if len(bids_outfiles) > 1:
-            lgr.warning("For now not embedding BIDS and info generated "
-                        ".nii.gz itself since sequence produced "
-                        "multiple files")
-        elif not bids_outfiles:
-            lgr.debug("No BIDS files were produced, nothing to embed to then")
-        elif outname:
-            embed_metadata_from_dicoms(bids_options, item_dicoms, outname, outname_bids,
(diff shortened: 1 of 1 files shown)


yarikoptic:
I really doubt that it would be a correct thing to do in many cases, e.g. multimodal sequences (T1w+PD) etc.  Even in case of multi-echo, should be done carefully, assuring correct EchoTime for each particular file

pvelasco:
OK. I'll try something simpler.

PR #413, opened 2020-02-14
BF (reproin, PY3): cast seqinfos into a list so entries could be modified if necessary

yarikoptic:
this fix is specific to python 3.x , see comment in the change

PR #419, opened 2020-02-20
BF: case seqinfos into a list so entries could be modified if necessary

yarikoptic:
this fix is specific to python 3.x , see comment in the change

This is a resurrected #413


yarikoptic:
coverage is effected due to my paranoid check.  Shouldn't effect functionality ;)

PR #420, opened 2020-02-20
Set the  field 'TaskName' in the json file for multiecho func data

### LINKED ISSUE #417: `'TaskName'` not set in json files for multi-echo data

pvelasco:
Fixes #417

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -322,9 +322,8 @@ def convert(items, converter, scaninfo_suffix, custom_callable, with_prov,
                     )
 
         if len(bids_outfiles) > 1:
-            lgr.warning("For now not embedding BIDS and info generated "
-                        ".nii.gz itself since sequence produced "
-                        "multiple files")
+            for bids_outfile in bids_outfiles:
+                add_taskname_to_infofile( bids_outfile )
         elif not bids_outfiles:
             lgr.debug("No BIDS files were produced, nothing to embed to then")
         elif outname:
@@ -628,3 +627,29 @@ def save_converted_files(res, item_dicoms, bids_options, outtype, prefix, outnam
             except TypeError as exc:  ##catch lists
                 raise TypeError("Multiple BIDS sidecars detected.")
     return bids_outfiles
+
+
+def  add_taskname_to_infofile(infofile):
+    """Add the "TaskName" field to json files corresponding to func images.
+
+    Parameters
+    ----------
+    infofile : filename of json file
+
+    Returns
+    -------
+    """
+
+    import re
+
+    meta_info = load_json(infofile)
+
+    try:
+        meta_info['TaskName'] = (re.search('(?<=_task-)\w+',
+                                           op.basename(infofile))
+                                 .group(0).split('_')[0])
+    except AttributeError:
+        pass
+
+    # write to outfile
+    save_json(infofile, meta_info)
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -4,6 +4,7 @@ import os.path as op
 import logging
 import shutil
 import sys
+import re
 
 from .utils import (
     read_config,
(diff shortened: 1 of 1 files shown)


heudiconv/convert.py
```diff
@@ -628,3 +627,29 @@ def save_converted_files(res, item_dicoms, bids_options, outtype, prefix, outnam
...
+    import re
+
+    meta_info = load_json(infofile)
+
+    try:
+        meta_info['TaskName'] = (re.search('(?<=_task-)\w+',
+                                           op.basename(infofile))
+                                 .group(0).split('_')[0])
+    except AttributeError:
+        pass
```
yarikoptic: why to pass and not just return while may be logging that failed to find any task field?

heudiconv/convert.py
```diff
@@ -628,3 +627,29 @@ def save_converted_files(res, item_dicoms, bids_options, outtype, prefix, outnam
...
+
+    Parameters
+    ----------
+    infofile : filename of json file
+
+    Returns
+    -------
+    """
+
+    import re
```
yarikoptic: no need to delay this import - could go to the top

heudiconv/convert.py
```diff
@@ -322,9 +322,8 @@ def convert(items, converter, scaninfo_suffix, custom_callable, with_prov,
                     )
 
         if len(bids_outfiles) > 1:
-            lgr.warning("For now not embedding BIDS and info generated "
```
yarikoptic: this warning is still in effect (somewhat).  We are not embedding "additional" metadata from DICOMs

heudiconv/convert.py
pvelasco: You are right, it may just return.
I don't know if it should log that it failed to find a task field because a.f.a.i.k. this function will be called for any multi-echo sequence, and not all of them are `func`.

pvelasco:
> * take the code which embeds task label out from `embed_nifti`, and move into a function like you did

Do you want this function in `convert.py`, `dicoms.py` or `utils.py`?

yarikoptic:
> > * take the code which embeds task label out from `embed_nifti`, and move into a function like you did
> 
> Do you want this function in `convert.py`, `dicoms.py` or `utils.py`?

has nothing to do with DICOMs per se, will be used in convert, so let's stick it there for now

pvelasco:
> * remove `min_meta` option from that `embed_nifti` and just do not bother even using it in embed_metadata_from_dicoms if `min_meta` is set

Do we need to call `embed_metadata_from_dicoms` at all if `min_meta` is set?
I mean, in `convert.py`, I can call `add_taskname_to_infofile`, and then skip the `if` statements related `embed_metadata_from_dicoms` if `min_meta` is set...

yarikoptic:
> Do we need to call `embed_metadata_from_dicoms` at all if `min_meta` is set?
> I mean, in `convert.py`, I can call `add_taskname_to_infofile`, and then skip the `if` statements related `embed_metadata_from_dicoms` if `min_meta` is set...

good observation.  That `embed_nifti` still has the same code as now in `add_taskname_to_infofile`, and that is (as far as I see it) the only piece which is used if `min_meta`, so I will strip it away from there (after all it is not embedding any metadata from dicoms ;)), and then indeed we could avoid calling `embed_metadata_from_dicoms` if minmeta.  That `with_prov` handling wouldn't be run, but if we are not embedding (min_meta is set) I think it is ok.  I will push this RFing shortly after making piece of mind (and tests) with it ;)

yarikoptic:
actually -- I will just merge this one, and then follow up with a new PR on top

PR #425, opened 2020-03-04
ReproIn: make specification of protocols2fix more flexible + do not bind fixup structs in func signatures

yarikoptic:
First commit in the series describes problem at large

    For some data we cannot rely on reliable presence of consistent study_description
    across scans.  That results in the necessity to duplicate the same protocols2fix
    records for multiple (not known in advance) study descriptions (actually -- their
    hashes, which makes it even more difficult).  But in the cases with custom
    overloads of protocols2fix, it is desired to just provide global rewrite rules,
    which could be applied to any collection since they did follow some convention.
    
    Now, with an empty "hash", those rules would be applied last, i.e. after possible
    study-specific fixups applied.
    
    As part of the solution I removed preliminary check to skip a call to
    fix_dbic_protocol altogether.  It should be better this way.
    
    I also removed binding of series_spec_fields and protocols2fix within function
    signature.  It should allow easier overloading at the module level when needed

and subsequent commits improved it even further, but altogether removing function signature level bindings, adding matching `study_description` using regular expression (so it could then be crafted per lab etc).

Example log output on the adjusted tests:
```shell
INFO: Considering study (0e4497bdaae2608324a85be1f54f60ed) specific substitutions
INFO:  field1: '02-anat-scout_run+_MPR_sag' -> '02-anat-THESCOUT-runX_MPR_sag'
INFO: Considering '^my.*' regex matching substitutions
INFO:  field1: '02-anat-THESCOUT-runX_MPR_sag' -> '02-anat-THESCOUT_MPR_sag'
INFO: Considering global substitutions
INFO:  field1: '02-anat-THESCOUT_MPR_sag' -> '02-anat-scout_MPR_sag'
INFO: Considering study (0e4497bdaae2608324a85be1f54f60ed) specific substitutions
INFO:  field2: '11-func_run-life2_acq-2mm692' -> '11-func_run+_task-life_acq-2mm692'
INFO: Considering '^my.*' regex matching substitutions
INFO: Considering global substitutions
```

### DIFF
--- a/heudiconv/heuristics/reproin.py
+++ b/heudiconv/heuristics/reproin.py
@@ -391,26 +391,38 @@ def fix_canceled_runs(seqinfo, accession2run=fix_accession2run):
     return seqinfo
 
 
-def fix_dbic_protocol(seqinfo, keys=series_spec_fields, subsdict=protocols2fix):
+def fix_dbic_protocol(seqinfo, keys=None, subsdict=None):
     """Ad-hoc fixup for existing protocols
     """
+    if subsdict is None:
+        subsdict = protocols2fix
+    if keys is None:
+        keys = series_spec_fields
+
     study_hash = get_study_hash(seqinfo)
 
-    if study_hash not in subsdict:
-        raise ValueError("I don't know how to fix {0}".format(study_hash))
-
-    # need to replace both protocol_name series_description
-    substitutions = subsdict[study_hash]
-    for i, s in enumerate(seqinfo):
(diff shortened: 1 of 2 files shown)


yarikoptic:
eh, 3.5 and 3.6 " AttributeError: module 're' has no attribute 'Pattern'"

yarikoptic:
tests pass, changes shouldn't break anything, will merge in a day or two unless hear otherwise

Issue #427, opened 2020-03-18
error v 0.6.0 philips fieldmaps and -b flag


cni-md:
### Summary

I try to convert a Philips Fieldmap with heudiconv v 0.6.0 The heuristic was working with version 0.5.4
Without the -b flag the conversion is done with version 0.6.0 as well

With -b:
`['/scratch/spralat_dti/tmp/dcm2niix7rw4dlrr/convert/scratch/tmp/SpraLat/dti/sub-do33/fmap/sub-do33_fmap_001_e1.json', '/scratch/spralat_dti/tmp/dcm2niix7rw4dlrr/convert/scratch/tmp/SpraLat/dti/sub-do33/fmap/sub-do33_fmap_001_e1a.json']
[None, 0.0023]
********************************
Traceback (most recent call last):
  File "/users/jstadler/studien/testspralat/bin/heudiconv", line 8, in <module>
    sys.exit(main())
  File "/users/jstadler/studien/testspralat/lib/python3.5/site-packages/heudiconv/cli/run.py", line 128, in main
    process_args(args)
  File "/users/jstadler/studien/testspralat/lib/python3.5/site-packages/heudiconv/cli/run.py", line 346, in process_args
    dcmconfig=args.dcmconfig,)
  File "/users/jstadler/studien/testspralat/lib/python3.5/site-packages/heudiconv/convert.py", line 209, in prep_conversion
    dcmconfig=dcmconfig,)
  File "/users/jstadler/studien/testspralat/lib/python3.5/site-packages/heudiconv/convert.py", line 306, in convert
    overwrite=overwrite)
  File "/users/jstadler/studien/testspralat/lib/python3.5/site-packages/heudiconv/convert.py", line 537, in save_converted_files
    for b in bids_files
TypeError: unorderable types: NoneType() < float()

I inserted in conversion.py line 532
`print(bids_files)`
` print([load_json(b).get('EchoTime', None) for b in bids_files])`
  print('********************************')  `
for debugging

I changed the code (lines 532...)  to:
`# Check for varying echo times`                                                                                                                                                   
`        from numpy import nan`
`        echo_times = sorted(list(set(`
`            load_json(b).get('EchoTime',nan)`
`            for b in bids_files`
`            if b`
`        ))) `

And now it is working, but I do not know the side effects.

### Platform details:

Choose one:
- [X] Local environment

- [ ] Container

on debian scratch and buster

- Heudiconv version: 

0.6.0

[...]

PR #428, opened 2020-03-18
BF: #427 fieldmap sort problem

cni-md:
Sort of None echotime was not working. Replaced with nan


heudiconv/convert.py
```diff
@@ -529,8 +529,9 @@ def save_converted_files(res, item_dicoms, bids_options, outtype, prefix, outnam
         #   echo times for all bids_files and see if they are all the same or not.
 
         # Check for varying echo times
+        from numpy import nan
```
yarikoptic: missed the PR, saw only commit in the issue, folded into #429 but adjusted commit to use math.nan not numpy.nan here

yarikoptic:
included in #429

PR #432, opened 2020-03-27
RF: embed_nifti -> embed_dicom_and_nifti_metadata  and not invoke it if min_meta

yarikoptic:
Follow up to #420 

- `embed_dicom_and_nifti_metadata` will not produce nifti file if that one doesn't exist yet . This functionality (hopefully) was not used for a while, and we should rely on dcm2niix for all such conversions
- it also changes how json's dumped by that function -- it will use the same `save_json` instead of ad-hoc `json.dump` call with `indent=3`.  So with this PR we would end up changing .json files formatting a bit, which could complicate regression testing, but I think consistency is worth it
- `@assert_cwd_unchanged` now can be used to assure that tests do not jump to some other directories, or force their return (note that if directory has symlinks in the path -- we would return to dereferenced path)

### DIFF
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -6,7 +6,13 @@ from collections import OrderedDict
 import tarfile
 
 from .external.pydicom import dcm
-from .utils import load_json, get_typed_attr, set_readonly, SeqInfo
+from .utils import (
+    get_typed_attr,
+    load_json,
+    save_json,
+    SeqInfo,
+    set_readonly,
+)
 
 import warnings
 with warnings.catch_warnings():
@@ -410,6 +416,7 @@ def embed_dicom_and_nifti_metadata(dcmfiles, niftifile, infofile, bids_info):
     import os.path as op
     import json
     import re
+    from heudiconv.utils import save_json
 
     from heudiconv.external.dcmstack import ds
     stack = ds.parse_and_stack(dcmfiles, force=True).values()
@@ -440,8 +447,7 @@ def embed_dicom_and_nifti_metadata(dcmfiles, niftifile, infofile, bids_info):
         meta_info.update(bids_info)
 
     # write to outfile
-    with open(infofile, 'wt') as fp:
-        json.dump(meta_info, fp, indent=3, sort_keys=True)
+    save_json(infofile, meta_info)
 
 
 def embed_metadata_from_dicoms(bids_options, item_dicoms, outname, outname_bids,
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -386,7 +386,7 @@ def compress_dicoms(dicom_list, out_prefix, tempdirs, overwrite):
     return outtar
 
 
-def embed_nifti(dcmfiles, niftifile, infofile, bids_info, min_meta):
+def embed_nifti(dcmfiles, niftifile, infofile, bids_info):
     """
 
     If `niftifile` doesn't exist, it gets created out of the `dcmfiles` stack,
@@ -403,7 +403,6 @@ def embed_nifti(dcmfiles, niftifile, infofile, bids_info, min_meta):
     niftifile
     infofile
     bids_info
-    min_meta
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


yarikoptic:
@satra @mgxd please check if my lobotimization of `embed_nifti` and changing min_meta handling is not introducing some side effect (e.g. may be someone used embed_nifti directly?) I am not aware of.

heudiconv/dicoms.py
```diff
@@ -402,68 +404,54 @@ def embed_nifti(dcmfiles, niftifile, infofile, bids_info, min_meta):
...
+    aff = orig_nii.affine
+    ornt = nb.orientations.io_orientation(aff)
+    axcodes = nb.orientations.ornt2axcodes(ornt)
+    new_nii = stack.to_nifti(voxel_order=''.join(axcodes), embed_meta=True)
+    meta_info = ds.NiftiWrapper(new_nii).meta_ext.to_json()
+    meta_info = json.loads(meta_info)
 
     if bids_info:
+        # make nice with python 3 - same behavior?
+        meta_info = meta_info.copy()
```
mgxd: I'm not sure why we make a copy here
```suggestion
```

heudiconv/tests/test_dicoms.py
```diff
@@ -26,35 +30,34 @@ def test_private_csa_header(tmpdir):
...
     dcmfiles = [op.join(TESTS_DATA_PATH, 'axasc35.dcm')]
     infofile = 'infofile.json'
 
-    # 1) nifti does not exist
-    out = embed_nifti(dcmfiles, 'nifti.nii', 'infofile.json', None, False)
-    # string -> json
-    out = json.loads(out)
-    # should have created nifti file
-    assert op.exists('nifti.nii')
+    # 1) nifti does not exist -- no longer supported
```
mgxd: we should test that the `NotImplementedError` is being hit

heudiconv/dicoms.py
```diff
@@ -402,68 +404,54 @@ def embed_nifti(dcmfiles, niftifile, infofile, bids_info, min_meta):
...
-                r'(?<=_task-)\w+', op.basename(infofile)
-            ).group(0).split('_')[0]
-        except AttributeError:
-            pass
     # write to outfile
-    with open(infofile, 'wt') as fp:
-        json.dump(meta_info, fp, indent=3, sort_keys=True)
-
-    return niftifile, infofile
+    save_json(infofile, meta_info)
```
mgxd: should we keep the same indentation here?

heudiconv/dicoms.py
yarikoptic: yep, seems to be no longer necessary -- removing!

heudiconv/dicoms.py
yarikoptic: I was on an edge for this one, and decided that we better do not delay fixing this inconsistency (which we should have done long ago).  Especially now since in case of min_meta, file would be saved with our "standard" indentation (of 4), thus differ from this one.  So, I decided to stay consistent and just use `save_json` with defaults...
but now you made me look at how we use `save_json` and it seems that we prefer `indent=2` in many places:
```
$> git grep -A20 'save_json(' | grep -e 'save_json' -e 'indent'
heudiconv/bids.py:        save_json(descriptor,
heudiconv/bids.py:        save_json(task_file, fields, indent=2, sort_keys=True, pretty=True)
heudiconv/bids.py:        save_json(jsonfile, json_, indent=2)
heudiconv/bids.py:            save_json(json_phasediffname, json_, indent=2)
heudiconv/bids.py:            save_json(participants_json,
heudiconv/bids.py-                indent=2)
heudiconv/bids.py:            save_json(scans_json,
heudiconv/bids.py-                indent=2)
heudiconv/convert.py:        save_json(filegroup_file, filegroup)
heudiconv/convert.py:        save_json(infofile, meta_info)
heudiconv/dicoms.py:    save_json(infofile, meta_info)
heudiconv/tests/test_utils.py:    save_json(valid_json_file, vcontent)
heudiconv/utils.py:def save_json(filename, data, indent=4, sort_keys=True, pretty=False):
heudiconv/utils.py-    indent : int, optional
heudiconv/utils.py-                data, sort_keys=sort_keys, indent=indent)
```

As a separate PR I will propose making default indent to be `2` and thus removing any "ad-hoc" indent use

heudiconv/tests/test_dicoms.py
yarikoptic: easy enough and wortwhile indeed - done in 4fd4391ccbbfab27a241348238af63f3c1a750e6

yarikoptic:
Ok, let's proceed

PR #433, opened 2020-03-29
search heuristic file path first

njaard:
This allows heuristic modules names that conflict with existing module
names, such as using a heuristic module named "heudiconv.py"

### DIFF
--- a/heudiconv/utils.py
+++ b/heudiconv/utils.py
@@ -319,7 +319,7 @@ def load_heuristic(heuristic):
         path, fname = op.split(heuristic_file)
         try:
             old_syspath = sys.path[:]
-            sys.path.append(path)
+            sys.path.append(0, path)
             mod = __import__(fname.split('.')[0])
             mod.filename = heuristic_file
         finally:


heudiconv/utils.py
```diff
@@ -319,7 +319,7 @@ def load_heuristic(heuristic):
         path, fname = op.split(heuristic_file)
         try:
             old_syspath = sys.path[:]
-            sys.path.append(path)
+            sys.path[0:0]=[path]
```
yarikoptic: thanks.  I think (can't come up with a use case when not) it indeed would be better to consider heuristic file path first.
I would though avoid "tricky" `[0:0]` but use explicit `.insert`:
```suggestion
            sys.path[0:0]=[path]
            sys.path.insert(0, path)
```

yarikoptic:
checking -- closed by mistake or discovered why it wouldn't be kosher?

njaard:
I created a new one with only one commit, but it seems GitHub detected it in a way I was not expecting.

PR #434, opened 2020-03-29
search heuristic file path first

njaard:
I modified the code per your recommendation and then rebased the repository so you don't get a noisy commit.

This allows heuristic modules names that conflict with existing module
names, such as using a heuristic module named "heudiconv.py"

### DIFF
--- a/heudiconv/utils.py
+++ b/heudiconv/utils.py
@@ -319,7 +319,7 @@ def load_heuristic(heuristic):
         path, fname = op.split(heuristic_file)
         try:
             old_syspath = sys.path[:]
-            sys.path.append(path)
+            sys.path.append(0, path)
             mod = __import__(fname.split('.')[0])
             mod.filename = heuristic_file
         finally:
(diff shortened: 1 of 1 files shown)


heudiconv/utils.py
```diff
@@ -319,7 +319,7 @@ def load_heuristic(heuristic):
         path, fname = op.split(heuristic_file)
         try:
             old_syspath = sys.path[:]
-            sys.path.append(path)
+            sys.path.append(0, path)
```
yarikoptic: should have been `insert` not `append` -- let's see if I can accept my own suggestion:
```suggestion
            sys.path.insert(0, path)
```

heudiconv/utils.py
```diff
@@ -319,7 +319,7 @@ def load_heuristic(heuristic):
         path, fname = op.split(heuristic_file)
         try:
             old_syspath = sys.path[:]
-            sys.path.append(path)
+            sys.path.insert(0, path)
```
yarikoptic: yeap, I could! Let's see how tests pass now ;)

yarikoptic:
Squashed and merged, thank you @njaard !

PR #436, opened 2020-03-29
RF+ENH: save_json - indent=2 by default

yarikoptic:
- `json_dumps_pretty` had it set to 2
- many invocations of `save_json` also had 2 set manually
- `embed_nifti` has its invocation of json.dumps still set to `3`, but when #432 is merged, it would just use `save_json` with defaults, thus go to 2 

Altogether, although it would change formatting for some files, I think we better make formatting more consistent sooner (i.e. now) than later.

### DIFF
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -171,7 +171,7 @@ def populate_aggregated_jsons(path):
             act = "Generating"
         lgr.debug("%s %s", act, task_file)
         fields.update(placeholders)
-        save_json(task_file, fields, indent=2, sort_keys=True, pretty=True)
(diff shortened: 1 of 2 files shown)


yarikoptic:
just came to my mind the argument "by dcm2niix uses ..." -- apparently it uses TAB for indentation.  I guess we just want to make sure to process all .json we get from dcm2niix (either using min_meta or not)

yarikoptic:
oh well, I think we better make things consistent within a single release (instead of piece meal in multiple), and since we are anyways inconsistent with dcm2niix (or "consistent" with TAB width set to 2 ;)) -- we better proceed by merging.

Issue #173, opened 2018-03-02
current combinatorics of the cmdline specifications lead to suboptimal reporting of incorrect specification


yarikoptic:
in all of those cases I expect us to report usage synopsis with some informative message instead of code tracebacks:

examples:
- [ ] no options specified whatsoever:
```shell
$> heudiconv
INFO: Running heudiconv version 0.4
Traceback (most recent call last):
  File "/usr/bin/heudiconv", line 11, in <module>
    load_entry_point('heudiconv==0.4', 'console_scripts', 'heudiconv')()
  File "/usr/lib/python2.7/dist-packages/heudiconv/cli/run.py", line 120, in main
    process_args(args)
  File "/usr/lib/python2.7/dist-packages/heudiconv/cli/run.py", line 240, in process_args
    heuristic = load_heuristic(args.heuristic)
  File "/usr/lib/python2.7/dist-packages/heudiconv/utils.py", line 274, in load_heuristic
    if os.path.sep in heuristic or os.path.lexists(heuristic):
TypeError: argument of type 'NoneType' is not iterable
```
- [ ] heuristic specified by no subject template or files:
```shell
$> heudiconv -f reproin
INFO: Running heudiconv version 0.4
Traceback (most recent call last):
  File "/usr/bin/heudiconv", line 11, in <module>
    load_entry_point('heudiconv==0.4', 'console_scripts', 'heudiconv')()
  File "/usr/lib/python2.7/dist-packages/heudiconv/cli/run.py", line 120, in main
    process_args(args)
  File "/usr/lib/python2.7/dist-packages/heudiconv/cli/run.py", line 244, in process_args
    args.subjs, grouping=args.grouping)
  File "/usr/lib/python2.7/dist-packages/heudiconv/parser.py", line 155, in get_study_sessions
    for f in files_opt:
TypeError: 'NoneType' object is not iterable
```
- [ ] ls command should work also without heuristic specified:
```shell
$> heudiconv  --command ls --files ~/picts/mris/dartmouth-phantoms/bids_test4-20161014/phantom-1/anat-angio_acq-fl3d                                                                                                         
Traceback (most recent call last):
  File "/usr/bin/heudiconv", line 11, in <module>
    load_entry_point('heudiconv==0.4', 'console_scripts', 'heudiconv')()
  File "/usr/lib/python2.7/dist-packages/heudiconv/cli/run.py", line 120, in main
    process_args(args)
[...]

PR #437, opened 2020-03-30
ENH: Provide informative exception if command needs a heuristic to be specified

yarikoptic:
partially addresses last item in #173 - ideally should work without heuristic, but not ATM, so I will keep that issue open, but at least it should crash more informatively


yarikoptic:
will add a bit more testing

yarikoptic:
nice and green now, merging

PR #439, opened 2020-04-15
DO NOT MERGE: Checks for 0.8.0 release to be commenced

yarikoptic:


yarikoptic:
ok, seems no surprises, so I will merge (locally) and release later today.

PR #442, opened 2020-04-18
[REF] Split main workflow from CLI for Python access

### LINKED ISSUE #440: Make main workflow callable from Python
### Summary

Currently, heudiconv's main workflow is included in the same function that parses the CLI arguments. With a little refactoring, we could separate these two a bit to allow users to call heudiconv from within Python. I think it would be fairly easy, and it would help users like me who prefer to call heudiconv within larger Python scripts.

This is somewhat related to [this NeuroStars post](https://neurostars.org/t/is-there-any-way-to-use-dcm2bids-or-neudiconv-with-python/4949) (specifically point 2).

tsalo:
Closes #440.

This refactor might be more aggressive than the devs might like, so I'm happy to scale it back if necessary.

Changes proposed:
- Make a new workflows module.
- Merge parts of `heudiconv.cli.run.main()` into `heudiconv.cli.run.process_args()` and move to `heudiconv.workflows.run.heudiconv_workflow()`.
    - `heudiconv.cli.run.main()` now does as little as possible (i.e., parses the arguments and does a single check that requires the parser's help message.
- Move the following functions from `heudiconv.cli.run` to `heudiconv.workflows.run`:
    - process_extra_commands
    - setup_exceptionhook
    - is_interactive
    - ensure_heuristic_arg
- Split `args` structure into individual arguments when passed to `heudiconv.workflows.run.heudiconv_workflow()`
    - This makes it easier for users to call the workflow directly in Python, as they won't need to build `args`.
    - It should also make it easier to track updates to the parameters in the parser vs. those used in the workflow.
    - The drawback is that it means updating both the CLI and the workflow function in tandem when changing parameters.
- Reformat the parser a little bit. I can change this back, of course, but I think it's more readable this way.

### DIFF
--- a/heudiconv/cli/run.py
+++ b/heudiconv/cli/run.py
@@ -46,7 +46,8 @@ def setup_exceptionhook():
     sys.excepthook = _pdb_excepthook
 
 
-def process_extra_commands(outdir, args):
+def process_extra_commands(outdir, command, files, dicom_dir_template,
+                           heuristic, session, subjs, grouping):
     """
     Perform custom command instead of regular operations. Supported commands:
     ['treat-json', 'ls', 'populate-templates']
@@ -58,17 +59,17 @@ def process_extra_commands(outdir, args):
     args : Namespace
         arguments
     """
-    if args.command == 'treat-jsons':
-        for f in args.files:
+    if command == 'treat-jsons':
+        for f in files:
             treat_infofile(f)
-    elif args.command == 'ls':
-        ensure_heuristic_arg(args)
-        heuristic = load_heuristic(args.heuristic)
+    elif command == 'ls':
+        ensure_heuristic_arg(heuristic)
+        heuristic = load_heuristic(heuristic)
         heuristic_ls = getattr(heuristic, 'ls', None)
-        for f in args.files:
+        for f in files:
             study_sessions = get_study_sessions(
-                args.dicom_dir_template, [f], heuristic, outdir,
-                args.session, args.subjs, grouping=args.grouping)
+                dicom_dir_template, [f], heuristic, outdir,
+                session, subjs, grouping=grouping)
             print(f)
             for study_session, sequences in study_sessions.items():
                 suf = ''
@@ -78,29 +79,29 @@ def process_extra_commands(outdir, args):
                     "\t%s %d sequences%s"
                     % (str(study_session), len(sequences), suf)
                 )
-    elif args.command == 'populate-templates':
-        ensure_heuristic_arg(args)
-        heuristic = load_heuristic(args.heuristic)
-        for f in args.files:
+    elif command == 'populate-templates':
+        ensure_heuristic_arg(heuristic)
+        heuristic = load_heuristic(heuristic)
+        for f in files:
             populate_bids_templates(f, getattr(heuristic, 'DEFAULT_FIELDS', {}))
-    elif args.command == 'sanitize-jsons':
-        tuneup_bids_json_files(args.files)
-    elif args.command == 'heuristics':
+    elif command == 'sanitize-jsons':
+        tuneup_bids_json_files(files)
+    elif command == 'heuristics':
         from ..utils import get_known_heuristics_with_descriptions
         for name_desc in get_known_heuristics_with_descriptions().items():
             print("- %s: %s" % name_desc)
-    elif args.command == 'heuristic-info':
-        ensure_heuristic_arg(args)
+    elif command == 'heuristic-info':
+        ensure_heuristic_arg(heuristic)
         from ..utils import get_heuristic_description
-        print(get_heuristic_description(args.heuristic, full=True))
+        print(get_heuristic_description(heuristic, full=True))
     else:
-        raise ValueError("Unknown command %s", args.command)
+        raise ValueError("Unknown command %s", command)
     return
 
 
-def ensure_heuristic_arg(args):
+def ensure_heuristic_arg(heuristic=None):
     from ..utils import get_known_heuristic_names
-    if not args.heuristic:
+    if not heuristic:
         raise ValueError("Specify heuristic using -f. Known are: %s"
                          % ', '.join(get_known_heuristic_names()))
 
@@ -113,25 +114,9 @@ def main(argv=None):
         lgr.warning("Nothing to be done - displaying usage help")
         parser.print_help()
         sys.exit(1)
-    # To be done asap so anything random is deterministic
-    if args.random_seed is not None:
-        import random
-        random.seed(args.random_seed)
-        import numpy
-        numpy.random.seed(args.random_seed)
-    # Ensure only supported bids options are passed
-    if args.debug:
-        lgr.setLevel(logging.DEBUG)
-    # Should be possible but only with a single subject -- will be used to
-    # override subject deduced from the DICOMs
-    if args.files and args.subjs and len(args.subjs) > 1:
-        raise ValueError(
-            "Unable to processes multiple `--subjects` with files"
-        )
 
-    if args.debug:
-        setup_exceptionhook()
-    process_args(args)
+    kwargs = vars(args)
+    process_args(**kwargs)
 
 
 def get_parser():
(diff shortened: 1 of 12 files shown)


yarikoptic:
Thank you @tsalo! It is a worthwhile refactoring. I will review within next days. Cheers

heudiconv/workflows/run.py
```diff
@@ -0,0 +1,343 @@
...
+    random_seed : int or None, optional
+        Random seed to initialize RNG. Default is None.
+    dcmconfig : str or None, optional
+        JSON file for additional dcm2niix configuration. Default is None.
+    queue : {'SLURM', None}, optional
+        Batch system to submit jobs in parallel. Default is None.
+    queue_args : str or None, optional
+        Additional queue arguments passed as single string of space-separated
+        Argument=Value pairs. Default is None.
+    """
```
yarikoptic, line 198: I don't know an "easy" way around it ATM, may be you would find some, but such duplication of docstrings in here and in cmdline argparse setup eventually would diverge.  I don't know an "easy" way since most of the time then people indeed either duplicate like you have done, or come up with some evolved way to generate both cmdline and python interfaces from the same specification (like we did in datalad, and I saw similar ones elsewhere).

In principle it should be possible for many (if not all -- didn't do matching which one is pertinent only here and not cmdline) to take argparse specification and use it to populate a docstring here, and while at it may be even verify consistency of the function.  It feels like a common problem, so may be someone already came up with a similar solution?  I was looking for the one for `click` based interfaces but found none... could you check @tsalo around to see if smth like that was already done, or may be just identify a showstopper which would forbid it?

tsalo: I tried looking around for tools to synchronize argument parsers with associated functions' docstrings, but couldn't find anything. If it's any consolation, we have the same separation in `tedana` and, while there's no automated control, keeping the function docstring up-to-date with the CLI is something we've been able to do. I could also add a line in the function description telling users to refer to the command-line help for the *definitive* argument descriptions.

tsalo:
I've moved `heudiconv.workflows.heudiconv_workflow()` to `heudiconv.main.workflow()` and eliminated the `workflows` module. The reason I originally had it was that I assumed that `heudiconv.cli.monitor.monitor()` would also be converted to a workflow at some point, but I hadn't actually looked at it at the time.

heudiconv/workflows/run.py
```diff
@@ -0,0 +1,343 @@
...
+import os.path as op
+import sys
+import logging
+
+from .. import __version__, __packagename__
+from ..parser import get_study_sessions
+from ..utils import load_heuristic, anonymize_sid, treat_infofile, SeqInfo
+from ..convert import prep_conversion
+from ..bids import populate_bids_templates, tuneup_bids_json_files
+from ..queue import queue_conversion
```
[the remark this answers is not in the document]
tsalo: Should be good now.

heudiconv/workflows/run.py
```diff
@@ -0,0 +1,343 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
tsalo: Done!

heudiconv/workflows/run.py
satra: wouldn't the default click behavior work. see the top example here: https://click.palletsprojects.com/en/7.x/

satra: for any developer it should be clear from looking at the decorators.

tsalo: I think that that would take whatever's in the function docstring and make it available in the command line help, but when you try `help(function)` in Python, you get a very long and unrelated string for `click.Command` instead of the function. I don't think click would help with within-Python documentation.

mgxd:
this looks good to me, thanks!

i'll wait for @yarikoptic's blessing before merging

heudiconv/main.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+def ensure_heuristic_arg(heuristic=None):
+    from .utils import get_known_heuristic_names
+    if not heuristic:
+        raise ValueError("Specify heuristic using -f. Known are: %s"
+                         % ', '.join(get_known_heuristic_names()))
+
+
+def workflow(dicom_dir_template=None, files=None, subjs=None,
```
yarikoptic, line 105: no sane person should invoke `workflow` providing any of those as positional args, I will add a leading `*, ` ;)

heudiconv/main.py
```diff
@@ -0,0 +1,340 @@
...
+
+    # processed_studydirs = set()
+
+    for (locator, session, sid), files_or_seqinfo in study_sessions.items():
+
+        # Allow for session to be overloaded from command line
+        if session is not None:
+            session = session
+        if locator is not None:
+            locator = locator
```
yarikoptic, line 268: Smart pycharm greyed out `locator` in the signature of this function which raised my concern.  Here code like
```
        if args.session is not None:
            session = args.session
        if args.locator is not None:
            locator = args.locator
```
was, so `args.` was providing defaults, whenever now loop variables mask away function arguments.  This needs to be fixed up.  It is pity that apparently we have no tests which would have triggered either of those cases!

heudiconv/cli/run.py
```diff
@@ -140,238 +30,147 @@ def get_parser():
...
-
-        anon_sid = anonymize_sid(sid, args.anon_cmd) if args.anon_cmd else None
-        if args.anon_cmd:
-            lgr.info('Anonymized {} to {}'.format(sid, anon_sid))
-
-        study_outdir = op.join(outdir, locator or '')
-        anon_outdir = args.conv_outdir or outdir
-        anon_study_outdir = op.join(anon_outdir, locator or '')
-
-        # TODO: --datalad  cmdline option, which would take care about initiating
```
tsalo, line 326: There is a datalad cmdline option now. Can this be removed?

heudiconv/main.py
tsalo: I've seen that once, but never used it. It's awesome! I also added a numpydoc Notes section to the docstring to clarify usage for those who, like me, might be confused by the way it's defined. I can remove it if you want though.

heudiconv/main.py
tsalo: That's a head-scratcher. I combed through the workflow with an eye out for this kind of pattern, but didn't find any other cases.

heudiconv/cli/run.py
yarikoptic: I guess -- good find

heudiconv/main.py
yarikoptic: Notes are good, thank you!

yarikoptic:
4 failures in travis now, e.g.
```
___________________________ test_notop[bidsoptions1] ___________________________
heudiconv/tests/test_heuristics.py:167: in test_notop
    assert op.exists(pjoin(tmppath, 'Halchenko/Yarik/950_bids_test4'))
E   AssertionError: assert False
E    +  where False = <function exists at 0x7fc5bc715bf8>('/tmp/pytest-of-travis/pytest-0/test_notop_bidsoptions1_0/Halchenko/Yarik/950_bids_test4')
E    +    where <function exists at 0x7fc5bc715bf8> = op.exists
E    +    and   '/tmp/pytest-of-travis/pytest-0/test_notop_bidsoptions1_0/Halchenko/Yarik/950_bids_test4' = pjoin('/tmp/pytest-of-travis/pytest-0/test_notop_bidsoptions1_0', 'Halchenko/Yarik/950_bids_test4')
```
otherwise looks good. After failures addressed, I will try to comb through it locally a bit more and then merge

tsalo:
I finally fixed the testing, so I think it's ready for your review.

tsalo:
I think that this PR could be helpful for the BIDS Curation GUI BrainHack project (https://github.com/ohbm/hackathon2020/issues/188), so I thought I'd ping it. Is there anything I can do to move this forward or will it have to hold off until after the hackathon?

yarikoptic:
FWIW: reviewing it now... gimme some minutes

heudiconv/main.py
```diff
@@ -0,0 +1,363 @@
...
+    elif command == 'heuristics':
+        from .utils import get_known_heuristics_with_descriptions
+        for name_desc in get_known_heuristics_with_descriptions().items():
+            print("- %s: %s" % name_desc)
+    elif command == 'heuristic-info':
+        ensure_heuristic_arg(heuristic)
+        from .utils import get_heuristic_description
+        print(get_heuristic_description(heuristic, full=True))
+    else:
+        raise ValueError("Unknown command %s", command)
```
yarikoptic, line 111: I will fix this (present in original version too)

heudiconv/main.py
```diff
@@ -0,0 +1,363 @@
...
+            import traceback
+            import pdb
+            traceback.print_exception(type, value, tb)
+            # print()
+            pdb.post_mortem(tb)
+        else:
+            lgr.warning(
+              "We cannot setup exception hook since not in interactive mode")
+
+    sys.excepthook = _pdb_excepthook
```
yarikoptic, line 42: FWIW: we have this function not covered by tests -- tested manually seems to still work )

heudiconv/main.py
```diff
@@ -0,0 +1,363 @@
...
+    #
+    if not heuristic:
+        raise RuntimeError("No heuristic specified - add to arguments and rerun")
+
+    if queue:
+        lgr.info("Queuing %s conversion", queue)
+        iterarg, iterables = ("files", len(files)) if files else \
+                             ("subjects", len(subjs))
+        queue_conversion(queue, iterarg, iterables, queue_args)
+        sys.exit(0)
```
yarikoptic, line 272: I think we must not `sys.exit(0)` here now -- I will RF so we exit in the CLI invocation

yarikoptic:
ok, pushed those few tune ups.  as far as I see it -- good to go. Let's wait for green and merge!

tsalo:
Thanks @yarikoptic!

PR #443, opened 2020-04-20
RF(+BF somewhat): remove no longer needed dependencies etc

yarikoptic:
Ran into it while testing neurodebian package for 0.8.0 -- ran into:
```
$> docker run -it --rm reproin-test-0.8.0:latest --help
  File "/usr/bin/heudiconv", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 3191, in <module>
    @_call_aside
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 3175, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 3204, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 583, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 900, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 786, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'pathlib' distribution was not found and is required by heudiconv
```
because `info.py` claims it as a dependency, and there were no `pip` install which would have pulled it from pypi (even though it is included in python itself)

### DIFF
--- a/heudiconv/info.py
+++ b/heudiconv/info.py
@@ -23,9 +23,7 @@ PYTHON_REQUIRES = ">=3.5"
 REQUIRES = [
     'nibabel',
     'pydicom',
-    'nipype >=1.0.0; python_version > "3.0"',
-    'nipype >=1.0.0,!=1.2.1,!=1.2.2; python_version == "2.7"',
-    'pathlib',
+    'nipype >=1.0.0',
     'dcmstack>=0.8',
     'etelemetry',
     'filelock>=3.0.12',
--- a/setup.py
+++ b/setup.py
@@ -19,9 +19,7 @@ def main():
     # Get version and release info, which is all stored in heudiconv/info.py
     info_file = op.join(thispath, 'heudiconv', 'info.py')
     with open(info_file) as infofile:
-        # exec(infofile.read(), globals(), ldict)
-        # Workaround for python 2.7 prior 2.7.8
-        eval(compile(infofile.read(), '<string>', 'exec'), globals(), ldict)
+        exec(infofile.read(), globals(), ldict)
 
 
     def findsome(subdir, extensions):


caldodge:
This is actually a major issue for heudiconv, since it requires features that exist in the built-in pathlib, but not the version installed via pip.  For example, the package version does not allow "exist_ok" as a parameter to "mkdir" in a Path object.

yarikoptic:
why would you install pathlib via pip if it is now included in the standard library and thus most likely pypi version is just outdated (last updated in 2014)?

caldodge:
I did not explicitly install pathlib.  Since heudiconv 0.8.0 (installed via
pip) lists pathlib as a requirement, PIP installed it.  I had to remove it,
remove pathlib references from the relevant locations, then reinstall it to
eliminate the problem. So apparently I can't trust whoever maintains
heudiconv on pypi to have the latest version.

yarikoptic:
;-) it does have the latest version!  it is just those (same) people just need to release heudiconv 0.8.1...

but the point here is that actually it should have been pathlib pypi maintainers which probably should have made it impossible to install pathlib for versions of python where it was already included.But let's get back to

> This is actually a major issue for heudiconv, since it requires features that exist in the built-in pathlib, but not the version installed via pip. For example, the package version does not allow "exist_ok" as a parameter to "mkdir" in a Path object.

```shell
(git)lena:~/picts/mris/heudiconv-master[master]git
$> git describe
v0.8.0-68-gd855f64
$> git grep exist_ok
$>
```

do you have more specific pointer of what feature of bundled pathlib we are using which is not present in pypi's?

caldodge:
Certainly.  mkdir in the built-in pathlib allows the parameter "exist_ok".
The version in the pathlib package does not. This was how we discovered it,
as a user who was trying out heudiconv received multiple messages that
mkdir didn't understand "exist_ok".

PR #446, opened 2020-05-05
[ENH] Adds extraction of physio signals from DICOMs

pvelasco:
It uses [`bidsphysio`](https://github.com/cbinyu/bidsphysio) to extract physiological signals from CMRR EPI DICOMs.

It includes regression test


Dockerfile
```diff
@@ -99,7 +105,8 @@ RUN export PATH="/opt/miniconda-latest/bin:$PATH" \
     && sync && conda clean -tipsy && sync \
     && bash -c "source activate base \
     &&   pip install --no-cache-dir --editable \
-             '/src/heudiconv[all]'" \
+             '/src/heudiconv[all]' \
+	     '/src/bidsphysio'" \
```
yarikoptic, line 109: this file is generated, shouldn't be modified directly. there is `utils/gen-docker-image.sh` to be tuned but even that one souldn't be changed in this case ;)

please first upload biosphysio to pypi and then just add it to `"extras"` within `heudiconv/info.py` nearby datalad.

heudiconv/convert.py
```diff
@@ -407,6 +409,43 @@ def convert_dicom(item_dicoms, bids_options, prefix,
                 shutil.copyfile(filename, outfile)
 
 
+def convert_physio(item_dicoms, bids_options, prefix):
+    """Save DICOM physiology as BIDS physio files
+    Parameters
```
yarikoptic, line 414: nit: add empty line before Parameters line

heudiconv/dicoms.py
```diff
@@ -265,8 +265,13 @@ def group_dicoms_into_seqinfos(files, grouping, file_filter=None,
         series_id = '-'.join(map(str, series_id))
         if mw.image_shape is None:
             # this whole thing has no image data (maybe just PSg DICOMs)
-            # nothing to see here, just move on
-            continue
+            # If this is a Siemens PhysioLog, keep it:
+            if mw.dcm_data.SeriesDescription.endswith('_PhysioLog'):
+                # just give it a dummy shape, so that we can continue:
+                mw.image_shape=(0,0,0)
```
yarikoptic, line 271: nit pep8: spaces around ` = `

heudiconv/tests/test_regression.py
```diff
@@ -84,6 +90,32 @@ def test_multiecho(tmpdir, subject='MEEPI', heuristic='bids_ME.py'):
         assert 'echo-' not in event
 
 
+@pytest.mark.skipif(not have_bidsphysio, reason="no bidsphysio")
+def test_physio(tmpdir, subject='samplePhysio', heuristic='bids_physio.py'):
```
yarikoptic, line 94: add 
```
@assert_cwd_unchanged(ok_to_chdir=True)
```
since there is `chdir` used inside

heudiconv/tests/test_regression.py
```diff
@@ -84,6 +90,32 @@ def test_multiecho(tmpdir, subject='MEEPI', heuristic='bids_ME.py'):
...
+    tmpdir.chdir()
+    outdir = tmpdir.mkdir('out').strpath
+    template = op.join("{subject}/*/*.dcm")
```
yarikoptic, line 97: what is the point of op.join here if just a single string?

heudiconv/convert.py
```diff
@@ -407,6 +409,43 @@ def convert_dicom(item_dicoms, bids_options, prefix,
...
+        item_dicoms = list(map(op.abspath, item_dicoms)) # absolute paths
+        if len(item_dicoms)>1:
+            lgr.warning(
+                "More than one PHYSIO file has been found for this series. "
+                "If each file corresponds to a different signal, all is OK. "
+                "If multiple files have the same signal, only the signal "
+                "from the last file will be saved."
+            )
+        for dicom_file in item_dicoms:
+            dcm2bidsphysio.dcm2bids( dicom_file, prefix )
```
yarikoptic, line 446: nit pep8 (?): unneeded spaces at ()?

yarikoptic (changes requested):
sweet!!!!!  Left notes from a very initial cursory review.

heudiconv/convert.py
```diff
@@ -407,6 +409,43 @@ def convert_dicom(item_dicoms, bids_options, prefix,
...
+        If not None then save to BIDS format. List may be empty
+        or contain bids specific options
+    prefix : string
+        Conversion outname
+    Returns
+    -------
+    None
+    """
+    if bids_options is not None:
+        try:
```
mgxd, line 428: to reduce indentation when possible
```suggestion
    if bids_options is None:
        return
```

heudiconv/convert.py
```diff
@@ -407,6 +409,43 @@ def convert_dicom(item_dicoms, bids_options, prefix,
...
+            from bidsphysio import dcm2bidsphysio
+        except ImportError:
+            lgr.warning(
+                "Dcm2bidsphysio not found. "
+                "Not extracting physiological recordings."
+            )
+            return
+
+        item_dicoms = list(map(op.abspath, item_dicoms)) # absolute paths
+        if len(item_dicoms)>1:
```
mgxd, line 438:
```suggestion
        if len(item_dicoms) > 1:
```

heudiconv/convert.py
```diff
@@ -407,6 +409,43 @@ def convert_dicom(item_dicoms, bids_options, prefix,
...
+    Parameters
+    ----------
+    item_dicoms : list of filenames
+        DICOMs to save
+    bids_options : list or None
+        If not None then save to BIDS format. List may be empty
+        or contain bids specific options
+    prefix : string
+        Conversion outname
+    Returns
```
mgxd, line 423:
```suggestion

    Returns
```

heudiconv/dicoms.py
```diff
@@ -265,8 +265,13 @@ def group_dicoms_into_seqinfos(files, grouping, file_filter=None,
         series_id = '-'.join(map(str, series_id))
         if mw.image_shape is None:
             # this whole thing has no image data (maybe just PSg DICOMs)
-            # nothing to see here, just move on
-            continue
+            # If this is a Siemens PhysioLog, keep it:
+            if mw.dcm_data.SeriesDescription.endswith('_PhysioLog'):
```
mgxd, line 269: is this a siemens standard?

heudiconv/dicoms.py
mgxd, line 271:
```suggestion
                mw.image_shape = (0, 0, 0)
```

mgxd (commented):
this is awesome, thank you!

a few style suggestions, though we might just consider `black`ing the whole repo and adhering to a stricter style guide.

another consideration while looking at this - WDYT about fetching the data via datalad instead of hosting it in the repo?

yarikoptic:
re blacking: I don't mind at all.  But we have outstanding sizeable PRs so it would introduce conflicts... will need to check with @tsalo and @bpinsard

pvelasco:
> another consideration while looking at this - WDYT about fetching the data via datalad instead of hosting it in the repo?

Of course, that would be great. You can just download the data from [my branch](https://github.com/cbinyu/heudiconv/tree/dcm_physio/heudiconv/tests/data/samplePhysio), or I can send you a zip file with them.  Once they are in DataLad, I'll modify the regression tests accordingly.

heudiconv/dicoms.py
pvelasco: Fixed in 9de5114535f08c3dd3d8a177326a920127d1262c

tsalo:
@yarikoptic, I like automated code formatting, but I think running black would introduce a ton of conflicts in #442. Could aggressive re-formatting be something that's done in its own PR after the pending larger PRs are handled?

mgxd:
of course, it would make it nearly unreviewable! I meant showing some love to formatting prior to the next release, and then inserting a style check into our CI

yarikoptic:
Just a note (I am ok either way): theoretically (didn't try yet) if there is no value in individual commits in a or (or if they are way to split from a full diff, eg different files), and both branches are auto blacked when pr branch having master branch merged right before blacking - should be easy to force refresh PRs with blacked version ;-) I can try if we decide to black it all

yarikoptic:
Re data - I think it is time to remove the middle man me! Let's (ab)use lfs of GitHub with datalad: http://handbook.datalad.org/en/latest/basics/101-138-sharethirdparty.html?highlight=Lfs#use-github-for-sharing-content . I will initiate a repo with the content of this PR data files, and invite everyone possibly interested ;-)...

yarikoptic:
Thinking about it - iirc dicoms compress nicely. I will try with enabling shared encryption - that should compress them while stored in lfs.

heudiconv/convert.py
pvelasco: Thanks for the suggestion!
Fixed in cbinyu/heudiconv@532c630c4170d4db031e78d8aecde35156eb714b
(I didn't accept your suggested change because I wanted to change the indentation of the function in the same commit).

heudiconv/dicoms.py
pvelasco: Yes.

heudiconv/info.py
```diff
@@ -43,7 +43,10 @@
     'extras': [
         'duecredit',  # optional dependency
     ],  # Requires patched version ATM ['dcmstack'],
-    'datalad': ['datalad >=%s' % MIN_DATALAD_VERSION]
+    'datalad': ['datalad >=%s' % MIN_DATALAD_VERSION],
+    'dcm2bids': [
```
tsalo, line 47: What about `physio` for this dependency group?

pvelasco: Good idea!

heudiconv/info.py
```diff
@@ -43,7 +43,10 @@
     'extras': [
         'duecredit',  # optional dependency
     ],  # Requires patched version ATM ['dcmstack'],
-    'datalad': ['datalad >=%s' % MIN_DATALAD_VERSION]
+    'datalad': ['datalad >=%s' % MIN_DATALAD_VERSION],
+    'physio': [
+        'bidsphysio.dcm2bids >=1.4.3; python_version>"3.5"',   # if dicoms with physio need to be converted
```
yarikoptic, line 48:
```suggestion
        'bidsphysio.dcm2bids >=1.4.3',   # if dicoms with physio need to be converted
```

as 3.5 is already below what we support

heudiconv/info.py
```diff
@@ -43,7 +43,10 @@
...
+    ]
 }
 
 # Flatten the lists
```
yarikoptic, line 52: all those binary files below adding up to over 1MB -- I think should be moved to some external dataset and used via `fetch_data`. ATM `fetch_data` hardcodes to access from datasets.datalad.org but I think we should change that -- if full url is provided - just use that URL, and then share them on some repo, could be even not annex, just regular git repo on github.

Later we should improve this `fetch_data` to make data persistent locally etc... but not for this PR.

heudiconv/utils.py
```diff
@@ -46,7 +46,8 @@
     'patient_sex',           # 23
     'date',                  # 24
     'series_uid',            # 25
- ]
+    'time',                  # 26
+]
```
yarikoptic, line 50: this would conflict with WiP of #637 / #581 but the main concern is that I do not see it needed/used at all by the target need of this PR -- conversion of physio data. Spurious wishful change? ;-)

yarikoptic (changes requested):
Sorry it took so long.  Did a pass. Also of interest either bidsphysio can convert fresh WiP 1066 Siemens physio data.

edit: I will convert it to draft. Please undraft/ping whenever ready for re-review etc

PR #451, opened 2020-05-12
[ENH] Retain sub-second resolution in scans files

### LINKED ISSUE #447: Retain sub-second resolution in acq_time column of scans tsv files
### Summary

I am working on a method to synchronize physio data from either BioPac or Labchart with BIDS-format scan data in order to convert the physio data to BIDS format (see https://github.com/physiopy/phys2bids/pull/219). This requires sub-second acquisition timing information from the scan data. At the moment, I use the `AcquisitionTime` field heudiconv retains in the json files, but I think that the scans.tsv file would be easier to access. The only unfortunate thing is that, in the conversion, sub-second timing information is discarded. See below:

https://github.com/nipy/heudiconv/blob/5681e0dce5250745c7c2f58866c5e86a93667fc4/heudiconv/bids.py#L407-L411

If that information could be retained, it would make it easier to synchronize the scans with externally-acquired data, including physio.

tsalo:
Closes #447.

Changes proposed:
- Keep sub-second timing information in ContentTime when compiling `acq_time` for scans.tsv files.

### DIFF
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -26,6 +26,19 @@ from .utils import (
 
 lgr = logging.getLogger(__name__)
 
+# Fields to be populated in _scans files. Order matters
+SCANS_FILE_FIELDS = OrderedDict([
+    ("filename", OrderedDict([
+        ("Description", "Name of the nifti file")])),
+    ("acq_time", OrderedDict([
+        ("LongName", "Acquisition time"),
+        ("Description", "Acquisition time of the particular scan")])),
+    ("operator", OrderedDict([
+        ("Description", "Name of the operator")])),
+    ("randstr", OrderedDict([
+        ("LongName", "Random string"),
+        ("Description", "md5 hash of UIDs")])),
+])
 
 class BIDSError(Exception):
     pass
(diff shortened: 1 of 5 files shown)


heudiconv/bids.py
```diff
@@ -406,9 +406,9 @@ def get_formatted_scans_key_row(dcm_fn):
     # parse date and time and get it into isoformat
     try:
         date = dcm_data.ContentDate
-        time = dcm_data.ContentTime.split('.')[0]
-        td = time + date
-        acq_time = datetime.strptime(td, '%H%M%S%Y%m%d').isoformat()
+        time = dcm_data.ContentTime
+        td = time + ':' + date
+        acq_time = datetime.strptime(td, '%H%M%S.%f:%Y%m%d').isoformat()
```
yarikoptic, line 411: is `.%f` guaranteed to be present (above code would work without)? I wonder if there should be a check for `'.'` to be found in td?
If you could while at it place it into an auxiliary function and give it a dedicated tiny unittest - would be awesome... but if not -- that is ok ;) just feels like worthwhile

tsalo: Sounds good. Just added a new function in utils, as well as a test for it.

yarikoptic:
pushed a little tune up and then decided to check BIDS... heh heh

https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#scans-file 

```
Datetime should be expressed in the following format 2009-06-15T13:45:30 (year, month, day, hour 
(24h), minute, second; this is equivalent to the RFC3339 "date-time" format, time zone is always 
assumed as local time). 
```

and according to https://tools.ietf.org/html/rfc3339 and https://en.wikipedia.org/wiki/ISO_8601 seems sub-second is allowed (phew) BUT also wikipedia says  `However, the number of decimal places needs to be agreed to by the communicating parties. For example, in Microsoft SQL Server, the precision of a decimal fraction is 3, i.e., "yyyy-mm-ddThh:mm:ss[.mmm]"`.  

So I am afraid that
- we need to submit a PR to bids-specification suggesting allowing `.[mmmmmm]` before we could change this column formatting. (takers?)
- **but** meanwhile we could just add a new one, e.g. `acq_time_precise` or alike.  (I will push a commit or two for that here now while at it)

yarikoptic:
I am not yet sure if I like it but here it was pushed.  @tsalo could you please raise a question or a PR with bids-specification about .microseconds? ;)

tsalo:
@yarikoptic My bad, I should have reviewed the specification before opening the issue/PR. I've opened https://github.com/bids-standard/bids-specification/issues/469, so we'll see if we can get microseconds supported.

yarikoptic:
Since BIDS community is so receptive to the change, may be we could just proceed with subsecond in `acq_time` (and later update BIDSVersion we claim to be compatible with whenever that gets released) !  Meanwhile I had paid tribute to my laziness in the last commit I did last night -- and split it now into two, so we could just drop the last one which introduces `acq_time_precise` if we decide to go that route ;)

yarikoptic:
So, @tsalo -- should I just drop the last commit and let's proceed with new improved `acq_time`? ;-)

tsalo:
@yarikoptic That sounds good to me!

yarikoptic:
ok, done. For the record -- pushed `_abandoned_acq_time_precise` tag with that abandoned now commit into our fork on https://github.com/dbic/heudiconv

tsalo:
Now that https://github.com/bids-standard/bids-specification/pull/470 is merged, can we merge this or should we wait for the validator (https://github.com/bids-standard/bids-validator/issues/957)?

yarikoptic:
oh, we definitely should merge ;)

PR #453, opened 2020-05-21
[BF] remove unassigned embedfunc outputs

ariekahn:
This fixes an error of `ERROR: Embedding failed: 'NoneType' object is not subscriptable`.

It looks like `embed_dicom_and_nifti_metadata` no longer returns a json and metadata, but the Node wrapping it in `embed_metadata_from_dicoms` still was expecting an output. This was getting caught inside the try/catch block, but caused it to skip the rest of the processing in `embed_metadata_from_dicoms`.

This patch just removes the outputs from the node, since it appears to be unused.


yarikoptic:
Thank you! indeed it returns nothing now!

PR #459, opened 2020-06-19
ENH: Use AcquisitionTime for acq_time scans file field

### LINKED ISSUE #450: Use AcquisitionTime for acq_time column of scans tsv files
### Summary

It looks like heudiconv [uses ContentTime](https://github.com/nipy/heudiconv/blob/75f2850f7f07a95ff5875aba1dfd76c09e4be201/heudiconv/bids.py#L408-L411) in the scans tsv files instead of AcquisitionTime, although I think that AcquisitionTime would be more appropriate (i.e., closer to the actual start of the scan).

Quoting @pvelasco in https://github.com/physiopy/phys2bids/pull/219#issuecomment-626899271: 
>>In dicoms, SeriesTime, InstanceCreationTime, and ContentTime are very similar, but AcquisitionTime is generally ~20 seconds earlier than the rest. We need to determine which time is closest to the scanner's trigger pulse.

>From some documentation I found from Siemens on image reconstruction:

>>AcquisitionTime: The Time the acquisition of data started.

>From an answer given by Chris Roden (dcm2niix):

>>Instance Creation Time 0008,0013 initially reflects the time an image leaves a parallel processing reconstructor, so it has a variable delay with regards to the acquisition time. Acquisition Time is more meaningful for data.

>So, if the reconstruction involves something more than a straight FFT (i.e., GRAPPA, SENSE, etc.), that will explain the ~20 sec delay you find between the AcquisitionTime and the other times.
>If I remember correctly, for Siemens scanners, the "scanner trigger" is sent at the time of the excitation pulse of the first slice of each (collected) volume. So the very first trigger will be sent just a few milliseconds before the first chunk of data for the first slice for the first volume is acquired (AcquisitionTime).

This issue is related to #447.

tsalo:
Closes #450. 

Changes proposed:
- Change ContentTime to AcquisitionTime for acq_time field in scans file.
- Change ContentDate to AcquisitionDate for acq_time field in scans file.

### DIFF
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -406,7 +406,7 @@ def get_formatted_scans_key_row(dcm_fn):
     # parse date and time and get it into isoformat
     try:
         date = dcm_data.ContentDate
-        time = dcm_data.ContentTime.split('.')[0]
+        time = dcm_data.AcquisitionTime.split('.')[0]
         td = time + date
         acq_time = datetime.strptime(td, '%H%M%S%Y%m%d').isoformat()
     except (AttributeError, ValueError) as exc:
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -404,7 +404,7 @@ def get_formatted_scans_key_row(dcm_fn):
     """
     dcm_data = dcm.read_file(dcm_fn, stop_before_pixels=True, force=True)
     # we need to store filenames and acquisition times
-    # parse date and time and get it into isoformat
+    # parse date and time of start of run acquisition and get it into isoformat
     try:
         date = dcm_data.ContentDate
         time = dcm_data.AcquisitionTime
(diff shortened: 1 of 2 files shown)


heudiconv/bids.py
```diff
@@ -404,10 +404,10 @@ def get_formatted_scans_key_row(dcm_fn):
     """
     dcm_data = dcm.read_file(dcm_fn, stop_before_pixels=True, force=True)
     # we need to store filenames and acquisition times
-    # parse date and time and get it into isoformat
+    # parse date and time of start of run acquisition and get it into isoformat
     try:
         date = dcm_data.ContentDate
-        time = dcm_data.ContentTime
+        time = dcm_data.AcquisitionTime
```
yarikoptic, line 410: is there AcquisitionDate?  very unlikely but still possible that date could be different in some very rare case of scanning around midnight.

tsalo: I _do_ have AcquisitionDate in my dicoms, and the standard has it listed, so I'll switch out ContentDate with AcquisitionDate.

tsalo: Should be good now.

yarikoptic:
Let the heaven not to fall upon our weak heads! ;)

yarikoptic:
Thank you @tsalo !

PR #461, opened 2020-06-26
FIX: Convert sets to lists for filename updaters

### LINKED ISSUE #460: Multi-echo and uncombined name updaters assume lists instead of sets
### Summary

When converting multi-echo data without an EchoNumber metadata field, `heudiconv.convert.update_multiecho_name()` attempts to index `echo_times`. The problem is that this parameter is a `set`, rather than a `list`, so doing this raises an error. The same should apply to `heudiconv.convert.update_uncombined_name()`.

Here is the traceback:
```python
  File "/opt/miniconda-latest/bin/heudiconv", line 33, in <module>
    sys.exit(load_entry_point('heudiconv', 'console_scripts', 'heudiconv')())
  File "/home/code/heudiconv/heudiconv/cli/run.py", line 24, in main
    workflow(**kwargs)
  File "/home/code/heudiconv/heudiconv/main.py", line 344, in workflow
    grouping=grouping,)
  File "/home/code/heudiconv/heudiconv/convert.py", line 211, in prep_conversion
    dcmconfig=dcmconfig,)
  File "/home/code/heudiconv/heudiconv/convert.py", line 459, in convert
    overwrite=overwrite)
  File "/home/code/heudiconv/heudiconv/convert.py", line 713, in save_converted_files
    bids_meta, this_prefix_basename, echo_times
  File "/home/code/heudiconv/heudiconv/convert.py", line 326, in update_multiecho_name
    echo_number = echo_times.index(metadata['EchoTime']) + 1
AttributeError: 'set' object has no attribute 'index'
```

This wasn't caught by the tests for these functions because I mistakenly used lists as the arguments in the tests.

EDIT: Introduced in #424, so at least it hasn't affected any releases.

### Platform details:

Choose one:
- [ ] Local environment

- [X] Container: Docker using current master (as of f4c2dd2).

- Heudiconv version: Current master (as of f4c2dd2).

tsalo:
Closes #460.

Changes proposed:
- In `heudiconv.convert.save_converted_files()`, convert variables `echo_times` and `channel_names` from sets to lists by sorting them. This fixes a bug in the filename updating functions (`heudiconv.convert.update_multiecho_name()` and `heudiconv.convert.update_uncombined_name()`) wherein these variables are indexed, which is not a supported method for `set`s.
- In `heudiconv.convert.update_multiecho_name()` and `heudiconv.convert.update_uncombined_name()` add typechecking assertions for the arguments `echo_times` and `channel_names`, respectively. These will raise an error if the variables are not lists.
- In `heudiconv.convert.save_converted_files()`, fill missing values in the variables `echo_times`, `channel_names`, and `image_types` with `False` instead of `nan`. This is necessary because the function determines if files are multi-echo, uncombined, and/or complex-valued based on whether these arrays have more than one element _after_ filtering out missing data. The filtering step did not work on `nan`s, but does work on `bool`s.


tsalo:
Test failures seem to be related specifically to datalad, with the error:

```python
AttributeError: 'Dataset' object has no attribute 'add'
```

I assume that the failures are due to some change in datalad rather than my changes, but I could be wrong.

tsalo:
NaNs have been used in the `metadata.get()` step since before #424, but we started using them with the filter in #424.

To clarify the filter:

```python
from math import nan
echo_times = (0.001, 0.002, nan)
set(filter(bool, echo_times))
```
Expected: `{0.001, 0.002}`
Actual result: `{nan, 0.001, 0.002}`

However, it works fine with either `False` or `None`, e.g.:

```python
echo_times = (0.001, 0.002, None)
set(filter(bool, echo_times))
```
Expected: `{0.001, 0.002}`
Actual: `{0.001, 0.002}`

I can include cases with missing metadata (whether None or False) in the tests, but the filter statements are buried in `heudiconv.convert.save_converted_files()`, which is a large function and difficult to build tests for.

heudiconv/convert.py
```diff
@@ -690,12 +694,14 @@ def save_converted_files(res, item_dicoms, bids_options, outtype, prefix, outnam
...
-            channel_names.add(metadata.get('CoilString', nan))
-            image_types.update(metadata.get('ImageType', [nan]))
+            echo_times.add(metadata.get('EchoTime', False))
+            channel_names.add(metadata.get('CoilString', False))
+            image_types.update(metadata.get('ImageType', [False]))
         is_multiecho = len(set(filter(bool, echo_times))) > 1  # Check for varying echo times
         is_uncombined = len(set(filter(bool, channel_names))) > 1  # Check for uncombined data
         is_complex = 'M' in image_types and 'P' in image_types  # Determine if data are complex (magnitude + phase)
+        echo_times = sorted(echo_times)  # also converts to list
+        channel_names = sorted(channel_names)  # also converts to list
```
yarikoptic, line 704: also -- is it guaranteed to be ok to sort here while possibly breaking correspondence to `res_files`?  not sure

tsalo: The order of echo times and channel names here determines which metadata value corresponds to which index value, but I don't think that's a problem. The association between metadata value (from `res_files`) and index value happens in the specific function.

mckenziephagen (commented):
lgtm, but why did you start sorting the echo times now?  wouldn't `list(echo_times)` also work?

tsalo:
@mckenziephagen Are you referring to how converting a set to a list automatically sorts the list items? TBH when I wrote this, I didn't realize that was consistent behavior. I thought it just placed them in a random order... which maybe was an improvement from Python 2 to 3? In any case, I could change `sorted` to `list`.

mckenziephagen:
Oh, I'm not sure if that's consistent behavior. I was actually wondering why sort at all, because that wasn't present in previous versions of `convert.py`.

tsalo:
Oh gotcha. It's not enforced in the specification (and won't be until 2.0), but I think it's good procedure to associate the entity index with the metadata fields in ascending order (e.g., echo-1 matches the earliest echo time). It's been a while since I wrote this code, but I believe that the order of the echo times from `save_converted_files()` will impact the order in `update_multiecho_name()`.

tsalo:
Is there anything blocking this PR? I can replace `sorted` with `list` if folks would prefer it, but I don't think it's a problem for matching the actual echo times with the `echo` index, and it should ensure that the `echo` index does correspond to ascending echo times.

yarikoptic (commented):
Left minor suggestion to avoid `assert` + a question on changing the logic from using `nan` (evals to True) to `False` (will be filtered out).

heudiconv/convert.py
```diff
@@ -690,12 +694,14 @@ def save_converted_files(res, item_dicoms, bids_options, outtype, prefix, outnam
         for metadata in bids_metas:
             if not metadata:
                 continue
-            echo_times.add(metadata.get('EchoTime', nan))
```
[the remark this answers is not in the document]
tsalo: I think you're right, but I don't know what the answer is. I don't know the proper approach when `EchoTime` is missing (if it ever even is). We have `EchoNumber`, but that has its own odd behavior, and I don't know if `EchoNumber` would be present when `EchoTime` is not.

Would there be a problem with just checking against EchoTime? Is there any case where that field would be missing completely?

@pvelasco, you started checking against EchoNumber in #293. Do you recall if there's a reason for using EchoNumber instead of just indexing EchoTimes? I see a note that the MGH ME-MPRAGE sequence uses EchoNumber, but the numbers do not follow the same order as the EchoTimes. Is it important that the order of the echo numbers be retained from the sequence instead of being placed in ascending order?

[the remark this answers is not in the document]
pvelasco: I think the reason was that `echo_times` (that gets passed to `update_multiecho_name`) is currently a `set`, so ATM it doesn't preserve the order given by `EchoNumber`.
In principle, indexing the sorted `EchoTime`s should work, but I haven't tested it.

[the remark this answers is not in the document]
tsalo: There's a comment that talks about EchoNumber:

https://github.com/nipy/heudiconv/blob/3f9a504270f83d00bb483c8f52ae0c228fc7d808/heudiconv/convert.py#L719-L727

The gist seems to be that EchoNumber might not be in the same order as EchoTime. I wonder if it's worth it to prefer EchoNumber over EchoTime in those cases. If we choose to just use EchoTime (and assume that EchoTime is always present), then that simplifies things.

heudiconv/convert.py
```diff
@@ -250,7 +250,7 @@ def update_complex_name(metadata, filename, suffix):
         Scan metadata dictionary from BIDS sidecar file.
     filename : str
         Incoming filename
-    suffix : str
+    suffix_counter : str
```
yarikoptic, line 251: well, `suffix_counter` is also confusing to me... we need to fix something up here:
- `counter` suggests an `int` of some kind but we have it as `str`
- its docstring says "An index ..." but the rest of it is not really providing much clarity to me
- we use that `suffix` (or `suffix_counter`) as a `%s` to be added to `_part-`
- `part` takes a `label` [bids spec](https://bids-specification.readthedocs.io/en/stable/99-appendices/09-entities.html#part) from a specific set of values (phase, mag, real and imag) not an index/counter.

so, overall -- we can't insert an index/count here as we would with counter... or what am I missing?

tsalo: I think this is a holdover from when heudiconv used `rec` for this information. Before #424, the "suffix" was just a number that could be used with `rec`:

https://github.com/nipy/heudiconv/blob/fdea77ce6263d3f92e51dd8a24ac506e7a2fce4e/heudiconv/convert.py#L512

https://github.com/nipy/heudiconv/blob/fdea77ce6263d3f92e51dd8a24ac506e7a2fce4e/heudiconv/convert.py#L553-L560

Perhaps the most BIDS-y solution would be to try to assign the right value to `part`, but, failing that, assign "suffix" (with a better variable name) to `rec`. So instead of `rec-magnitude`/`res-phase`/`rec-2` (the old way) or `part-mag`/`part-phase`/`part-2` (the current way), we would have `part-mag`/`part-phase`/`rec-2`, which would be BIDS-compliant.

I'm still not sure what to call `suffix`/`suffix_counter` though. It _is_ an index into the list of filenames, but that integer is converted to a string for the sake of adding it to the updated filename.

yarikoptic: let's just raise the RuntimeError on this `else:` and completely drop that suffix* kwarg.  The rationale is that we invoke this function only if `is_complex` (see https://github.com/nipy/heudiconv/pull/461/files#diff-d19d089741b3844c199c8236ec1c0fcd43400b2ce53cd587d203255add863c39R797) which is
```python
is_complex = 'M' in image_types and 'P' in image_types  # Determine if data are complex (magnitude + phase)
```
so we must have both M and P, and thus should not hit this `else`.  We are test covering this line of code just because we have a dedicated test for that function: https://github.com/tsalo/heudiconv/blob/master/heudiconv/tests/test_convert.py#L18

or am I missing smth?

tsalo: Oh, that's a good point! That seems like a great solution.

tsalo: Should be done in [e3e3d4f](https://github.com/nipy/heudiconv/pull/461/commits/e3e3d4f5b3a1a45ca886c9bcd6e0c6f3775ce256). Let me know if that's not what you had in mind.

tsalo:
I'm confused. GitHub is telling me that this branch is up-to-date with master.

sitek:
Little bump since I'm running into #460's `set`/`index` issue now with SWI scans. Thanks for putting this together @tsalo!

yarikoptic:
> I'm confused. GitHub is telling me that this branch is up-to-date with master.

Ha, for some reason for this PR it just switched (persistently) to "rebase and merge" strategy for me, and that is what it was complaining about. Switched back to regular Merge and no complaints.

yarikoptic:
Ok, I still feel that situation is not yet totally clear but it should indeed become less foggy , so let's proceed with the merge.

Issue #462, opened 2020-07-22
Temporary directories linger after successful heudiconv run


mgxd:
Originally brought up in https://neurostars.org/t/heudiconv-v0-8-0-fails-to-remove-tmp-directories/15388

greydongilmore:
@mgxd thanks for opening this, looking forward to the patch 👍

PR #466, opened 2020-08-06
FIX/ENH/RF: Introduce configuration module and catch lingering tmpdirs

mgxd:
Spawned while fiddling with #462.

This config module improves accessibility of arguments across the various modules of the heudiconv workflow. Since `heudiconv` started as a single, massive script, the implementation favored run-level access to all / most variables.

During this "enh-factor", the legacy ``TempDirs`` class was removed, and most temporary directories were changed into
``tempfile.TemporaryDirectory`` instances, with the exception being the tmpdir housing the extracted tarball files. This is tracked directly by the config class, and should catch any leaked directories from lingering past heudiconv's execution.

P.S. - We are in desperate need of standardizing the coding style here - if/when this is and any other almost-ready PRs are merged, I'm inclined to run an autolinter (preferably `black`) and give the codebase some needed love.

### DIFF
--- a/heudiconv/config.py
+++ b/heudiconv/config.py
@@ -1,0 +1,185 @@
+"""
+A Python module for managing configurations across a heudiconv run.
+"""
+
+import atexit
+import os
+from pathlib import Path
+
+
+_initialized = False
+# avoid telemetry check if user desires
+_disable_et = bool(
+    os.getenv("NO_ET") is not None
+    or os.getenv("NIPYPE_NO_ET") is not None
+)
+os.environ["NIPYPE_NO_ET"] = "1"
+os.environ["NO_ET"] = "1"
+_latest_version = "Unknown"
+
+if not _disable_et:
+    # Ensure analytics hit only once
+    from contextlib import suppress
+    from requests import get, ConnectionError, ReadTimeout
+    with suppress((ConnectionError, ReadTimeout)):
+        res = get("https://rig.mit.edu/et/projects/nipy/heudiconv", timeout=0.1)
+    try:
+        _latest_version = res.json()['version']
+    except Exception:
+        pass
+
+
+class _Config:
+    """An abstract class forbidding instantiation."""
+
+    _paths = tuple()
+
+    def __init__(self):
+        """Avert instantiation."""
+        raise RuntimeError('Configuration type is not instantiable.')
+
+    @classmethod
+    def load(cls, settings, init=True):
+        """Store settings from a dictionary."""
+        for k, v in settings.items():
+            if v is None:
+                continue
+            if k in cls._paths:
+                setattr(cls, k, Path(v).absolute())
+                continue
+            if k in cls._multipaths:
+                setattr(cls, k, tuple(Path(i).absolute() for i in v))
+            if hasattr(cls, k):
+                setattr(cls, k, v)
+
+        if init:
+            try:
+                cls.init()
+            except AttributeError:
+                pass
+
+    @classmethod
+    def get(cls):
+        """Return defined settings."""
+        out = {}
+        for k, v in cls.__dict__.items():
+            if k.startswith('_') or v is None:
+                continue
+            if callable(getattr(cls, k)):
+                continue
+            if k in cls._paths:
+                v = str(v)
+            if k in cls._multipaths:
+                v = tuple(str(i) for i in v)
+            out[k] = v
+        return out
+
+
+class workflow(_Config):
+    """Configure run-level settings."""
+
+    _paths = (
+        'dcmconfig',
+        'outdir',
+        'conv_outdir',
+    )
+    _multipaths = (
+        'files',
+    )
+
+    dicom_dir_template = None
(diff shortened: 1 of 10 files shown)


satra:
> give the codebase some needed love.

love is always much appreciated. 

in the future, this may also be a good codebase to move over to pydra and have the dcm2niix interface be a pydra task.

tsalo:
> P.S. - We are in desperate need of standardizing the coding style here - if/when this is and any other almost-ready PRs are merged, I'm inclined to run an autolinter (preferably `black`) and give the codebase some needed love.

I'll open an issue about this.

mgxd:
> in the future, this may also be a good codebase to move over to pydra and have the dcm2niix interface be a pydra task.

Definitely, our nipype usage is extremely minimal, I think just the `dcm2niix` wrapper and provenance tracking.

yarikoptic:
Re nipype - I thought we had a dedicated issue to just get that dependency removed in favor of straight execution of dcm2niix but I did only find even older https://github.com/nipy/heudiconv/issues/6

yarikoptic:
If there a chance to
- get tempdir handling in a separate PR?
- get some description of what config module is about (ie explained in PR description)?

It is a large PR to review as a whole with two aspects interleaved and not entirely clear config intent

mgxd:
> get tempdir handling in a separate PR?

We definitely could do this, but this largely defeats the whole motivation behind the refactoring. Previously, there was no true run-level place where standalone functions could send information to. And as a result, we were not properly removing temporary directories generated at:
https://github.com/nipy/heudiconv/blob/ca402a9fa701551bf2c74f72d77223aaad7e5d68/heudiconv/parser.py#L52-L69

by adding `heudiconv.config._tmpdirs`, we can run some basic cleanup (similar to what `TempDirs.__del__` did). LMKWYT

> get some description of what config module is about (ie explained in PR description)?

It's largely inspired from [work introduced in fmriprep early this year](https://github.com/poldracklab/fmriprep/pull/2018), with the omission of writing out a file with all arguments used. Though I might be inclined to bring that in and dump them inside `<outdir>/.heudiconv`. One future change I've been wanting to see is reducing the number of arguments that need to be passed across all the various functions - the config would allow setting them once and then fetching them via `config.<workflow/other>.<var>`.

That said, I've only tested this via the supplied tests, so I am hesitant to push for a merge until this goes through a more rigorous trial (any testing on your side would be super helpful 😄 )

heudiconv/utils.py
```diff
@@ -536,3 +489,5 @@ def get_datetime(date, time, *, microseconds=True):
     if not microseconds:
         datetime_str = datetime_str.split('.', 1)[0]
     return datetime_str
+
+
```
yarikoptic, line 493: nit picking: not needed even for styling

mgxd, line 493:
```suggestion
```

heudiconv/config.py
```diff
@@ -0,0 +1,188 @@
...
+if not _disable_et:
+    # Ensure analytics hit only once
+    from contextlib import suppress
+    from requests import get, ConnectionError, ReadTimeout
+    with suppress((ConnectionError, ReadTimeout)):
+        res = get("https://rig.mit.edu/et/projects/nipy/heudiconv", timeout=0.5)
+    try:
+        _latest_version = res.json()['version']
+    except Exception:
+        pass
```
yarikoptic, line 28: Why direct calls via `requests` and not  use of `etelemetry` module client? by all means we want to avoid reimplementing what it does/created for.

heudiconv/config.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class _Config:
+    """An abstract class forbidding instantiation."""
+
+    _paths = tuple()
+
+    def __init__(self):
+        """Avert instantiation."""
+        raise RuntimeError('Configuration type is not instantiable.')
```
yarikoptic, line 38: why not to start using `abc` and thus declare this as an `abc.abstractmethod`? according to https://docs.python.org/3/library/abc.html#abc.abstractmethod could be applied to `@classmethod` as well

heudiconv/config.py
```diff
@@ -0,0 +1,188 @@
...
+
+def get(flat=False):
+    """Get config as a dict."""
+    settings = {
+        'workflow': workflow.get(),
+    }
+    if not flat:
+        return settings
+
+    return {'.'.join((section, k)): v
```
yarikoptic, line 158: note: seems to be not covered by the tests.

yarikoptic:
@mgxd would you have time/motivation to bring this over the finish line?

mgxd:
Looking back after dropping this, I think this can probably be closed and the relevant bits extracted into a fresh PR. I will admit I'm not up to date on the expected way to use `heudiconv`, so perhaps the singleton config option can be scrapped in favor of a more robust per-dataset approach. I probably won't get to this anytime soon though

However, I do think I'll carve out some time to ensure we handle the removal of temporary directories

PR #471, opened 2020-10-27
ENH: Support extracting DICOMs from ZIP files (and possibly other archives) by switching to use shutil.unpack_archive instead of tarfile module functionality

HippocampusGirl:
Dear heudiconv team,

I was recently working on converting a DICOM dataset to BIDS with heudiconv. Our imaging facility sends us the DICOM files as ZIP archives.

While I could extract the ZIP files manually before running heudiconv, I thought it may be useful to have on-the-fly extraction just like for tarballs. I implemented this via the built-in Python module "zipfile".


HippocampusGirl:
I just discovered that this works even without the change I uploaded. Sorry, my bad :-)

yarikoptic:
interesting -- stock heudiconv doesn't work for me:
```shell
$> heudiconv -f reproin --bids -o /tmp/h --files bids_anat-T1w.zip 
INFO: Running heudiconv version 0.9.0 latest 0.9.0
INFO: Analyzing 1 dicoms
INFO: Filtering out 0 dicoms based on their filename
WARNING: Ignoring bids_anat-T1w.zip since not quite a "normal" DICOM: 'FileDataset' object has no attribute 'SeriesNumber'
INFO: Generated sequence info for 0 studies with 0 entries total
INFO: Need to process 0 study sessions
```
so may be there is still a need in this/such PR.  Do you remember with which version it has worked?  

Haven't reviewed changes but tried with this PR and above worked... so we still need it seems to me.

yarikoptic:
From cursory review looks ok but needs a unittest to ensure that it works. Test could reuse existing files, zip them, and do conversion.

HippocampusGirl:
Interesting, maybe some zip files work with `tarfile` and some do not. Or maybe I made a mistake and got things mixed up, who knows ;-)

I agree that a unit test would be important. I am currently quite busy, but I will have time to write one next week, so please stay tuned.

heudiconv/parser.py
```diff
@@ -76,24 +77,44 @@ def get_extracted_dicoms(fl):
...
+            # check content
+            zmembers = zf.infolist()
+            # get all files, assemble full path in tmp dir
+            zf_content = [m.filename for m in zmembers if not m.is_dir()]
+            # store full paths to each file, so we don't need to drag along
+            # tmpdir as some basedir
+            sessions[session] = [op.join(tmpdir, f) for f in zf_content]
+            session += 1
+            # extract into tmp dir
+            zf.extractall(path=tmpdir)
```
yarikoptic, line 114: to remove code duplication etc, IMHO we should follow advice of https://docs.python.org/3.8/library/tarfile.html and "use ... the higher-level functions in [shutil](https://docs.python.org/3.8/library/shutil.html#archiving-operations)." . so `if tarfile.is_tarfile(f) or zipfile.is_zipfile(f): `... `shutil.unpack_archive` and then just list files in the `tmpdir` .

yarikoptic (changes requested):
Not sure yet how it could have worked on zip, most likely some fluke (e.g. `.tar.gz` was named with `.zip` extension or alike). So I think this enhancement has value so I will keep it live. 

It will also need a unittest (produce .zip in the test code, give to function, see that does the right thing).

Meanwhile I will convert to `draft`. please advise if you are going to work on it further @HippocampusGirl . Cheers

HippocampusGirl:
Hi @yarikoptic, I think I got confused when I closed the pull request. I also agree that this is a valuable feature and have been using my changes for multiple projects. 

I want to work on this some more, but cannot prioritize this right now. If you or anyone else is willing to step in, then things may be faster.

psadil:
> I want to work on this some more, but cannot prioritize this right now. If you or anyone else is willing to step in, then things may be faster.

Hello @HippocampusGirl, is it still the case that this feature is on hold? I would be interested in the ability to work with zip files and would be happy to bring this pull request over the finish line.

HippocampusGirl:
Hi @psadil, That would be wonderful!

psadil:
> Hi @psadil, That would be wonderful!

Great!

I wonder what's going to be a good way to contribute code. I have [this potential solution](https://github.com/psadil/heudiconv/tree/enh/add-zipfile-support) that builds on the pull request's initial commit, [d076a07](https://github.com/nipy/heudiconv/pull/471/commits/d076a07ceb85922f60156735790c6b8d791cadc1). Do I submit a pull request into [HippocampusGirl:add-zipfile-support](https://github.com/HippocampusGirl/heudiconv/tree/add-zipfile-support) so that it shows up in this #471 request? Can I directly add on to #471? Should I open a new pull request for the main nipy/heudiconv repo?

@yarikoptic perhaps you have guidance?

Thank you

HippocampusGirl:
Hi @psadil, I have given you write access to my repository. You should be able to push into my branch now, so that it shows up as part of the pull request :-)

psadil:
@yarikoptic (tagging you as the original reviewer) - I think this may be ready to go. As you requested in the original review, the unpacking is now done with the `shutil` module, and several new tests of `get_extracted_dicoms` have been added.

heudiconv/parser.py
```diff
@@ -24,6 +25,7 @@
 
 _VCS_REGEX = r'%s\.(?:git|gitattributes|svn|bzr|hg)(?:%s|$)' % (op.sep, op.sep)
 
+_UNPACK_FORMATS = tuple(chain.from_iterable([x[1] for x in shutil.get_unpack_formats()]))
```
yarikoptic, line 28: sweet, just FYI although not even sure if preferable, but more "native python" way is
```suggestion
_UNPACK_FORMATS = tuple(sum((x[1] for x in shutil.get_unpack_formats()), []))
```
or just
```suggestion
_UNPACK_FORMATS = sum((x[1] for x in shutil.get_unpack_formats()), [])
```
if ok to stay with a list

heudiconv/parser.py
```diff
@@ -7,8 +7,9 @@
 
 from collections import defaultdict
 
-import tarfile
-from tempfile import mkdtemp
+import shutil
+import typing
```
yarikoptic, line 11: I think frequent `typing.` makes code too cluttered. I am ok with imports of the constructs we use as we already do elsewhere

```
heudiconv/dicoms.py:from typing import List, Optional
```

yarikoptic: note -- ATM we do not check typing information at all. Filed https://github.com/nipy/heudiconv/issues/653

heudiconv/parser.py
```diff
@@ -60,58 +62,87 @@ def find_files(regex, topdir=op.curdir, exclude=None,
...
-    to different sessions, so here we would group into sessions and return
-    pairs  `sessionid`, `files`  with `sessionid` being None if no "sessions"
-    detected for that file or there was just a single tarball in the list
-    """
-    # TODO: bring check back?
-    # if any(not tarfile.is_tarfile(i) for i in fl):
-    #     raise ValueError("some but not all input files are tar files")
+    Parameters
+    ----------
+    fl : Iterable[str]
```
yarikoptic, line 73: I am not sure how valuable (I dislike duplication) is type annotation here.   I would say strip them off.
Unfortunately, apparently this question is not yet addressed within numpy itself , so there is no definitive mandate/answer - https://github.com/numpy/numpydoc/issues/196 , so I would have gone for 

```suggestion
    fl
```

heudiconv/tests/test_archives.py
```diff
@@ -0,0 +1,90 @@
...
+
+@pytest.fixture
+def get_dicoms_zip(tmpdir: Path) -> typing.List[str]:
+    tmp_file = tmpdir / "dicom"
+    archive = shutil.make_archive(
+        str(tmp_file),
+        format="zip", 
+        root_dir=TESTS_DATA_PATH, 
+        base_dir="01-anat-scout")
+    return [archive]
```
yarikoptic, line 34: Sorry -- I have too strong allergy to duplication... what about

```suggestion
def get_dicoms_archive(format: str, tmpdir: Path) -> typing.List[str]:
    tmp_file = tmpdir / "dicom"
    archive = shutil.make_archive(
        str(tmp_file),
        format=format, 
        root_dir=TESTS_DATA_PATH, 
        base_dir="01-anat-scout")
    return [archive]

@pytest.fixture
def get_dicoms_gztar(tmpdir: Path) -> typing.List[str]:
    return get_dicoms_archive("gztar", tmpdir)

@pytest.fixture
def get_dicoms_zip(tmpdir: Path) -> typing.List[str]:
    return get_dicoms_archive("zip", tmpdir)
```

FWIW found some HOWTOs on parametric fixtures in https://stackoverflow.com/questions/42228895/how-to-parametrize-a-pytest-fixture .  Since tests below seems to do exactly that -- could be RFed to use one of those approaches or just become `@pytest.mark.parametrize(`'ed and use suggested above `get_dicoms_archive` directly.

heudiconv/tests/test_archives.py
```diff
@@ -0,0 +1,90 @@
...
+
+def test_get_extracted_dicoms_single_session_is_none(get_dicoms_gztar: typing.List[str]):
+    for session_, _ in get_extracted_dicoms(get_dicoms_gztar):
+        assert session_ is None
+
+
+def test_get_extracted_dicoms_multple_session_integers(get_dicoms_gztar: typing.List[str]):
+    sessions = []
+    for session, _ in get_extracted_dicoms(get_dicoms_gztar + get_dicoms_gztar):
+        sessions.append(session)
```
yarikoptic, line 50: IMHO 

```suggestion
    sessions = [
        session
        for session, _ in get_extracted_dicoms(get_dicoms_gztar + get_dicoms_gztar)
    ]
```
is a bit cleaner

heudiconv/tests/test_archives.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def test_get_extracted_dicoms_from_file_list(get_dicoms_list: typing.List[str]):
+    for _, files in get_extracted_dicoms(get_dicoms_list):
+        assert all(op.isfile(file) for file in files)
+
+
+def test_get_extracted_have_correct_permissions(get_dicoms_gztar: typing.List[str]):
+    for _, files in get_extracted_dicoms(get_dicoms_gztar):
+        assert all(stat.S_IMODE(os.stat(file).st_mode) == 448 for file in files)
```
yarikoptic, line 74: why not to fold this within `test_get_extracted_dicoms_from_*` test -- running each one separately might have benefits of cause but virtually doubles run time since the same drill needs to be done for the fixtures twice

heudiconv/tests/test_archives.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def test_get_extracted_are_absolute(get_dicoms_gztar: typing.List[str]):
+    for _, files in get_extracted_dicoms(get_dicoms_gztar):
+        assert all(op.isabs(file) for file in files)
```
yarikoptic, line 79: same here -- I would have folded into the same test operating on `get_extracted_dicoms` and testing various aspects to be consistent

heudiconv/parser.py
```diff
@@ -60,58 +62,87 @@ def find_files(regex, topdir=op.curdir, exclude=None,
...
+        An iterable (e.g., list or tuple) of files to process.
+
+    Returns
+    -------
+    ItemsView[int | None, list[str]]
```
yarikoptic, line 78: Unfortunately on this aspect there even were not discussion so I initiated one https://github.com/numpy/numpydoc/issues/196#issuecomment-1488734277 . I guess we are doomed to keep it as this for now here.

heudiconv/parser.py
```diff
@@ -60,58 +62,87 @@ def find_files(regex, topdir=op.curdir, exclude=None,
...
+    session.
+    """
+    input_list_of_unpacked = any(fl) and all(
+        not t.endswith(_UNPACK_FORMATS) for t in fl
+        )
+    
+    if not (
+        input_list_of_unpacked or
+        all(t.endswith(_UNPACK_FORMATS) for t in fl)
+        ):
```
yarikoptic, line 104:
```suggestion
    ):
```

<details>
<summary>well, we need to add some auto-blacking eventually etc, but ATM here and elsewhere let's stay consistent with dominant way -- aligning to original block indent, not nested:
</summary> 

```shell
❯ git grep -3  '^ *):$'
heudiconv/bids.py-            all([other[k] == v for k, v in self._entities.items()])
heudiconv/bids.py-            and self.extension == other.extension
heudiconv/bids.py-            and self.suffix == other.suffix
heudiconv/bids.py:        ):
heudiconv/bids.py-            return True
heudiconv/bids.py-        else:
heudiconv/bids.py-            return False
--
heudiconv/convert.py-    if (
heudiconv/convert.py-        op.exists(target_heuristic_filename) and
heudiconv/convert.py-        file_md5sum(target_heuristic_filename) != file_md5sum(heuristic.filename)
heudiconv/convert.py:    ):
heudiconv/convert.py-        # remake conversion table
heudiconv/convert.py-        reuse_conversion_table = False
heudiconv/convert.py-        lgr.info(
--
heudiconv/dicoms.py-    if mw.dcm_data[0x0008, 0x0016].repval in (
heudiconv/dicoms.py-        'Raw Data Storage',
heudiconv/dicoms.py-        'GrayscaleSoftcopyPresentationStateStorage'
heudiconv/dicoms.py:    ):
heudiconv/dicoms.py-        return
heudiconv/dicoms.py-    try:
heudiconv/dicoms.py-        file_studyUID = mw.dcm_data.StudyInstanceUID
--
heudiconv/external/dcmstack.py-    if not (
heudiconv/external/dcmstack.py-        ("No module" in e_str and "dicom" in e_str) or
heudiconv/external/dcmstack.py-        ('has been removed in pydicom version' in e_str)
heudiconv/external/dcmstack.py:    ):
heudiconv/external/dcmstack.py-        raise
heudiconv/external/dcmstack.py-    # a butt plug due to rename of dicom -> pydicom
heudiconv/external/dcmstack.py-    import sys
--
heudiconv/heuristics/bids_PhoenixReport.py-        if (
heudiconv/heuristics/bids_PhoenixReport.py-            'PhoenixZIPReport' in s.series_description
heudiconv/heuristics/bids_PhoenixReport.py-            and s.image_type[3] == 'CSA REPORT'
heudiconv/heuristics/bids_PhoenixReport.py:        ):
heudiconv/heuristics/bids_PhoenixReport.py-            info[phoenix_doc].append({'item': s.series_id})
heudiconv/heuristics/bids_PhoenixReport.py-        if 'scout' in s.series_description.lower():
heudiconv/heuristics/bids_PhoenixReport.py-            info[scout].append({'item': s.series_id})
--
heudiconv/tests/test_bids.py-        expected_prefix,
heudiconv/tests/test_bids.py-        simulation_function,
heudiconv/tests/test_bids.py-        match_param
heudiconv/tests/test_bids.py:):
heudiconv/tests/test_bids.py-    """
heudiconv/tests/test_bids.py-    Test find_compatible_fmaps_for_session.
heudiconv/tests/test_bids.py-
--
heudiconv/tests/test_bids.py-    expected_prefix,
heudiconv/tests/test_bids.py-    simulation_function,
heudiconv/tests/test_bids.py-    match_param
heudiconv/tests/test_bids.py:):
heudiconv/tests/test_bids.py-    """
heudiconv/tests/test_bids.py-    Test populate_intended_for.
heudiconv/tests/test_bids.py-    Parameters

```
</details>

heudiconv/parser.py
```diff
@@ -60,58 +62,87 @@ def find_files(regex, topdir=op.curdir, exclude=None,
...
+        raise ValueError("Some but not all input files are archives.")
```
yarikoptic, line 105: I think prior code was handling that just fine -- and add non-archive to `None` session.  Is there a reason to introduce this breaking behavior?  I think it might have really bad manifestations (didn't check) e.g. by pointing to a folder with some tarballs and some files to just be ignored which we ignore if not dicom later (?) in the code, etc.

heudiconv/parser.py
```diff
@@ -60,58 +62,87 @@ def find_files(regex, topdir=op.curdir, exclude=None,
...
+    
+    sessions: typing.DefaultDict[
+        typing.Optional[int], 
+        typing.List[str]
+        ] = defaultdict(list)
+    
+    if not isinstance(fl, list):
         fl = list(fl)
```
yarikoptic, line 113: do we even need this "casting"?

heudiconv/parser.py
```diff
@@ -60,58 +62,87 @@ def find_files(regex, topdir=op.curdir, exclude=None,
...
+        # fl being empty)
+        session = 0
+        # needs sorting to keep the generated "session" label deterministic
+        for _, t in enumerate(sorted(fl)):
+            # Each file extracted must be associated with the proper session,
+            # but the high-level shutil does not have a way to list the files
+            # contained within each archive. So, files are temporarily
+            # extracted into unique tempdirs
+            # cannot use TempDirs since will trigger cleanup with __del__
+            tmpdir = tempdirs(prefix="heudiconvDCM")
```
yarikoptic, line 129: make that tmpdir `0o700` right away here so no files extracted would have any moment in time when they are accessible to non-authorized user (i.e. until chmoding below)?

yarikoptic (changes requested):
- major concern -- seems to be a change in behavior in case of a mixed list of files/archives. I would prefer to keep original behavior unless there is really a strong need
- some minor comments on styling etc.

heudiconv/parser.py
yarikoptic: in that issue I was pointed out to https://pypi.org/project/sphinx-autodoc-typehints/ , which seems might work with numpy style , could you check it out?

psadil:
> * major concern -- seems to be a change in behavior in case of a mixed list of files/archives. I would prefer to keep original behavior unless there is really a strong need

Ah, sorry. I thought that I was making something explicit that was previously only implicit. I didn't have a strong need. I'll update it to match the current behavior

> * some minor comments on styling etc.

Great! I'm looking through these and will revise accordingly

heudiconv/parser.py
psadil: Yeah, can do. Were you thinking about this as a separate issue, or something to figure out for this pull?

heudiconv/parser.py
psadil: I'll check. There was a similar cast before
 https://github.com/nipy/heudiconv/blob/c83e06c6d3c71a43c3d570f9c05d12f36dc3a066/heudiconv/parser.py#L85-L86
but perhaps that can be dropped altogether

heudiconv/parser.py
yarikoptic: sphinx extension -- separate PR probably. Here -- I would just keep it concise aiming for future sanity of the tools and the extension ;)

heudiconv/parser.py
yarikoptic: yeah -- that is what I wondered on why it was needed before. I wish smarter earlier us had left a comment ;)

heudiconv/parser.py
psadil: For this strategy: https://github.com/nipy/heudiconv/blob/254dce7f996760dc50d890530a9b028e6851a36c/heudiconv/parser.py#L110
I think that `_UNPACK_FORMATS` needs to remain a `tuple`

heudiconv/parser.py
psadil: Sounds good. Looking forward to newer versions that allow the use of builtins like `list` or `|` :)

heudiconv/parser.py
psadil: Good point! This seems especially relevant considering that the files are unpacked into a generic tmp directory, which could be a surprising risk to privacy.

heudiconv/tests/test_archives.py
psadil: First time I've seen parameterized fixtures, but that makes sense to me. Note that I've kept the `get_dicoms_gztar` to allow for a test that mixes archived and non-archived files 

https://github.com/nipy/heudiconv/blob/99b0bf6109732c365721250fd7843c70a74002c1/heudiconv/tests/test_archives.py#L75-L77

psadil:
Okay, now revised

heudiconv/parser.py
psadil: Hmm, it seems like that extension might already be (partly) integrated.

https://github.com/nipy/heudiconv/blob/c83e06c6d3c71a43c3d570f9c05d12f36dc3a066/docs/conf.py#L44

Though, not sure whether another configuration needs to be set

psadil: FWIW, the extension seems to more or less "just work" by adding `sphinx_autodoc_typehints` to 

https://github.com/nipy/heudiconv/blob/c83e06c6d3c71a43c3d570f9c05d12f36dc3a066/docs/conf.py#L43-L4

And the extension also handles return types from the signature.

psadil:
Checking in on this -- the code is ready for another round of review

yarikoptic (approved):
code styling is a bit odd to me, but without black I cannot insist really ;)

Otherwise I think it looks great and let's give it a try!  Thank you @HippocampusGirl and @psadil !!!

PR #454, opened 2020-05-21
ENH: zenodo referencing in README.rst + support for ducredit for heudiconv and reproin

yarikoptic:
Individual commits might provide more info.  Here is a result of a sample run with duecredit enabled:

```
DueCredit Report:
- Flexible DICOM converter for organizing brain imaging data / heudiconv (v 0.8.0) [1]
  - ReproIn heudiconv heuristic for turnkey conversion into BIDS / heudiconv.heuristics.reproin (v 0.8) [2]
- I/O library to access to common neuroimaging file formats / nibabel (v 3.0.1) [3]
- Scientific tools library / numpy (v 1.17.4) [4]

3 packages cited
1 module cited
0 functions cited

References
----------

[1] Halchenko, Y. et al., 2020. nipy/heudiconv v0.8.0.
[2] Visconti di Oleggio Castello, M. et al., 2020. ReproNim/reproin 0.6.0.
[3] Brett, M. et al., 2015. Nibabel 2.0.1.
[4] Van Der Walt, S., Colbert, S.C. & Varoquaux, G., 2011. The NumPy array: a structure for efficient numerical computation. Computing in Science & Engineering, 13(2), pp.22–30.
DUECREDIT_ENABLE=1 heudiconv -o /tmp/out -f reproin --files   5.33s user 0.35s system 98% cpu 5.767 total
```

<details>
<summary>With BibTeX output produced by `duecredit summary --format bibtex`:</summary> 

```bibtex
@misc{https://doi.org/10.5281/zenodo.1012598,
  doi = {10.5281/ZENODO.1012598},
  url = {https://zenodo.org/record/1012598},
  author = {Halchenko, Yaroslav and Goncalves, Mathias and Castello, Matteo Visconti Di Oleggio and {Satrajit Ghosh} and Hanke, Michael and Salo, Taylor and {, Dae} and Kent, James and {Pvelasco} and Amlien, Inge and Brett, Matthew and Tilley, Steven and Markiewicz, Chris and Lukas, Darren Christopher and Gorgolewski, Chris and Kim, Sin and {Joerg Stadler} and Kaczmarzyk, Jakub and Lee, John and Lurie, Daniel and Pellman, John and Braun, Henry and Melo, Bruno and Poldrack, Benjamin and Schiffler, Björn and {Michał Szczepanik} and Samuels, Kalle and Carlin, Johan and Feingold, Franklin and Kahn, Ari},
  title = {nipy/heudiconv v0.8.0},
  publisher = {Zenodo},
[...]

PR #477, opened 2020-12-07
ENH: Replace rec entity with part entity for complex-valued data

### LINKED ISSUE #476: Replace rec entity with part entity for complex-valued data
### Summary

Now that bids-standard/bids-specification#424 has been merged, BIDS will support the `part` entity for dissociating magnitude and phase data in its next release. Currently, heudiconv uses `rec-<magnitude|phase>` instead. To fix this, [`heudiconv.convert.update_complex_name()`](https://github.com/nipy/heudiconv/blob/d855f64d5013f8a0e41789766a094d3c3a91552f/heudiconv/convert.py#L236) needs to be updated to use `part` instead of `rec` (also `magnitude` needs to changed to `mag`).

### Platform details:

Choose one:
- [ ] Local environment

- [ ] Container

- Heudiconv version:

tsalo:
Closes #476.

Changes proposed:

- Replace `rec-<magnitude|phase>` with `part-<mag|phase>` now that BIDS supports the `part` entity.
- Update associated test.

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -235,7 +235,7 @@ def prep_conversion(sid, dicoms, outdir, heuristic, converter, anon_sid,
 
 def update_complex_name(metadata, filename, suffix):
     """
-    Insert `_rec-<magnitude|phase>` entity into filename if data are from a
+    Insert `_part-<mag|phase>` entity into filename if data are from a
     sequence with magnitude/phase part.
 
     Parameters
@@ -254,7 +254,10 @@ def update_complex_name(metadata, filename, suffix):
         Updated filename with rec entity added in appropriate position.
     """
     # Some scans separate magnitude/phase differently
-    unsupported_types = ['_bold', '_phase',
+    # A small note: _phase is deprecated, but this may add part-mag to
+    # magnitude data while leaving phase data with a separate suffix,
+    # depending on how one sets up their heuristic.
+    unsupported_types = ['_phase',
                          '_magnitude', '_magnitude1', '_magnitude2',
                          '_phasediff', '_phase1', '_phase2']
     if any(ut in filename for ut in unsupported_types):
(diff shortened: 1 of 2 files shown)


yarikoptic:
I am all of the change and great to have BIDS supporting it (thank you @tsalo for making it happen).  
My questions/concerns:
- is BIDS already released with this change?  what would be the version?  I have requested your review on https://github.com/nipy/heudiconv/pull/454 and if accepted -- we would need to boost BIDS version specified there to match here for the BIDS version which has `_part`
- are you aware of any processing pipeline/app which could have relied on `_rec-` and now would get broken if we start using `_part-`?

tsalo:
> is BIDS already released with this change?

`part` (and `inv` and `flip`) haven't been released yet. I believe they will be fairly soon though (i.e., as soon as BEP005 and/or BEP001 are fully merged in).

> are you aware of any processing pipeline/app which could have relied on `_rec-` and now would get broken if we start using `_part-`?

I am not aware of any. Possibly [`phaseprep`](https://github.com/ostanley/phaseprep), but that hasn't had any releases yet, and I can open an issue to be safe. I'm pretty sure that most pipelines that _should_ take part into account don't do so even with `rec` already in use. For example, fMRIPrep doesn't deselect single-band reference images with `rec-phase` at the moment (which could cause problems if you have complex-valued single-band reference images and fMRIPrep tries to use them for reference image creation), so I will need to follow up there and possibly elsewhere.

Also, support for `part-phase` + `_bold` is new (and indeed `rec-phase` + `_bold` wasn't support here) because the `_phase` suffix was not deprecated until this upcoming release. As a result, a number of tools will need to actively deselect phase data with the `_bold` suffix anyway, whether using `rec` or `part`.

yarikoptic:
so it seems that it should be safe to proceed.  We would just not forget to make it a "major" (well "0.minor") release with a corresponding CHANGELOG entry.  Thank you!

Issue #474, opened 2020-11-04
Heudiconv: Automatically include “IntendedFor” field in the metadata


ChristianNSchmitz:
Dear heudiconv-team,
we are using heudiconv and fmriprep for preprocessing functional MRI data. As we would like to use fieldmaps to unwarp these images, the fieldmap-metadata has to include the key “IntendedFor” in order to be used with fmriprep.
Is there a way for heudiconv to automatically include this data in the metadata using the heuristic file? I have seen that the package fw-heudiconv does include it, but we would appreciate using the docker version of heudiconv.
Thank you very much for you support!
Best,
Christian

yarikoptic:
Unfortunately not yet automagically. Original issue https://github.com/nipy/heudiconv/issues/138

mirestrepo:
In case it's helpful, [this](https://github.com/brown-bnc/xnat-tools/blob/39d9bd2d9a2788df3305dd3d91d08c5101cf979d/xnat_tools/bids_utils.py#L14) is our work around, we do it as a post processing step

pvelasco:
I have written a module to do this (and also add a "NumberOfVolumes" field to the functional jsons), and I would like to add it to this repo.
To run it, you need to have all the json files extracted in the destination folder, so what would be the place to insert the call?

yarikoptic:
I think it would be useful also as a "tool" (command) to run, so why not to
- add `populate_intended_for(path_to_bids_session)` to heudiconv/bids.py
- interface in `main.py:process_extra_commands` by sweeping across provided paths
- then we would need to interface it also in `convert.py:convert`

yarikoptic:
we have now implementation within heudiconv and also supplementary command to populate intended for in already converted data.

PR #482, opened 2020-12-18
ENH: Adds populate_intended_for for fmaps

pvelasco:
This commits adds a new `populate_intended_for` functionality to `heudiconv`. As its name indicates, it populates the `"IntendedFor"` field in the `fmap` json files of a given session.
At the moment, it can be called as a `--command` option. In the future, it should also be called automatically every time there is a bids extraction.

 * Currently, if the `"ShimSetting"` field is present in the `fmap` json file, it must match the `"ShimSetting"` in the other run (`func`, `dwi`, etc.) for this other run to be added to the `"IntededFor"` for that `fmap`. If the `"ShimSetting"` is not present in the json files, the match is based on the `<acq>` entity in the `fmap` file name: jsons with `acq-dwi` will be intended for runs in the `dwi` folder, `acq-fmri` or `acq-func` will be intended for runs in the `func` folder, etc.  The logic for enforcing that the `"ShimSetting"` (if present) matches is that if the scanner changes the shims between a `fmap` and its `"IntededFor"` run, the actual B0 field will be different between them, so the field measured (or estimated, from a `pepolar` pair) from the `fmap` should not be applied to correct the other run.
 * If more than one `fmap` fulfills the requirement above, the first of them (and any other `fmap` with matching `<acq>` and `<run>` entities, so that `pepolar` pairs and magnitude/phase pair have the same `"IntededFor"` runs) will get the `"IntendedFor"`. In the future, we might want to change this (e.g., so that it uses the one closest in time), or we might want to give the user the option to choose one or the other.

It addresses #474.


yarikoptic:
THANK YOU @pvelasco , I will review/try it out in upcoming days and provide feedback.  Happy holidays!

yarikoptic:
I saw activity ongoing through the productive (for some) holidays thus did not review.  Is it ready @pvelasco ? ;-)

pvelasco:
Yes. I think it is ready. I had some bugs in the unit tests, since I could not test them locally (#486), but they are now fixed.
If preferred, I could squash all those Bug Fix commits into one. Otherwise, it is ready for merging.

yarikoptic (changes requested):
I only started to review (so review is not complete, but decided to not delay until it is "well done" ;)), and still thinking about it 
- I really like an idea to (ab)use `_acq-` as an indicator for such grouping.  I even at some point thought to establish a dedicated one within `reproin` heuristic, so the "loosely specified in BIDS" `_acq-` seems to be reasonable one to use.  I just still wonder if such new behavior would not somehow introduce some undesired assignment, in particular if ShimSettings are not there (Philips, GE, ?).
- we might want to add additional, or alternative (if no ShimSettings?) constraints, e.g. based on resolution and "sform" (have not looked what would be the analog at DICOM level, ImageOrientationPatientDICOM?).  If I got it right it is still debatable if fieldmaps could be acquired/reconstructed at resolution different from the target images... but may be we for starters could constrain based on entire sform, thus at least guaranteeing that positioning and sizes are the same
- in one of our use cases I noted that shimming was just a bit different:

```
[-7371, 5580, 11925, -269, -131, -448, -111, -459] vs
[-7372, 5589, 11937, -276, -130, -360, -157, -451]
```

yet to clarify with original collection site/techs -- but it raised the question on either it is completely useless, or we could still associate, and thus here may be allow for some %difference?
- in case of ambiguity (multiple fmaps matching) - should we take the one closest in time?
- since different people might want/have different strategies - should this be more of a heuristic, i.e. other heuristics could reuse in theirs etc... or at least may be we should use this one only if heuristic does not provide one

heudiconv/bids.py
```diff
@@ -457,3 +461,145 @@ def convert_sid_bids(subject_id):
...
+    # get the BIDS folder (if "data_folder" includes the session, remove it):
+    if op.basename(path_to_bids_session).startswith('ses-'):
+        bids_folder = op.dirname(path_to_bids_session)
+    else:
+        bids_folder = path_to_bids_session
+
+    fmap_dir = op.join(path_to_bids_session, 'fmap')
+    if not op.exists(fmap_dir):
+        lgr.warning('Fmap folder not found in {}.'.format(path_to_bids_session))
+        lgr.warning('We cannot add the IntendedFor field')
```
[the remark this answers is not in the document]
pvelasco: > NB generally `%` form + passing positional args as args to the call is preferred for `lgr` since it would not even cause strings interpolation unless logging is enabled for the level.

Thanks for the tip!

Addressed in 9b397de36544e297b88cf27983bcbfb05f0d7b9a

heudiconv/bids.py
```diff
@@ -457,3 +461,145 @@ def convert_sid_bids(subject_id):
...
+    user's intentions were different, he/she will have to manually edit the
+    fmap json files.
+
+    Parameters:
+    ----------
+    path_to_bids_session : str or os.path
+        path to the session folder (or to the subject folder, if there are no
+        sessions).
+    """
+    lgr.info('')
```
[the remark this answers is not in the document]
pvelasco: Addressed in 9b397de36544e297b88cf27983bcbfb05f0d7b9a

heudiconv/bids.py
```diff
@@ -457,3 +461,145 @@ def convert_sid_bids(subject_id):
...
+
+    fmap_dir = op.join(path_to_bids_session, 'fmap')
+    if not op.exists(fmap_dir):
+        lgr.warning('Fmap folder not found in {}.'.format(path_to_bids_session))
+        lgr.warning('We cannot add the IntendedFor field')
+        return
+
+    # Get a list of all fmap json files in the session:
+    # (we will remove elements later on, so don't just iterate)
+    fmap_jsons = sorted([j for j in glob(op.join(path_to_bids_session, 'fmap/*.json'))])
```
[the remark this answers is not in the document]
pvelasco: For the current approach (using the first compatible fmap), yes, you need to make sure they are sorted. I don't know if `glob` returns them in alphabetical order.

pvelasco:
> * in one of our use cases I noted that shimming was just a bit different:
> 
> ```
> [-7371, 5580, 11925, -269, -131, -448, -111, -459] vs
> [-7372, 5589, 11937, -276, -130, -360, -157, -451]
> ```
> 
> yet to clarify with original collection site/techs -- but it raised the question on either it is completely useless, or we could still associate, and thus here may be allow for some %difference?

That's a very good question.
Because the magnetic field is linear, using images collected under shimset_1 (the fieldmap) to correct for distortions in other images collected under shimset_2 (the functional, diffusion, etc) should, in principle, leave a residual (i.e., uncorrected) distortion in the image that is identical to the distortion created by the field difference between both sets of shims (I don't know if this makes sense.)  So, if both sets of shims are very close, the residual distortion should be very small.
Then, it would make sense to allow for some %difference. But, how much? Maybe a 1%, a 5%?
Also, how do you calculate the total %diff: the mean of the absolute value of the coefficients? The root mean square?
I'm inclined to use the RMS, since the shim settings are the coefficients for either spherical harmonics or cylindrical harmonics (I forget which), which are normalized.
Finally, the first coefficient corresponds to the zero-order harmonic, which is just a constant B0 field. This corresponds to a shift of the image in the readout direction, so it should have no effect on the image distortions, so we don't need to take it into account.

pvelasco:
> I only started to review (so review is not complete, but decided to not delay until it is "well done" ;)), and still thinking about it
> [...]
> * since different people might want/have different strategies - should this be more of a heuristic, i.e. other heuristics could reuse in theirs etc... or at least may be we should use this one only if heuristic does not provide one

In response to the comment:
Yes, it all makes sense. We should cover for more complex situations: no shim-settings present, more than one fieldmap with matching shim-settings, etc. Also we should allow for the user to chose from a set of pre-defined strategies, if there is more than one match (first match, closest in time, etc.), or even allow the user to define his/her own.

I will think some more about it and propose some solution on how to organize it.

Stay tuned (or propose your own :-)

yarikoptic:
> I will think some more about it and propose some solution on how to organize it.

I do not have one handy, but was thinking that some might indeed just annotate with `_acq-` or some other field (since not necessarily protocol names are following  BIDS or ReproIn heuristic) .

Re RMS as the choice -- ignorant me doesn't know the exact nature of those values (i.e. of `[-7371, 5580, 11925, -269, -131, -448, -111, -459]`) :-/ , i.e are they comparable in absolute (as if e.g. were corresponding to an angle between -180 and 180 degrees) or need normalization to judge their deviations... could you please enlighten me (a url to a good description would be enough)

pvelasco:
> Re RMS as the choice -- ignorant me doesn't know the exact nature of those values (i.e. of `[-7371, 5580, 11925, -269, -131, -448, -111, -459]`) :-/ , i.e are they comparable in absolute (as if e.g. were corresponding to an angle between -180 and 180 degrees) or need normalization to judge their deviations... could you please enlighten me (a url to a good description would be enough)

I haven't been able to find a description of it. I remember reading (probably in the Siemens developers discussion group) that those numbers are the currents (in some units) going through the different shim coils.  [Siemens shim coils are designed](https://mriquestions.com/uploads/3/4/5/7/34572113/siemens_brochure_on_shimming.pdf) to generate shim fields shaped like the lowest order [real spherical harmonics] (https://en.wikipedia.org/wiki/Table_of_spherical_harmonics#Real_spherical_harmonics). So I assume the 8 elements of the ShimSetting array are the coefficients of first and second order.  (Looking at the numbers that show up in the Manual Adjustment tool, the units of those coefficients might be in Hz/cm or Hz/mm for the first order, Hz/cm2 or Hz/mm2 for the second order, etc.)

The distortions in the EPI images are proportional to the Delta_B0 [[Weiskopf et al., NeuroImage 24 (2005) 1068–1079](https://doi.org/10.1016/j.neuroimage.2004.10.012), eq. 3] (deviations from the nominal 3T). The Delta_B0 are the deviations from the nominal 3T, and will be the one created by the object you are imaging, plus the correction you manage with your shims. Because magnetic fields are linear, if you were to image the same object with two different shims (Delta_shim), the difference in the field Delta_B0_2 - Delta_B0_1, because the distortions created by the object are the same, will be Delta_shim, and the difference in distortions (across the whole image) between the images collected with the different shims will be proportional to the difference in fields generated by the two shim settings.

If we call the difference in distortion fields D:
D(x,y,z) ~ Delta_shim = Shim_2 - Shim_1

The shim field created by a given shim setting is:
Shim_i(x,y,z) = sum_{l,m}(s_{l,m}\_i * Y_{l,m}(x,y,z))

So the difference in distortions will be proportional to:
D(x,y,z) ~ sum_{l,m}(s_{l,m}\_2 * Y_{l,m}(x,y,z)) - sum_{l,m}(s_{l,m}\_1 * Y_{l,m}(x,y,z)) = sum_{l,m}[(s_{l,m}\_2 - s_{l,m}\_1) * Y_{l,m}(x,y,z)]

Therefore, for the question here: if you have a distorted image (say, functional or diffusion) that was collected using a shim set, and a field map (either `gre` or `pepolar`) collected with a different shim, can I still use my estimate of the field? The answer would be: depends on the relative error, doesn’t it?

One problem is how you quantify the total distortion in an image. That is, how to come up with a single number representative of the distortions across the whole image. One candidate would be to use the RMS of the distortions in the whole image (or a sphere of a given radius), or the mean absolute displacement.  For the case of the RMS, because the spherical harmonics are a set of basis functions, the displacements end up being the square root of the sum of the coefficients (that is, the RMS of the coefficients). Then, the error in the distortion when you measure the field with a different set of shims would be the RMS of the difference of the coefficients for the shims.

Another problem is that for the relative distortion you need the value of the distortion in the denominator, and that total distortion would not just be the distortion created by the shim, but also the one created by the object. And in the header, we have the shim information, but not the field created by the object. So, in principle, you can’t simply use the relative differences in the shim coefficients.

So I don’t know how to determine if it is OK to use a fieldmap if the shims are not the same.

kodiweera:
Re: 'pepolar' distortion images

'pepolar' distortion images are used to correct for epi distortions.  EPI distortions (displacements) depend on the field offset, PE(blips) gradient strength and Echo Spacing. Sequence related parameters are PE gradeint Stregth and Echo Spacing. Practically, change in the field offset during scaning is uncontrolable even with good shimming due to heating and subject motion. Minimization of this field offset will result in less distorions.

Shimming changes even during the acquistion. That's why there is a drift in the BOLD signal. The reason is global frequency shift due to heating of the shim and gradient elements. This effect can also be reduced with 'pepolar' distortion correction (e.g. topup).

The most important parameter is 'Echo Spacing'. This parameter should be matched between distortion images and relevant image modality to impose the same epi distortions between them. Usually, 'pepolar' images are acquired using a Spin Echo Sequence (SE-EPI, without diffusion gradients). So, in total we use three different sequences if we acquire BOLD(GE-EPI) and DWI (SE-EPI-Diff).

Possible binding points between 'pepolar' images and BOLD (DWI) images are;
1. Echo Spacing
2. Voxel size
3. Reciever Bandwitdh
4. Echo Train Length

Here is HCP detail about Spin Echo Based fieldmap correction;
See section 7. "Why Spin Echo Field Maps" 
 https://static-content.springer.com/esm/art%3A10.1038%2Fnn.4361/MediaObjects/41593_2016_BFnn4361_MOESM36_ESM.pdf

pvelasco:
> Shimming changes even during the acquistion. That's why there is a drift in the BOLD signal. The reason is global frequency shift due to heating of the shim and gradient elements. This effect can also be reduced with 'pepolar' distortion correction (e.g. topup).

There are two types of shims: active and passive. **Passive shims** are those pieces of metal placed around the bore that try to make the field as homogeneous as possible (within a specific volume around isocenter) when there is no object being imaged. These pieces of metal are set it place at the scanner installation, and are hardly ever changed. **Active shims** are a series of coils designed to create fields of a specific shape driven by their corresponding amplifiers. By changing the current going through those coils, you can correct (up to a point) for the inhomogeneities in the field created when you put an object in the scanner.

Running the scanner produces heat (large currents going through the gradient coils). If the cooling system cannot keep up, the temperature increase around the bore will change the susceptibility of the **passive shims**, but not of the **active shims** (if the scanner is designed properly, it should keep the current though the shim coils constant, even if the coil resistance changes; so the field they produce stays constant). The changes in susceptibility of the passive shims change the field they create. So, if you collect one image when the scanner is cold, and compare it to another image collected after running the scanner with a heavy duty cycle, there will be differences in the amount of distortions, even if the active shims are set to be the same. The amount of this difference in the distortion will be less if the scanner is designed properly (good cooling and passive shims that don't change susceptibility too much when heated; in that regard, the Prisma is a really good scanner, when compared to an old Allegra).

Now, the `ShimSetting` we are talking about in this PR refers to the **active shims**. This `ShimSetting` does not change during the acquisition (unless you are using customer sequences which changes the shims on the fly).

In the absence of heating effects (and subject motion), the field doesn't change if the active shims don't change. This is what allows you to measure the field in one run and use it to correct the distortions in a different image. However, if the active shims change between two runs, the field that you are measuring in one run will be different from the field producing distortions in a second image. That is why it is important that the `ShimSetting`s match.

You can measure the field either by calculating the phase difference in two images collected with different echo times (typically with a gradient echo sequence) or by estimating the distortions in the image with a `pepolar` pair of images. Either way, it is very important that the scanner doesn't change the (active) shims between the `fmap` sequence and the `func`/`dwi` sequence. Otherwise you will be measuring a different field.

> The most important parameter is 'Echo Spacing'. This parameter should be matched between distortion images and relevant image modality to impose the same epi distortions between them. Usually, 'pepolar' images are acquired using a Spin Echo Sequence (SE-EPI, without diffusion gradients). So, in total we use three different sequences if we acquire BOLD(GE-EPI) and DWI (SE-EPI-Diff).

Matching the distortions between the `pepolar` images and the BOLD/DWI images is just convenient, but not necessary. It is convenient because if the distortions are identical, it is straightforward to register the BOLD/DWI image and the `pepolar` image that has the same PE gradient polarity. However, this is not mandatory. As long as you make sure you are measuring the same magnetic field you want to correct for (i.e., you make sure the shims don't change), it doesn't matter if the parameters you use in the `pepolar` readout are different from those you use in the BOLD/DWI, as long as you keep track of those differences. This is why the fourth column in the `topup` [`--datain` file](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/topup/TopupUsersGuide#A--datain) asks for the total readout time: as long as you take into account the differences in the total readout time, you can have a totally different EPI readout in the BOLD/DWI compared to the `pepolar` images.

> Possible binding points between 'pepolar' images and BOLD (DWI) images are;
> 
> 1. Echo Spacing
> 2. Voxel size
> 3. Reciever Bandwitdh
> 4. Echo Train Length

Based on what is discussed above, the most important parameter to 'bind' `fmap` images (of any "case") and BOLD/DWI images are the `ShimSetting`, because it guarantees that the field you are measuring is the the same for both images.  Absent that, the other option you have is try to guess when the scanner changed the (active) shims or kept them constant.  For Siemens scanners, a pretty good guess is what @yarikoptic mentioned: if your imaging volume is identical, the scanner will normally not change the (active) shims. Identical imaging volumes means that the total "box" lengths are the same, and that the orientation is the same, module possible dimension swapping. However, this is just a guess: you can have images with different imaging volumes that have the same (active) shims (if you copied the "Adjust Volume"); and you can also have images with the same imaging volume, but with different active shims (because they had different "Shim Mode" settings, or because the user changed the (active) shims manually).  But, again, this is just a guess. I prefer having the user indicate somehow the intent in the `ProtocolName`, and propagate that (via the heuristic file) into the `_acq` entity in the BIDS filename.

(Apologies for the long comment.)

kodiweera:
Shimming I meant both active and passive collectively. "Due to heating of the gradient system and ferromagnetic shim elements, the shim and global frequency in the magnet may change during an imaging experiment." Weiscopf et al.  And shim-setting is not a sequence parameter. Even when we have active shimming constant across runs, what if 'echo spacing' is different for runs. we will have different EPI distortions among images which is undesirable!

pvelasco:
> Even when we have active shimming kept invariant (fixed) across runs, what if 'echo spacing' is different for runs. we will have different EPI distortions among images which are undesirable!

Why is it undesirable? As I explained in my comment above, you can still measure/estimate the field using a pair of `pepolar` images with an echo spacing and use it to correct for the distortions in an EPI (BOLD/DWI) that has a different echo spacing.  It is just more convenient if they are the same.

Also, what matters is the `TotalReadoutTime` (or, alternatively, the `EffectiveEchoSpacing`), not the echo spacing. You can have two images with the same echo spacing, but if in one of them you use iPAT 2, the distortions are cut in half.

kodiweera:
> > Even when we have active shimming kept invariant (fixed) across runs, what if 'echo spacing' is different for runs. we will have different EPI distortions among images which are undesirable!
> 
> Why is it undesirable? As I explained in my comment above, you can still measure/estimate the field using a pair of `pepolar` images with an echo spacing and use it to correct for the distortions in an EPI (BOLD/DWI) that has a different echo spacing. It is just more convenient if they are the same.
> 
> Also, what matters is the `TotalReadoutTime` (or, alternatively, the `EffectiveEchoSpacing`), not the echo spacing. You can have two images with the same echo spacing, but if in one of them you use iPAT 2, the distortions are cut in half.

Correct. 'Effective Echo Spacing' it is! Undesirable becasue producing the same amount of distortion is desirable in the case you don't know the 'totalreadout time'  so that you can use the same number (whatever the number is) for 'totalreadout time'. 

When I saw below comment in FSL, I felt like I was wasting time going after acuracy. I think that acuracy of these parameters are not that important for correction (I agree I don't know what the algorithm is). See below comment in FSL.

Look for this  line:
"Don't worry if you can't find the "total readout time" for your own data: you can almost always use 0.05."
somewhere in
https://fsl.fmrib.ox.ac.uk/fslcourse/lectures/practicals/fdt1/index.html

pvelasco:
> Undesirable becasue producing the same amount of distortion is desirable in the case you don't know the 'totalreadout time' so that you can use the same number (whatever the number is) for 'totalreadout time'.

I agree.
Before I started using `dcm2niix`, which returns the TotalReadoutTime in the json file (if present in the DICOM header), I just used the exact same EPI readout for both `pepolar` and BOLD/DWI, so when using `topup` I would always use a `1` in the 4th column.
However, now want to handle the use case of somebody collecting the images with different EPI readout parameters.

kodiweera:
> > Undesirable becasue producing the same amount of distortion is desirable in the case you don't know the 'totalreadout time' so that you can use the same number (whatever the number is) for 'totalreadout time'.
> 
> I agree.
> Before I started using `dcm2niix`, which returns the TotalReadoutTime in the json file (if present in the DICOM header), I just used the exact same EPI readout for both `pepolar` and BOLD/DWI, so when using `topup` I would always use a `1` in the 4th column.
> However, now want to handle the use case of somebody collecting the images with different EPI readout parameters.

So, can we issue warnings about mismatch of these binding points ( e.g. shimSettings and effective echo spacing / total readout time), and still do the BIDS conversion? One more point, if there are separate 'pepolar images' for BOLD and DWI, you need one more binding point other than 'shim setting' to distighish what belongs to what. TotalReadouttime/effective echo spacing is one option. Most probably voxel size also can be used.

pvelasco:
This PR is almost ready.
I have added two new options: one is what heudiconv should check for when matching `fmaps` to scans (`Shims` or `ImagingVolume`); the second one is which `fmap` to assign if more than one matches (`First` one or `Closest` in time).
@yarikoptic: what would be a good way to allow the user to specify those two parameters?  Currently, the user can request heudiconv to populate the `IntendedFor` field via `--command populate-intended-for`, but I don't know how to specify those two options.
Alternatively, we could make `--populate-intended-for` its own argument (options `True` or `False`), and have two other arguments: `--fmap-matches` (values `Shims` or `ImagingVolume`) and `--fmap-assignment` (values `First` or `Closest`). (Or some other argument names).
Any thoughts?

yarikoptic:
Just "to get it done" without major changes etc, the "easiest" (consistent with other similar cases but insufficiently documented/hidden) way could be to meld it into "options" any heuristic could specify.  

<details>
<summary>We already consult heuristics for some "settings":</summary> 

```shell
$> git grep 'getattr(heuri'
heudiconv/convert.py:                file_filter=getattr(heuristic, 'filter_files', None),
heudiconv/convert.py:                dcmfilter=getattr(heuristic, 'filter_dicom', None),
heudiconv/convert.py:                custom_grouping=getattr(heuristic, 'grouping', None))
heudiconv/convert.py:                scaninfo_suffix=getattr(heuristic, 'scaninfo_suffix', '.json'),
heudiconv/convert.py:                custom_callable=getattr(heuristic, 'custom_callable', None),
heudiconv/convert.py:                                    getattr(heuristic, 'DEFAULT_FIELDS', {}))
heudiconv/main.py:        heuristic_ls = getattr(heuristic, 'ls', None)
heudiconv/main.py:            populate_bids_templates(f, getattr(heuristic, 'DEFAULT_FIELDS', {}))
heudiconv/parser.py:            file_filter=getattr(heuristic, 'filter_files', None),
heudiconv/parser.py:            dcmfilter=getattr(heuristic, 'filter_dicom', None),
heudiconv/parser.py:            custom_grouping=getattr(heuristic, 'grouping', None)
heudiconv/parser.py:        if not getattr(heuristic, 'infotoids', None):
```
</details>

where of particular exemplar instance would be e.g. ` heuristic_ls = getattr(heuristic, 'ls', None)` for `ls` command, or `populate_bids_templates(f, getattr(heuristic, 'DEFAULT_FIELDS', {}))` for `populate-templates` command, see around https://github.com/nipy/heudiconv/blob/master/heudiconv/main.py#L82 .

This way any heuristic then could also specify the desired way for how to be introduced here functionality should behave.  May be to make it at least somewhat more "descriptive", we could reserve it to be `populate_intended_for_opts` dictionary which would provide those keyword args for the `populate_intended_for` function.

Specification from command line though would not be easily possible, besides quickly creating a "dummy" heuristic file to point to via `-f`.  WDYT? 

At some point, accounting for all lessons learned and use-cases currently implemented, someone would needs to RF heudiconv to provide more consistent and "descriptive" API and work out configuration (global, per heuristic; tunable in command line). At the moment though, I would prefer to not proliferate command line arguments with more of such command-line options.

heudiconv/utils.py
```diff
@@ -495,7 +521,11 @@ def create_tree(path, tree, archives_leading_dir=True):
             executable = False
             name = file_
         full_name = op.join(path, name)
-        if isinstance(load, (tuple, list, dict)):
+        if name.endswith('.json') and isinstance(load, dict):
+            # (For a json file, we expect the content to be a dictionary, so
+            #  don't continue creating a tree, but just write dict to file)
+            save_json(full_name, load, indent=4)
```
yarikoptic, line 527:
```suggestion
            save_json(full_name, load)
```
so we just stay consistent with our current/default behavior across files.

heudiconv/utils.py
```diff
@@ -265,6 +265,32 @@ def json_dumps_pretty(j, indent=2, sort_keys=True):
...
+                f=json_file,
+                k=key,
+                v=value,
+            )
+        )
+
+    with open(json_file) as f:
+        data = json.load(f)
+    data.update(field_and_value)
+    save_json(json_file, data, pretty=True)
```
yarikoptic, line 291: Unfortunately we aren't 100% consistent and use `pretty=True` only in  a few spots:
```shell
$> git grep -p pretty=True
heudiconv/bids.py=def populate_aggregated_jsons(path):
heudiconv/bids.py:        save_json(task_file, fields, sort_keys=True, pretty=True)
heudiconv/utils.py=def treat_infofile(filename):
heudiconv/utils.py:    save_json(filename, j_slim, sort_keys=True, pretty=True)

```
use of this function now needs to be aware of specifically to what "style" of json (indent and either pretty-fied) it conform, to ideally not change actual formatting of the file.  To at least allow to stay consistent, let's not hardcode `pretty` here and either make into

```suggestion
    save_json(json_file, data, pretty=pretty)
```
and add it as kwarg to signature (let's try to stay consistent with default indent and not allow to tune it unless will come needed)

heudiconv/utils.py
```diff
@@ -265,6 +265,32 @@ def json_dumps_pretty(j, indent=2, sort_keys=True):
     return js_
 
 
+def add_field_to_json(json_file, field_and_value):
```
yarikoptic, line 268:
```suggestion
def update_json(json_file, data):
```
or alike would be a better name:
- `.update` might not `add` but replace
- it is not just one but many (entire dict).

note: may be in the future we would like to extend it to use https://pypi.org/project/deepmerge/ ?

heudiconv/utils.py
```diff
@@ -265,6 +265,32 @@ def json_dumps_pretty(j, indent=2, sort_keys=True):
...
+    Parameters:
+    -----------
+    json_file : str or Path
+        path for the corresponding json file
+    field_and_value : dict
+        pair of "key": "value" to add to the json file
+    """
+    for key, value in field_and_value.items():
+        lgr.debug(
+            'File: "{f}": Adding {k}: {v}'.format(
```
yarikoptic, line 281:
```suggestion
            'File "{f}": Setting {k} to {v}'.format(
```
- since not necessarily adding, unless we do add some logic to at least warn or skip if record already exist (I have not  yet analyzed on where it is used)
- too many `:`s - would make it hard(er) to parse that log message

heudiconv/tests/test_convert.py
```diff
@@ -111,3 +114,75 @@ def test_b0dwi_for_fmap(tmpdir, capfd):
...
+    "subjects, sesID, expected_session_folder", [
+        (['Jason', 'Bourne'], None, 'sub-{sID}'),
+        (['Bourne'], 'Treadstone', op.join('sub-{{sID}}', 'ses-{{ses}}')),
+    ]
+)
+# Two possibilities: with or without heuristics:
+@pytest.mark.parametrize(
+    "heuristic", ['example', None]       # heuristics/example.py
+)
+def test_convert(tmpdir, monkeypatch, capfd,
```
yarikoptic, line 132:
```suggestion
def test_populate_intended_for(tmpdir, monkeypatch, capfd,
```
since the entire file is to test convert in various scenarios - test name should be more specific

heudiconv/tests/test_convert.py
```diff
@@ -111,3 +114,75 @@ def test_b0dwi_for_fmap(tmpdir, capfd):
...
+        """
+        print('session: {}'.format(session))
+        print('matching_parameter: {}'.format(matching_parameter))
+        print('criterion: {}'.format(criterion))
+        return
+    # mock the "populate_intended_for" attribute for the module from
+    # which "convert" was imported:
+    monkeypatch.setattr(
+        sys_modules[convert.__module__], "populate_intended_for", mock_populate_intended_for
+    )
```
yarikoptic, line 155: just `import heudiconv.convert` and  just `monkeypatch.setattr(heudiconv.convert, ...` or
`mock.patch('heudiconv.convert.populate_intended_for', new_callable=mock_populate_inteded_for)` . no need to drag sys_modules lookup into it IMHO

yarikoptic (changes requested):
Thank you!  Some comments from a partial review.  Hopefully we converge some time soon ;-)

heudiconv/tests/test_convert.py
pvelasco: I was just trying not to have to import the whole `heudiconv.convert` module, but rather only the functions being used, for consistency with other `tests/` files.
I followed your suggestion in 47bab3c8e07c818dec493b8dcfede767534c710a

heudiconv/utils.py
pvelasco: Done in 0650b68a2c6a936915ec96855993a935c103d4b4

heudiconv/utils.py
pvelasco: Addressed in c4a22240c4e5226a257ddc9248b3e6bf40e25280

heudiconv/bids.py
```diff
@@ -46,6 +50,49 @@ class BIDSError(Exception):
...
+        age = ('%.2f' if age != int(age) else '%d') % age
+    else:
+        age = age.rstrip('Y')
+    if age:
+        # strip all leading 0s but allow to scan a newborn (age 0Y)
+        age = '0' if not age.lstrip('0') else age.lstrip('0')
+        if age.startswith('.'):
+            # we had float point value, let's prepend 0
+            age = '0' + age
+    return age
```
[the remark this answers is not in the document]
pvelasco: Ooops.  Sorry about that.
Addressed in 267bf62fa87b0bb9f652f0a291998e118efcb314

yarikoptic:
I have pushed a tiny c4c758797f58da072d10bb0e9d51d4f63faca7b2 to trigger `pretty` to on -- was needed in my case, but now I wonder if I have not done a mistake with that: were sidecar json files you obtained before adding intended for not prettified?

pvelasco:
> [...] were sidecar json files you obtained before adding intended for not prettified?

I guess they were. Currently you only run `populate_intended_for` as a `--command` argument. That is, only after running through `heudiconv --bids` once.

yarikoptic (commented):
Merged master (with minor conflict resolving).

Why don't we make it run automagically also after completing conversion of a bids subject/session?

heudiconv/tests/test_convert.py
```diff
@@ -6,12 +6,9 @@
 import pytest
 from .utils import TESTS_DATA_PATH
 
-from heudiconv.convert import (update_complex_name,
-                               update_multiecho_name,
-                               update_uncombined_name,
-                               DW_IMAGE_IN_FMAP_FOLDER_WARNING,
-                               )
+import heudiconv.convert
```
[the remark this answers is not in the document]
pvelasco: > I guess we could revert back later on -- I see no reason for removing the specific functions import and making it necessary to always call with `heudiconv.convert` here.

That would be great.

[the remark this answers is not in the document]
pvelasco: Fixed in ae61382c021031c000d87e9c8ea27ebd435e165d

pvelasco:
> Why don't we make it run automagically also after completing conversion of a bids subject/session?

That would be great

yarikoptic:
actually -- I think it already does that (populates at convert) at here: https://github.com/nipy/heudiconv/pull/482/files#diff-d19d089741b3844c199c8236ec1c0fcd43400b2ce53cd587d203255add863c39R527  ... trying it out or some sample cases locally.

yarikoptic:
I guess I have misread the code since it didn't populate them although whenever I post-ran command manually it did... 
FTR, my attempt was on https://datasets.datalad.org/?dir=/dbic/QA sub emmet (lego phantom) session 20180531 via 
```
OUTDIR=compare-versions-emmet-20180531 heudiconv/utils/test-compare-two-versions.sh heudiconv{-master,} --bids -f reproin --datalad --files $PWD/dbic-QA/sourcedata/sub-emmet/ses-20180531
```

pvelasco (commented):
> I guess I have misread the code since it didn't populate them

Thanks for finding the bug. It is fixed now (723fa67673a6b504da9a3b2a0b2e09814615305f)

docs/heuristics.rst
```diff
@@ -84,4 +84,54 @@ or::
...
+.. note::
+  Example::
+
+    POPULATE_INTENDED_FOR_OPTS = {
+            'matching_parameter': 'ImagingVolume',
+            'criterion': 'Closest'
+    }
+
+If ``POPULATE_INTENDED_FOR_OPTS`` is not present in the heuristic file, ``IntendedFor``
+will not be populated automatically.
```
yarikoptic, line 137: hm,  I think it is not the case currently -- I do not spot any condition at  https://github.com/nipy/heudiconv/pull/482/files#diff-d19d089741b3844c199c8236ec1c0fcd43400b2ce53cd587d203255add863c39R546 or around, and when I tried now with v0.9.0-97-g723fa67 an `-f reproin` it happily populated it.

I am ok either way.  But if to demand it to be defined, make check for it to be None and not just empty dict, so empty dict would instruct "enable but use default opts".

Also I wondered if we should make `matching_parameter` into `matching_parameters` and default to have both zooms and sizes to be matched.  This way we have a better chance of doing "the right thing". Also it would allow later to expand with additional ones we have discussed at some point.

pvelasco: > I am ok either way. But if to demand it to be defined, make check for it to be None and not just empty dict, so empty dict would instruct "enable but use default opts".

It would be [here](https://github.com/cbinyu/heudiconv/blob/723fa67673a6b504da9a3b2a0b2e09814615305f/heudiconv/convert.py#L213) right?

pvelasco: > Also I wondered if we should make `matching_parameter` into `matching_parameters` and default to have both zooms and sizes to be matched. This way we have a better chance of doing "the right thing". Also it would allow later to expand with additional ones we have discussed at some point.

You mean, have `matching_parameters` to be a list and have all elements in the list to match?
Sure, I can do that.

yarikoptic: > > I am ok either way. But if to demand it to be defined, make check for it to be None and not just empty dict, so empty dict would instruct "enable but use default opts".
> 
> It would be [here](https://github.com/cbinyu/heudiconv/blob/723fa67673a6b504da9a3b2a0b2e09814615305f/heudiconv/convert.py#L213) right?

yes, there it would default to `None`, not `{}` and then somewhere [there in convert](https://github.com/cbinyu/heudiconv/blob/723fa67673a6b504da9a3b2a0b2e09814615305f/heudiconv/convert.py#L527) would do `if populate_intended_for_opts is not None:` for the subsequent block of code which deals with triggering `populate_intended_for`

yarikoptic: > > Also I wondered if we should make `matching_parameter` into `matching_parameters` and default to have both zooms and sizes to be matched. This way we have a better chance of doing "the right thing". Also it would allow later to expand with additional ones we have discussed at some point.
> 
> You mean, have `matching_parameters` to be a list and have all elements in the list to match?
> Sure, I can do that.

yes.  We better do it now so later we do not need to deal with handling both `matching_parameter` and `matching_parameters`.

PS I am so looking forward to have this PR merged! ;) Thank you for all the work and bearing with all my comments/wishes

pvelasco: OK. I think both issues have been addressed in https://github.com/nipy/heudiconv/pull/482/commits/5a06034e4ed0e2e52ca8e446b1384bf92aa40e60 and https://github.com/nipy/heudiconv/pull/482/commits/ccb96dfdce52a6376a42d47b07951882570d044a.
I also [made `ImagingVolume` the default](https://github.com/nipy/heudiconv/pull/482/commits/b72ecd8f5b701a3898dc67a6930d6fe499275998) for `populate_intended_for` (that means that the position, orientation, resolution and numbers of voxels should match).

HippocampusGirl:
Thank you for working on this very useful feature! I was wondering how you decided on the `matching_parameter` to use. Is it always necessary to match all acquisition parameters including `ShimSetting`? Are there cases where the voxel size and image dimensions might be different between the image and the fieldmap, and the field map can still be used? In other words, are there cases where the distortions within a session will be more or less the same across (slightly) different sequences?

pvelasco:
> Thank you for working on this very useful feature! I was wondering how you decided on the `matching_parameter` to use. Is it always necessary to match all acquisition parameters including `ShimSetting`?

Theoretically, you "only" need to match the `ShimSetting`. I say "only" because even if the other parameters differ, if the scanner doesn't change the field, then the distortions introduced by the head stay the same. (True, there will be some changes in the field that the scanner creates e.g. because of heating, but you can argue these are minor effects, at least in modern scanners.)
And also "only" because if you keep the other parameters the same, if the scanner changed the `ShimSetting` (e.g., in a Siemens scanner you change from `Normal` to `Tune Up`), then the field that the scanner creates changes, and what you measured in the `fmap` doesn't match the field distorting the `func` and `dwi` images.

If the `ShimSetting` is not available (either because your DICOM files didn't have it in the header to begin with, or because when you convert them to BIDS that info was lost), at least for Siemens scanners it is relatively safe to assume that the scanner didn't shim if you kept the same "Adjust Volume", which normally is the same as the imaging volume (that is, the image dimensions as well as the position and orientation). So that is why I give the option to chose `matching_parameters = ['ImagingVolume']`.

> Are there cases where the voxel size and image dimensions might be different between the image and the fieldmap, and the field map can still be used? In other words, are there cases where the distortions within a session will be more or less the same across (slightly) different sequences?

I don't know. I haven't looked carefully into that. I know that (in Siemens scanners) if you change the box and the scanner re-shims, the field you measure is different. However, I don't have a sense of how different it might be, meaning if it would be better not to apply any correction in that case.
Here you can see [what Jesper Andersson, from the FSL team, has to say about it](https://www.jiscmail.ac.uk/cgi-bin/wa-jisc.exe?A2=ind2104&L=FSL&P=R78395). He has a lot more experience than I do.

So maybe I should add another matching parameter to the allowed ones (maybe called `'force'`), which will force `heudiconv` to consider any field map in the session to be `'IntendedFor'` (of course, if there are several fmaps, it will use the one depended on the `criterion` you chose: `'First'` or `'Closest`')

heudiconv/bids.py
```diff
@@ -487,3 +509,409 @@ def convert_sid_bids(subject_id):
...
+            "Fmap matching_parameter %s not allowed." % matching_parameter
+        )
+
+    # loop through the possible criteria and extract the info needed
+    if matching_parameter == 'Shims':
+        key_info = [get_shim_setting(json_file)]
+    elif matching_parameter == 'ImagingVolume':
+        from nibabel import load as nb_load
+        nifti_file = glob(remove_suffix(json_file, '.json') + '.nii*')
+        nifti_header = nb_load(nifti_file).header
```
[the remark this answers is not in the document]
pvelasco: Fixed with your PR (cbinyu#10)

heudiconv/bids.py
```diff
@@ -487,3 +509,409 @@ def convert_sid_bids(subject_id):
...
+    )
+    fmap_groups = OrderedDict()
+    for k in prefixes:
+        fmap_groups[k] = [
+            fm for fm in fmap_jsons if fmap_regex.sub('', remove_suffix(op.basename(fm), '.json')) == k
+        ]
+    return fmap_groups
+
+
+def get_key_info_for_fmap_assignment(json_file, matching_parameter='ImagingVolume'):
```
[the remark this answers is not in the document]
pvelasco: Addressed in 18678984b1ae07c9d34cec6e73832e050823b541.

heudiconv/main.py
```diff
@@ -108,6 +108,12 @@ def process_extra_commands(outdir, command, files, dicom_dir_template,
         ensure_heuristic_arg(heuristic)
         from .utils import get_heuristic_description
         print(get_heuristic_description(heuristic, full=True))
+    elif command == 'populate-intended-for':
+        for subj in subjs:
```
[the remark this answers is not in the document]
pvelasco: I guess we could either read the `participants.tsv` or do a `glob` for `sub-*` in the `outdir`.
Which is preferred?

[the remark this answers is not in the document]
yarikoptic: since we would need to glob sessions anyways, ideally
- glob sub-*
- read from participants.tsv and if differs -- issue a warning
- glob session per each sub

heudiconv/bids.py
```diff
@@ -487,3 +510,505 @@ def convert_sid_bids(subject_id):
...
+    Returns:
+    -------
+    str with "ShimSetting" value
+    """
+    data = load_json(json_file)
+    try:
+        shims = data[SHIM_KEY]
+    except KeyError as e:
+        lgr.error('File %s does not have "ShimSetting".'
+                  'Please use a different "matching_parameters" in your heuristic file', json_file)
```
yarikoptic, line 533:
```suggestion
        lgr.error('File %s does not have "%s". '
                  'Please use a different "matching_parameters" in your heuristic file',
                   json_file, SHIM_KEY)
```
to not hardcode what is stored in SHIM_KEY

heudiconv/bids.py
```diff
@@ -487,3 +510,505 @@ def convert_sid_bids(subject_id):
...
+        sessions).
+    matching_parameters : list of str from AllowedFmapParameterMatching
+        matching_parameters that will be used to match runs
+
+    Returns:
+    -------
+    compatible_fmap : dict
+        Dict of compatible_fmaps_groups (values) for each non-fmap run (keys)
+    """
+    if type(matching_parameters) is not list:
```
yarikoptic, line 714:
```suggestion
    if isinstance(matching_parameters, list):
        assert isinstance(matching_parameters, str), "matching_parameters must be a str or a list, got %s" % matching_parameters
```
note: didn't check if would work as expected

heudiconv/bids.py
```diff
@@ -487,3 +510,505 @@ def convert_sid_bids(subject_id):
...
+        the other image
+    """
+    if not op.exists(json_file):
+        raise FileNotFoundError(
+            errno.ENOENT, os.strerror(errno.ENOENT), json_file
+        )
+    if matching_parameter not in AllowedFmapParameterMatching:
+        raise ValueError(
+            "Fmap matching_parameter %s not allowed." % matching_parameter
+        )
```
yarikoptic, line 608: there is 4 identical spots verifying value to be the one expected.  Should probably be done once only at the point of where that value is coming in (heuristic?)

yarikoptic:
@pvelasco sorry for not being as active as I should have ;)  could you please let me know if you don't have any pending changes to be pushed and not working on it ATM?  I would like to get a (hopefully) through pass through it done and send you a PR for review (or if you give blessing -- push here directly)

heudiconv/bids.py
pvelasco: Shouldn't it be `if not isinstance(matching_parameters, list):`? (Note the `not`)

heudiconv/bids.py
pvelasco: Sure, I can do it only once. However, I wanted to make each function self-sufficient, so that it could be re-used if needed.  (Disclaimer: I have no Computer Science/Programming training, so I really don't know what the convention is for these things...:blush: )

pvelasco:
Sorry for the delay, @yarikoptic. I'm back to work on this PR.
I'm working on your suggestion from Jun 11 to auto-populate subjects and sessions in `process_extra_commands`.  It should be done in a day or two.

heudiconv/bids.py
yarikoptic: right, sorry for a buggy suggestion -- fixed it now!

heudiconv/bids.py
yarikoptic: we all learn as we go/do more coding ;)  typically if there is repeated / copy-pasted code, it is a sign for either needing to take that code into a function for reuse, or as in this case AFAIK, move checking somewhere closer to where the value could "enter" (be specified by the user) the code path.

heudiconv/main.py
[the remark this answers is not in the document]
pvelasco: Addressed in 51d1a96d45fd599ab48eb1ed060ec65688146c37

heudiconv/bids.py
pvelasco: Addressed in d53e50b7b9ddbbeded6765c353fd12bf882664fb

pvelasco:
Hi @yarikoptic,
I don't have any more changes in the pipeline ATM, so if you want to make any changes, go ahead.

burdinskid13:
Thanks so much for adding this to heudiconv, this is super helpful! :) Is it ready to be used or are there still checks/edits that need to be made?

yarikoptic:
it is still wip but would be great if you give it a shot @burdinskid13 and let us know if it worked out nicely for your usecase

Giuseppe1988:
Dear all,
sorry for the very silly question.

I am trying to use the 'populate-intended-for' command but I get the following error:

"populate_intended_for() missing 2 required positional arguments: 'matching_parameters' and 'criterion'".

While I understand the meaning of the additional arguments and their possible values, It is not clear to me how to pass them to the 'populate-intended-for' command.

Could you please give me a hint?

Thank you so much for your work!

pvelasco:
Hi @Giuseppe1988,
Since this PR is not yet merged, the explanation of the usage is not yet in the official `heudiconv` documentation. You can find it [here](https://github.com/cbinyu/heudiconv/blob/adds_populate_intended_for/docs/heuristics.rst#populate_intended_for_opts).
Please give it a try and let us know if it works.

Giuseppe1988:
Dear @pvelasco 
it worked, I used the 'AcquisitionLabel' value as matching parameter and it went just fine, thank you so much!

If I may ask, I'd have two related questions:
- I see that one of the commits creates an option to add a field to the json file, but I'm not sure of how and where this option should be specified. I am trying to add the "PhaseEncodingDirection" field to the dwi json (which is not automatically created in the conversion from Philips scanners).
- I would need to run dcm2niix with the "-z n -m y -i y" flags. I created a json file to pass to the --dcmconfig option but I'm not sure of how the dcm2niix flags should be specified in it.

Am I missing something?

neurorepro:
Hello @pvelasco and @yarikoptic , this is a great PR !

I had some issues when using it in my case. The main problems were:
* the `ImagingVolume` parameter did not lead to a proper comparison between fmap and modality json info
* the `AcquisitionLabel` did not allow for selecting custom acquisition label, but only modality. This does not work in my use case as i have several field maps for several different fmri tasks (i.e. all "func" modality)

I made a PR to address the first issue [here](https://github.com/cbinyu/heudiconv/pull/12), and i already have code to address the second issue but it'd be easier if the first PR could be validated before i submit the second PR.

What do you guys think ?

neurorepro:
@pvelasco  and @yarikoptic  I have now made [this PR](https://github.com/cbinyu/heudiconv/pull/13) to address custom labels, which includes updated doc and test suite.

neurorepro:
@pvelasco  and @yarikoptic , how can i help finalize this PR so that it can be merged into the main branch ?

heudiconv/bids.py
```diff
@@ -496,3 +520,507 @@ def convert_sid_bids(subject_id):
...
+                    op.relpath(remove_suffix(json_file, '.json') + '.nii.gz', start=subj_folder)
+                )
+        if intended_for:
+            intended_for = sorted([str(f) for f in intended_for])
+            # Add this intended_for to all fmap files in the fmap_group:
+            for fm_json in unique_fmap_groups[fmap_group]:
+                update_json(fm_json, {"IntendedFor": intended_for}, pretty=True)
+
+
+class BIDSFile(object):
```
yarikoptic, line 938: woohoo -- here it is also!  ideally #544 should be finished/merged first (so would be driven by a specific copy of the schema etc), but doesn't need to be.  If this PR is merged first, that one should be rebased/code moved/RFed and would immediately take advantage of this use-case use

nikhil153:
Very useful addition @pvelasco, @neurorepro, and @yarikoptic. I am doing some local tests with the main branch, and seeing this warning when I run with "--command populate-intended-for" option. 
```
INFO: Running heudiconv version 0.10.0 latest 0.10.0
INFO: Adding "IntendedFor" to the fieldmaps in /scratch/bids_test/sub-PD01759D757223/ses-01.
WARNING: We cannot add the IntendedFor field: no fmap/ in /scratch/bids_test/sub-PD01759D757223/ses-01  
```
Interestingly, if I remove  "--command populate-intended-for" option, as long as I have POPULATE_INTENDED_FOR_OPTS in my heuristics, the fmap jsons get populated correctly with "IntendedFor" field!

pvelasco:
Hi @nikhil153,

The warning you receive when running with the `--command populate-intended-for` option means that there are no `fmaps` in that session folder. Did you run `heudiconv` (without the `--command` option) first?
The intended use, AFAIK, is to run it first without any `--command`, and then run the `--command` on the initially processed data.
From the help (note the `instead`):
``` (shell)
  --command {heuristics,heuristic-info,ls,populate-templates,sanitize-jsons,treat-jsons,populate-intended-for}
                        Custom action to be performed on provided files
                        instead of regular operation.
```

WangYunHong98:
Hi @pvelasco

I meet the same issue that ```ShimSetting``` in ```_phasediff.json``` is slightly different from ```ShimSetting``` in ```_bold.json```.

For example, 
in ```_bold.json```, the ```ShimSetting``` is  ```[2957, 4394, -6624, 354, -7, -55, -115, -26]```
in ```_phasediff.json```, the ```ShimSetting``` is ```[2965, 4393, -6624, 368, 13, -63, -91, -25]```

So this slight difference will have an effect on ```func``` when it does SDC?

This is also causing heudiconv errors when adding ```IntendedFor``` to ```_phasediff.json``` file.

Best, 
Yunhong

pvelasco:
Hi @WangYunHong98,

Lately I switched jobs and now I'm working on different projects. I haven't been able to work on this project for about two years, but I think the way to make the parameter comparison less restrictive would be to add a relative tolerance in the call to `np.allclose` [here](https://github.com/nipy/heudiconv/blob/7caff37cf040d0ba12967b41b0385548ce6a51bd/heudiconv/bids.py#L817).

That said, first principles say that if the `ShimSettings` for the acquisition you want to correct are not the same as for the field/distortion maps, the correction you are applying by using the field/distortion map will not be correct, since the B0 field affecting the distorted image is not the same that you measured with the field/distortion maps. (The exception being the first entry, which is a frequency offset and will result in an image shift in the readout direction, which can be corrected when you do motion correction.)

Some people smarter than me argue that using an approximate field correction is better than no field correction but, afaik, there have been no publication examining whether that's true or not, or what can be considered "slightly different parameters". So I would argue that a very good solution would be to have a configuration parameter somewhere that allows you to specify how restrictive you want to be, and let the user specify it.

Right now, I don't have the bandwidth to make those changes. Do you feel up to the task? 😉

PR #485, opened 2020-12-22
BF: split prefix into path (if any) and provide only basename as prefix for dcm2niix invocation

### LINKED ISSUE #481: Question: Why is dcm2niix saving to /tmp
### Summary

In our logs I can see that `dcm2niix` is saving to a random folder in `/tmp`. For instance I get messages like this one:
```
INFO: [Node] Running "convert" ("nipype.interfaces.dcm2nii.Dcm2niix"), a CommandLine Interface with command:
dcm2niix -b y -z y -x n -t n -m n -f func -o . -s n -v n /tmp/dcm2niix0odmo4o6/convert
```
If we isolate our containers from host, i.e, `--contain` flag in singularity, then the container `/tmp` may run out of memory to store all the nifti files. While we have ways to work around this, it took a while to figure out what was going on. Our `nii.gz` where just showing missing blocks. I think it is unexpected that heudiconv would be using  a hardcoded `/tmp` to temporarily store conversion.  Is there a way to change the path or make that assumption more clear to the user? 

### Platform details:

Choose one:
- [ ] Local environment

- [x] Container

This is a custom container 
https://hub.docker.com/repository/docker/brownbnc/xnat-tools
All versions contain the same version of heudiconv and dcm2nii
While we create a docker container we use it mostly as a singularity container from our HPC

- Heudiconv version: 

- heudiconv —version 0.6.0
- dcm2niiX version v1.0.20190902

yarikoptic:
dcm2niix started to segfault whenever -f is too long https://github.com/rordenlab/dcm2niix/issues/465 .
And it feels that we abused -f wherever -o should have been used.
I have not looked though through history of changes if such use of -f
was warranted, but as I see that output_dir is not explicitly defined,
decided to just use that one instead.

Initial commit  resolved the dcm2niix segfault during heudiconv/tests/test_heuristics.py::test_scans_keys_reproin
when testing on my laptop, BUT resulted in test failures since now placing the files in the output directory copying to destination was not working since it was destination.  I added conditional in `safe_copyfile` (0f6abbf) but that was not sufficient and also made me wonder -- if placing already at the destination (or near) why don't we move instead of copying?  That would also resolve #481 and make heudiconv more robust IMHO in cases where `/tmp` has smaller capacity than the destination.  

The only negative side-effect might be that the target directory, in case of abnormal heudiconv operation, would have those `_heudiXXX` files.  I could present it as a feature which would ease troubleshooting etc ;)

So I followed up with 8b4b7a0 where I have also added a random suffix to the target suffix given to nipype's dcm2niix call.  **This is the most radical change** in this PR which might introduce some bugs if we are not moving/renaming files properly.  But looking at the code we seems to be doing all needed, and then

```shell
$> find pytest-of-yoh -iname *_heudi*
pytest-of-yoh/pytest-0/test_embed_dicom_and_nifti_met0/nifti_heudiconv222.nii.gz
```
as in that explicit test we do not bother renaming etc, whenever all the other files in there seems to look good.  So I have good confindence that we should be ok.

then while at it I followed up with 7ad241c17967f18f33edb4c9109e25be3b87238e which should address #462 .

TODO
- [x] ensure passes existing tests
- [x] add to description since it grew bigger/more radical
- [x] seek feedback (due to dcm2niix change in behavior/limits for `-f` we need it fast though https://github.com/rordenlab/dcm2niix/issues/465 so I will be merging/releasing as soon as I am satisfied)

edit: additional benefit
- by placing into target directory right away we would not loose some files nipype or us forgot to move (like we had with DWI .bvec) -- they just will have an odd suffix.  I like it

### DIFF
--- a/heudiconv/utils.py
+++ b/heudiconv/utils.py
@@ -358,6 +358,9 @@ def safe_copyfile(src, dest, overwrite=False):
     """
     if op.isdir(dest):
         dest = op.join(dest, op.basename(src))
+    if op.realpath(src) == op.realpath(dest):
+        lgr.debug("Source %s = destination %s", src, dest)
+        return
     if op.lexists(dest):
         if not overwrite:
             raise RuntimeError(
--- a/heudiconv/utils.py
+++ b/heudiconv/utils.py
@@ -379,7 +379,7 @@ def _safe_op_file(src, dest, operation, overwrite=False):
         if not overwrite:
             raise RuntimeError(
                 "was asked to %s %s but destination already exists: %s"
-                % (src, operation, dest)
+                % (operation, src, dest)
             )
         os.unlink(dest)
     getattr(shutil, operation)(src, dest)
--- a/heudiconv/utils.py
+++ b/heudiconv/utils.py
@@ -354,7 +354,21 @@ def get_known_heuristics_with_descriptions():
 
 
 def safe_copyfile(src, dest, overwrite=False):
-    """Copy file but blow if destination name already exists
+    """Copy file but blow if destination name already exists"""
+    return _safe_op_file(src, dest, "copyfile", overwrite=overwrite)
+
+
+def safe_movefile(src, dest, overwrite=False):
+    """Move file but blow if destination name already exists"""
+    return _safe_op_file(src, dest, "move", overwrite=overwrite)
+
+
+def _safe_op_file(src, dest, operation, overwrite=False):
+    """Copy or move file but blow if destination name already exists
+
+    Parameters
+    ----------
+    operation: str, {copyfile, move}
     """
     if op.isdir(dest):
         dest = op.join(dest, op.basename(src))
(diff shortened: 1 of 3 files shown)


yarikoptic:
Dear @mgxd @tsalo @satra and @nipy/team-heudiconv  -- you speedy feedback on this change would be appreciated.  We need to workaround a regression (IMHO) in dcm2niix which could lead to segfaults with current version.  See description of PR and individual commits for more information.  If no objections are voiced within a day, I would aim to release tomorrow (sorry for a short notice etc, but want to get this one out) and will bear the responsibility for fixing what I broke (if anything)

heudiconv/utils.py
```diff
@@ -354,18 +354,35 @@ def get_known_heuristics_with_descriptions():
...
+    if op.realpath(src) == op.realpath(dest):
+        lgr.debug("Source %s = destination %s", src, dest)
+        return
     if op.lexists(dest):
         if not overwrite:
             raise RuntimeError(
-                "was asked to copy %s but destination already exists: %s"
-                % (src, dest)
+                "was asked to %s %s but destination already exists: %s"
+                % (src, operation, dest)
```
yarikoptic, line 382: doh -- spotted that those should be reversed -- will fixup now.

yarikoptic:
FWIW, wherever dependencies allowed, builds went find on neurodebians

PR #491, opened 2021-02-01
BF: Ensure bvec/bval files are only created for dwi output

pvelasco:
Currently, `.bvec`/`.bval` files are copied into the output folder whenever they are present in the `dcm2niix` output. `dcm2niix` (v1.0.20201102) creates those files whenever it detects a diffusion sequence and the images are not derived (https://github.com/rordenlab/dcm2niix/issues/352#issuecomment-548841549). So, with the current `heudiconv`, if you extract an image collected with a diffusion sequence, but with b-value = 0 as a `fmap` ([`pepolar` modality](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#case-4-multiple-phase-encoded-directions-pepolar)), it will copy the `.bvec`/`.bval` files to the `fmap` folder.  However, this doesn't conform to BIDS format, which only allows for `.bvec`/`.bval` for data in the [`dwi` folder](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#diffusion-imaging-data). Consequently, BIDS-validator (v1.5.10) gives an error:
```
bids-validator@1.5.10

	1: [ERR] Files with such naming scheme are not part of BIDS specification. This error is most commonly caused by typos in file names that make them not BIDS compatible. Please consult the specification and make sure your files are named correctly. If this is not a file naming issue (for example when including files not yet covered by the BIDS specification) you should include a ".bidsignore" file in your dataset (see https://github.com/bids-standard/bids-validator#bidsignore for details). Please note that derived (processed) data should be placed in /derivatives folder and source data (such as DICOMS or behavioural logs in proprietary formats) should be placed in the /sourcedata folder. (code: 1 - NOT_INCLUDED)
		./sub-FuncDiff/fmap/sub-FuncDiff_acq-dwi_dir-AP_epi.bval
			Evidence: sub-FuncDiff_acq-dwi_dir-AP_epi.bval
		./sub-FuncDiff/fmap/sub-FuncDiff_acq-dwi_dir-AP_epi.bvec
			Evidence: sub-FuncDiff_acq-dwi_dir-AP_epi.bvec
		./sub-FuncDiff/fmap/sub-FuncDiff_acq-dwi_dir-PA_epi.bval
			Evidence: sub-FuncDiff_acq-dwi_dir-PA_epi.bval
		./sub-FuncDiff/fmap/sub-FuncDiff_acq-dwi_dir-PA_epi.bvec
			Evidence: sub-FuncDiff_acq-dwi_dir-PA_epi.bvec
```

This PR makes sure we only copy the `.bvec`/`.bval` if the destination folder for an image is `dwi`. It also adds a unit test, and the necessary dicom file and heuristics.


pvelasco:
It fails CI because the version it installs from Neurodebian is still `1.0.20181125`, while the new behavior seemed to [have been introduced](https://github.com/rordenlab/dcm2niix/commit/2a07b9b646812ec4d110dad34ecbcc9f54e3ebd0) in version `1.0.20201102`.
I'll have to investigate some more.

heudiconv/convert.py
```diff
@@ -674,9 +674,14 @@ def save_converted_files(res, item_dicoms, bids_options, outtype, prefix, outnam
...
-        safe_movefile(res.outputs.bvecs, outname_bvecs, overwrite)
-        safe_movefile(res.outputs.bvals, outname_bvals, overwrite)
+        if prefix_dirname.endswith('dwi'):
+            outname_bvecs, outname_bvals = prefix + '.bvec', prefix + '.bval'
+            safe_movefile(res.outputs.bvecs, outname_bvecs, overwrite)
+            safe_movefile(res.outputs.bvals, outname_bvals, overwrite)
+        else:
+            os.remove(res.outputs.bvecs)
+            os.remove(res.outputs.bvals)
+            lgr.debug("%s and %s were removmed", res.outputs.bvecs, res.outputs.bvals)
```
yarikoptic, line 684:
```suggestion
            lgr.debug("%s and %s were removed since not dwi", res.outputs.bvecs, res.outputs.bvals)
```

yarikoptic (changes requested):
I think it would be better to be safe than sorry and check if bval is really 0. If not - issue a warning and not remove.  That should also then alert bids validator, and user would need to investigate why they have those files (and why it wasn't 0 bval)

pvelasco:
> I think it would be better to be safe than sorry and check if bval is really 0. If not - issue a warning and not remove.

That sounds sensible. I will work on it.

pvelasco:
@yarikoptic,
I think 9746fe61670ba3d679e65ec1229a7c33ec560e52 addresses your suggestion. I have also modified the unit test accordingly.

heudiconv/convert.py
```diff
@@ -806,3 +820,23 @@ def  add_taskname_to_infofile(infofiles):
...
+
+
+def bvals_are_zero(bval_file, threshold=5):
+    """Checks if all entries in a bvals file are zero (or below the threshold).
+    Returns True if that is the case, otherwise returns False
+
+    Parameters
+    ----------
+    bval_file : file with the bvals
+    threshold : b-value below which they are considered to be zero. (Default: 5)
```
yarikoptic, line 832: posted against some original commit : do you have immediate usecases where they are <5 but not 0 and should still be considered?

pvelasco:
> you saw usecases with non absolute 0 bvals where they still should be considered? just curious

No. I would not use a diffusion-weighted image (b non zero) for `fmap` estimation purposes, because you will always get lower SNR than the equivalent b-value = 0 image.
Note: Siemens uses all the gradients in the sequence to compute the real b-value. So even if you select b-value=0, you get a value of 5 in the header.

heudiconv/convert.py
pvelasco: No, not `<5`, but yes `=5`.
As I mentioned, for Siemens scanners with relatively up-to-date software versions (I have tested it for VE11C), when you select a b-value of 0, it ends up writing `5` in the header (it takes into account slice-encoding gradients and readout gradients).

**Usecase:**
(Tested in Siemens scanners): If you want to collect a pair of Spin-Echo (SE) EPI images (`pepolar`) to estimate the `fmap`, the best way to ensure that all the readout parameters of the DW images you are going to correct are matched is by making a copy of the DW sequence, set the number of b-values to 1, and set this b-value to zero. That way you have the same FoV, matrix size, acceleration factor, partial-Fourier and echo spacing. (Then, I set the TE to the minimum allowed by the sequence, to get higher SNR, but that is not necessary.) Because the b-value is zero, there are no diffusion-weighting gradients, and the resulting image is a SE image.
The resulting images have `'DIFFUSION'` in the `image_type` list, and a b-value of 5, and need to be extracted as `fmap`.
I have found cases that, for some reason or other, if I started with a standard SE sequence and tried to modify the sequence parameters to match those of the diffusion-weighted sequence, I haven't been able to. I don't remember exactly when it happens, but it could have been that I hadn't been able to match the echo-spacing, or the matrix size, or something else. But if you just start with a copy of the DW sequence, you are guaranteed to have same readout.

yarikoptic: hm, should we then make it like 
```python
bvals_unique = set(float(b) for b in bvals)
return bvals_unique == {0.} or bvals_unique == {5.}
```
to be "exact" and not silently succumb to some other odd cases?

pvelasco: Sure.

pvelasco: Resolved in 5099d01a086af776f454c92446331428b69d67e6

yarikoptic:
Great, thank you @pvelasco !

PR #494, opened 2021-02-06
BF: datalad - use public call_git (added in 0.12.0), not deprected not protected method

yarikoptic:
detected while test building debian pkg


pvelasco:
I think this is a duplicate of #492

yarikoptic:
Indeed! Ufff, need to pay better attention ;-) thank you!

PR #495, opened 2021-02-12
BF: Make sure participant sex is not left empty in 'participants.tsv'

pvelasco:
If empty string, replace it with `N/A`.

### DIFF
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -282,7 +282,7 @@ def add_participant_record(studydir, subject, age, sex):
         f.write(
           '\t'.join(map(str, [participant_id,
                               age.lstrip('0').rstrip('Y') if age else 'N/A',
-                              sex,
+                              sex if sex else 'N/A',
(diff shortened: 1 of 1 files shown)


heudiconv/bids.py
```diff
@@ -282,7 +282,7 @@ def add_participant_record(studydir, subject, age, sex):
         f.write(
           '\t'.join(map(str, [participant_id,
                               age.lstrip('0').rstrip('Y') if age else 'N/A',
-                              sex,
+                              sex if sex else 'N/A',
```
yarikoptic, line 285: bids spec seems to like it lower case `n/a` thus
```suggestion
                              sex if sex else 'n/a',
```

pvelasco:
I was just trying to mimic the previous line. Upper case `N/A` passes the validator.
Still, I agree with you that we should follow [the spec](https://bids-specification.readthedocs.io/en/v1.4.1/03-modality-agnostic-files.html#participants-file).

PR #496, opened 2021-02-19
ENH+BF: find_files to take list of topdirs, look for _bold only under  sub-* directories

yarikoptic:
Originally issue reported on
https://neurostars.org/t/heudicov-permission-denied-bids-derivatives-mriqc/16543
where heudiconv would crash while trying to create an _events.tsv file under derivatives/mriqc.

heudiconv should not care to look anywhere besides immediate subject directories, even if people
place derivatives direcly in the dataset into which they are also converting their data into
(making YODA unhappy).

TODOs
- [ ] verify on a sample case that there is no side effect (not sure if I would be up for composing a test)


yarikoptic:
AFAIK should be ok to go

PR #501, opened 2021-03-31
BF: Don't fail if etelemetry.get_project returns None

kyleam:
main.workflow() crashes with a TypeError when logging the latest
version if "NO_ET" is in os.environ because etelemetry.get_project()
returns None.  Guard against this failure as well the KeyError that
would be triggered if get_project(), for whatever reason, returns a
dictionary without a "version" key.


yarikoptic:
THANK YOU!

PR #507, opened 2021-04-30
RF/ENH: populate a singular scans.json at the top level of the bids dataset

yarikoptic:
instead of exactly the same .json alongside every `_scans.tsv`.  See the 2nd commit (https://github.com/nipy/heudiconv/commit/864da69a3b1acc875da2429290984126dad34b18) for more motivation etc.

While at it also RF'ed `_canonical_dump` into a tiny `json_dumps`

causes only minor merge conflict for #482

### DIFF
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -18,7 +18,7 @@ from .utils import (
     load_json,
     save_json,
     create_file_if_missing,
-    json_dumps_pretty,
+    json_dumps,
     set_readonly,
     is_readonly,
     get_datetime,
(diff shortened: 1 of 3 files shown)


yarikoptic:
it is a short one but with a notable effect on all BIDS conversions: it should remain "backward compatible" in that even if prior subjects converted with older heudiconv, and new do not get swarm of `_scan.json` files, AFAIK it would remain BIDS-kosher.
Dear @nipy/team-heudiconv -- speedy review/feedback would be very appreciated!

yarikoptic:
ok, let's proceed and I will handle a fallout if any ;)

Issue #509, opened 2021-05-05
Commit message uses non-anonymized subject id


yarikoptic:
Here
```
$> git grep "Converted subj"
heudiconv/main.py:            msg = "Converted subject %s" % dl_msg
```
even if `--anon-cmd` was provided

yarikoptic:
doh -- this was not produced by heudiconv but reproin helper prototype script https://github.com/ReproNim/reproin/blob/master/bin/reproin , nothing to fix here besides that I ran into #511

PR #511, opened 2021-05-06
BF: make anonymize_script actually output anything and map determinstically

yarikoptic:
hash() would get a new value for every run for the same sid, so not a good
choice.  Without print script produced no output so code just saw empty
string and assumed that no annonymization was needed etc

was looking into the anonymization logic for #509 to realize that this script does nothing... fixed it but it seems that test is still working as expected which imho it shouldn't

TODOs
- [ ] figure out what is up with the test or my assumptions on how it should all work

### DIFF
--- a/heudiconv/tests/anonymize_script.py
+++ b/heudiconv/tests/anonymize_script.py
@@ -2,12 +2,12 @@
 
 import sys
 import re
-import ctypes
+import hashlib
 
 
 def bids_id_(sid):
     parsed_id = re.compile(r"^(?:sub-|)(.+)$").search(sid).group(1)
-    return str(ctypes.c_size_t(hash(parsed_id)).value)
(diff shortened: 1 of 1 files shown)


yarikoptic:
> figure out what is up with the test or my assumptions on how it should all work

anonymization script printed nothing, which then taken as empty sid, and thus main code just proceeded without any anonymization.  In 74550b8fb815e4139970d3959de08e6b8a4463e8 have made the `anonymize_sid` more robust and fail if anon script returned nothing or multiple words.  This would be a change in behavior but IMHO for good.

In subsequent commit 1d6b41bc4eacd4332fba45a218ee48e145b57dcf I have fixed the test -- it was just comparing "nothing" to be equal "nothing", and thus not useful and passing.

PR #512, opened 2021-05-07
FIX: Lingering temporary directories

mgxd:
This is a spin-off of #466 and a continuation of #485

After first confirming the problem with a failing test, this PR:
- Removes custom temporary directory object from various function signatures
- Replaces temporary directory creation with context managers (where possible)
- Simplifies and moves ``TempDirs`` object into a dedicated module

### DIFF
--- a/heudiconv/_tmpdirs.py
+++ b/heudiconv/_tmpdirs.py
@@ -1,0 +1,30 @@
+import logging
+import tempfile
+
+
+class TempDirs:
+    """A helper to centralize handling and cleanup of dirs"""
+
+    dirs = set()
+    lgr = logging.getLogger('heudiconv.tempdirs')
+
+    def __init__(self):
+        import atexit
+        atexit.register(self.cleanup)
+
+    def add_tmpdir(self, prefix=None):
+        tmpdir = tempfile.mkdtemp(prefix=prefix)
+        self.dirs.add(tmpdir)
+        return tmpdir
+
+    def cleanup(self):
+        """Cleanup tracked temporary directories"""
+        import shutil
+
+        for tmpdir in self.dirs:
+            try:
+                shutil.rmtree(tmpdir)
+            except FileNotFoundError:
+                pass
+
+tmpdirs = TempDirs()
--- a/heudiconv/_tmpdirs.py
+++ b/heudiconv/_tmpdirs.py
@@ -21,10 +21,13 @@ class TempDirs:
         """Cleanup tracked temporary directories"""
         import shutil
 
+        self.lgr.debug("Removing %d temporary directories", len(self.dirs))
         for tmpdir in self.dirs:
             try:
+                self.lgr.debug("Removing %s", tmpdir)
                 shutil.rmtree(tmpdir)
             except FileNotFoundError:
                 pass
+
 
 tmpdirs = TempDirs()
--- a/heudiconv/tests/test_heuristics.py
+++ b/heudiconv/tests/test_heuristics.py
@@ -3,7 +3,9 @@ from heudiconv.cli.run import main as runner
 import os
 import os.path as op
 from mock import patch
+from pathlib import Path
 from six.moves import StringIO
+import tempfile
 
 from glob import glob
 from os.path import join as pjoin, dirname
@@ -35,6 +37,9 @@ def test_smoke_convertall(tmpdir):
 
     args.extend(['-f', 'convertall'])
     runner(args)
+
+    # ensure we are not leaving lingering temporary directories
+    assert not list(Path(tempfile.gettempdir()).glob('heudiconv*'))
 
 
 @pytest.mark.parametrize('heuristic', ['reproin', 'convertall'])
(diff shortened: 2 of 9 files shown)


yarikoptic:
the fail
```
=================================== FAILURES ===================================
____________________________ test_smoke_convertall _____________________________
heudiconv/tests/test_heuristics.py:42: in test_smoke_convertall
    assert not list(Path(tempfile.gettempdir()).glob('heudiconv*'))
E   AssertionError: assert not [PosixPath('/tmp/heudiconvDCM42wzdx58'), PosixPath('/tmp/heudiconvDCMqb920gwc'), PosixPath('/tmp/heudiconvDCM3w5cw9bq')]
E    +  where [PosixPath('/tmp/heudiconvDCM42wzdx58'), PosixPath('/tmp/heudiconvDCMqb920gwc'), PosixPath('/tmp/heudiconvDCM3w5cw9bq')] = list(<generator object Path.glob at 0x7fa918d4df20>)
E    +    where <generator object Path.glob at 0x7fa918d4df20> = <bound method Path.glob of PosixPath('/tmp')>('heudiconv*')
E    +      where <bound method Path.glob of PosixPath('/tmp')> = PosixPath('/tmp').glob
E    +        where PosixPath('/tmp') = Path('/tmp')
E    +          where '/tmp' = <function gettempdir at 0x7fa924f2e040>()
E    +            where <function gettempdir at 0x7fa924f2e040> = tempfile.gettempdir
------------------------------ Captured log call -------------------------------
main.py                    262 INFO     Running heudiconv version 0.9.0 latest 0.9.0
main.py                    262 INFO     Running heudiconv version 0.9.0 latest 0.9.0
main.py                    291 INFO     Need to process 0 study sessions
=============== 1 failed, 55 passed, 4 skipped in 203.21 seconds ===============
The command "coverage run `which py.test` -s -v heudiconv" exited with 1.```

File: heudiconv/_tmpdirs.py [hunk 1]
```diff
@@ -0,0 +1,30 @@
+import logging
+import tempfile
+
+
+class TempDirs:
+    """A helper to centralize handling and cleanup of dirs"""
+
+    dirs = set()
+    lgr = logging.getLogger('heudiconv.tempdirs')
```
yarikoptic, line 9: all `lgr.` invocations were removed.  I think they should be brought back at debug level

yarikoptic: but it seems that their use was completely replaced with `tempfile.` constructs, then TempDirs isn't used at all, or what am I missing?

Although I am all for mitigating lingering temp dirs, and like that we would no longer pass them in functions (not sure why that was done originally anyways), I am not sure if switching to tempfile constructs wouldn't have detrimental DX effects:
- Absent easy to trigger logging to discover details on what is going on
- Absent centralization making it impossible to disable cleanup e.g. for the purpose of debugging

so eventually the centralized control might come back but would be improved to also provide context-manager style of invocation support .  So why not to do that right here and make it possible to do

```python
with tmpdirs(prefix='dcm2niix') as tmpdir:
```
which would take care about logging and cleaning it up right away (unless centrally disabled)?

yarikoptic (changes requested):

heudiconv/_tmpdirs.py
mgxd: In my ideal world, we would purge the hacky `TempDirs` (which is just a remnant of the old single script days) and wrap all necessary code with `tempfile.TemporaryDirectory` context managers. Then but doing so requires restructuring / rewriting some of the logic (it all stems from `get_extracted_dicoms`), and I am not up for that task.

> Absent centralization making it impossible to disable cleanup e.g. for the purpose of debugging

Are we doing this now? This feels like it should be addressed in a separate PR.

heudiconv/tests/test_heuristics.py
```diff
@@ -36,6 +38,9 @@ def test_smoke_convertall(tmpdir):
     args.extend(['-f', 'convertall'])
     runner(args)
 
+    # ensure we are not leaving lingering temporary directories
+    assert not list(Path(tempfile.gettempdir()).glob('heudiconv*'))
```
yarikoptic, line 42: note: wouldn't be safe (fail) in case of parallel testing, not that we are doing it now but someone could with e.g. py.test xdist

yarikoptic: and on the other hand - since  we do not use `heudiconv` prefix consistently across all uses - it might miss some temp dir to not be cleaned up.

I will push adding consistent `heudiconv` prefix (possibly changing names) to all temp dirs invocations.  Not that it how it was, but I think it would be good to have all temp dirs with consistent prefix.

mgxd: good point - do you have any recommendations on a cleaner way to test for linger tmpdirs?

yarikoptic: patch `TMPDIR` env var for this run, point to some directory... at little fragile since AFAIK we would not know if underlying code actually uses it, so would need an extra check (would be a useful test on its own IMHO) to test that `tmpdirs.add_tmpdir` and `tempfile.TemporaryDirectory(prefix='heudiconv-')` create a directory there.

also note -- `runner` AFAIK runs in the same process, doesn't it? so atexit wouldn't work out, so I guess we would need manual `.cleanup` here?

yarikoptic:
I will just close this one for now since haven't seen activity for awhile and IIRC @mgxd moved on to other adventures. But feel welcome to reopen or otherwise reincarnate

PR #522, opened 2021-08-20
TEMP: Run tests  on travis against dcm2niix snapshot build

yarikoptic:


yarikoptic:
didn't run anything... since new dcm2niix is out, I guess we will just test with some new PR ;)

PR #525, opened 2021-09-15
Drop outdated 3.5, fix compatibility with recent datalads

yarikoptic:


yarikoptic:
woohoo -- green. Merging

Issue #519, opened 2021-07-27
inconsistent Seqinfos when extracted from tarfile with mag+phase SBRef 


bpinsard:
### Summary

Running heudiconv on a tarfile with CMRR-MB BOLD with mag+phase export, the SBRef sometimes get seqinfos with `ImageType` `M`, and sometimes with `ImageType` `P`, causing trouble when writing the heuristics (related to #518, where for some series, I need to assign the `part-` BIDS tag in the heuristics). When running it on the separately extracted tarball it is consistent.
It seems that the order in which the files are scanned (or how a single dicom is chosen to extract the Seqinfo) varies when extracting from the tarfile here:
https://github.com/nipy/heudiconv/blob/f44c09992542819a0710f9bdcbac16c0d2eb2a37/heudiconv/parser.py#L100-L103 
One fix could for heudiconv to call: https://github.com/nipy/heudiconv/blob/a5e6718af00dadcd5f055f1a1997aa17ad61b362/heudiconv/parser.py#L29
on the dicom folder extracted in the temp folder instead of scanning the tarfile? 

Thanks!

### Platform details:

Choose one:
- [x] Local environment

- [ ] Container

- Heudiconv version: 

0.9.0

PR #535, opened 2021-11-05
try a simple fix for wrongly ordered files in tar file

bpinsard:
solves #519

### DIFF
--- a/heudiconv/parser.py
+++ b/heudiconv/parser.py
@@ -97,7 +97,7 @@ def get_extracted_dicoms(fl):
         for tm in tmembers:
             tm.mode = 0o700
         # get all files, assemble full path in tmp dir
-        tf_content = [m.name for m in tmembers if m.isfile()]
+        tf_content = sorted([m.name for m in tmembers if m.isfile()])
         # store full paths to each file, so we don't need to drag along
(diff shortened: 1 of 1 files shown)


yarikoptic:
But isn't the underlying issue that we (or is it a heuristic?) have some order effect?

yarikoptic:
File names could be 1, 2, ..., 10 and string sorting would screw it up

bpinsard:
You're right, our files are alphanumerically sorted by acquisition/content-creation order because it uses Siemens UID, and that might not be the general case.
However, I think it is sorted as well when not pulled from a tar filed (eg obtained from a directory).
https://github.com/nipy/heudiconv/blob/80a65385ef867124611d037fceac27d27e30f480/heudiconv/parser.py#L163-L164

In most cases the order of the file won't matter to the indexing nor to dcm2niix.

But for that particular case described in #519 :  complex data (at least 2 dicoms with M and P ImageType) in a single series will give a varying ImageType to the seqinfo depending on the sorting, as the parse will choose the first one as the example dicom.

yarikoptic:
your argument that we sort in non-tar handling convinces me. Let's proceed with this, and see if any fallout happens ;) Thank you @bpinsard !

PR #540, opened 2022-01-14
strip non-alphanumeric from session ids too

keithcallenberg:
Similar to our subject IDs, many of our session IDs have underscores as well. This causes BIDS validation to fail:

> [ERR] Ses label contain an Illegal Character hyphen or underscore. Please edit the filename as per BIDS spec. (code: 63 - SESSION_VALUE_CONTAINS_ILLEGAL_CHARACTER)

The same function in heudiconv that strips non-alphanumeric characters for subject IDs (`convert_sid_bids`) can be used for session IDs here too.

### DIFF
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -472,9 +472,8 @@ def get_formatted_scans_key_row(dcm_fn):
     row = ['n/a' if not str(e) else e for e in row]
     return row
 
-
 def convert_sid_bids(subject_id):
-    """Strips any non-BIDS compliant characters within subject_id
+    """Shim for stripping any non-BIDS compliant characters within subject_id
 
     Parameters
     ----------
@@ -487,12 +486,34 @@ def convert_sid_bids(subject_id):
     subject_id : string
         Original subject ID
     """
-    cleaner = lambda y: ''.join([x for x in y if x.isalnum()])
-    sid = cleaner(subject_id)
-    if not sid:
-        raise ValueError(
-            "Subject ID became empty after cleanup.  Please provide manually "
-            "a suitable alphanumeric subject ID")
-    lgr.warning('{0} contained nonalphanumeric character(s), subject '
-                'ID was cleaned to be {1}'.format(subject_id, sid))
+    sid, subject_id = sanitize_label(subject_id)
+    lgr.warning('Deprecation warning: convert_sid_bids() is deprecated, '
+                'please use sanitize_label() instead.')
+    
     return sid, subject_id
+
+
+def sanitize_label(label):
+    """Strips any non-BIDS compliant characters within label
+
+    Parameters
+    ----------
+    label : string
+
+    Returns
+    -------
+    clean_label : string
+        New, sanitized label
+    label : string
+        Original label
+    """
+    cleaner = lambda y: ''.join([x for x in y if x.isalnum()])
+    clean_label = cleaner(label)
+    if not clean_label:
+        raise ValueError(
+            "Label became empty after cleanup.  Please provide manually "
+            "a suitable alphanumeric label.")
+    if clean_label != label:
+        lgr.warning('{0} contained nonalphanumeric character(s), label '
+                    'was cleaned to be {1}'.format(label, clean_label))
+    return clean_label, label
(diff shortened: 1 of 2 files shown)


yarikoptic:
Thank you for the PR!  Indeed we should do that! But 
- reuse of `convert_sid_bids` for session might be misleading. If we decide to introduce some subject specific changes to that function, we could affect session cleansing as well.
- let's rename `convert_sid_bids` to `sanitize_label` or alike (I do not think it is worth keeping `bids` for every function in `bids` module, and we do have a mix ATM), provide a shim for `convert_sid_bids` which would also issue a deprecation warning suggesting to use `sanitize_label` (eventually we would remove it after a few releases)
- looking [at its code](https://github.com/nipy/heudiconv/blob/022772d1f6763ebe0998b6170696e2e2169adf0c/heudiconv/bids.py#L476:5) I am a little confused why it unconditionally issues a warning, which IMHO should be issued only if sanitized value differs from original
- we would need to add/adjust some unittest so we trigger/cover this new condition

yarikoptic:
@keithcallenberg interested to bring this PR to/over the finish line? ;)

keithcallenberg:
@yarikoptic I think I covered all but your last bullet point. Are there any existing tests that deal with invalid subject_ids?

heudiconv/bids.py
```diff
@@ -487,12 +486,34 @@ def convert_sid_bids(subject_id):
     subject_id : string
         Original subject ID
     """
+    sid, subject_id = sanitize_label(subject_id)
+    lgr.warning('Deprecation warning: convert_sid_bids() is deprecated, '
+                'please use sanitize_label() instead.')
```
yarikoptic, line 491: wow -- apparently we never issued any deprecation warning!  for those we should not use logger but rather warnings system.  

```shell
import warnings
...
warnings.warn("Message", DeprecationWarning)
```

heudiconv/bids.py
```diff
@@ -487,12 +486,34 @@ def convert_sid_bids(subject_id):
...
+    
+    return sid, subject_id
```
yarikoptic, line 493: just issue warning before even calling and then here

```
return sanitize_label(subject_id)
```
since there is no need to assign those to any variable meanwhile right?

yarikoptic:
> @yarikoptic I think I covered all but your last bullet point. Are there any existing tests that deal with invalid subject_ids?

dunno (would need to look - might be not), but you can add basic unittest for the function within `test_bids.py` to ensure it does what you expect it to do.

yarikoptic:
ping @keithcallenberg - interested to bring this PR to/over the finish line? thanks in advance for working on this regardless of that decision! ;)

l-espana:
Hi! Is there any update to this? We have a lot of longitudinal data and it would be very useful. Happy to help if it's needed!

l-espana:
Okay, I think I have some testing and some updates, but this is my first time contributing to something like this...what's the best way to commit/push/merge and which branch should I be using? I don't want to mess anything up, ha.

heudiconv/bids.py
```diff
@@ -487,12 +486,34 @@ def convert_sid_bids(subject_id):
...
-            "Subject ID became empty after cleanup.  Please provide manually "
-            "a suitable alphanumeric subject ID")
-    lgr.warning('{0} contained nonalphanumeric character(s), subject '
-                'ID was cleaned to be {1}'.format(subject_id, sid))
-    return sid, subject_id
+            "Label became empty after cleanup.  Please provide manually "
+            "a suitable alphanumeric label.")
+    if clean_label != label:
+        lgr.warning('{0} contained nonalphanumeric character(s), label '
+                    'was cleaned to be {1}'.format(label, clean_label))
```
yarikoptic, line 518:
```suggestion
        lgr.warning('%s contained nonalphanumeric character(s), label '
                    'was cleaned to be %s', label, clean_label)
```
so that if label or clean_label contains some %, logger doesn't freak out when tries to %-interpolate it

yarikoptic: I see that similar wrong code pattern was used before, but we better fix that since we found it now ;)

yarikoptic:
I have revived this PR in https://github.com/nipy/heudiconv/pull/647 to bring over the finish line.  I didn't push into this one directly since comes from `master`. @keithcallenberg your review of #647 would be welcome.

Issue #541, opened 2022-02-01
Automatic placement of `echo` results in BIDS-invalid format


andrew-yian-sun:
### Summary

I'm trying to get the following BIDS-valid output:
`sub-0061_echo-1_part-mag_MEGRE.nii.gz` Related Github issue at bids-specification: [https://github.com/bids-standard/bids-specification/issues/979](url)

However, it seems like `echo` is automatically generated by HeuDiConv to appear directly before `_MEGRE`, making it BIDS invalid (`echo` has to come before `part`).

The relevant part of my heuristic:
`megre_mag = create_key('sub-{subject}/{session}/anat/sub-{subject}_{session}_part-mag_MEGRE')`

I omit `echo` because it otherwise will generate 2 instances of echo in my output.

### Platform details:

Choose one:
- [ ] Local environment

- [x] Container: Docker, nipy/heudiconv:latest

- Heudiconv version: 0.9.0

pvelasco:
Hi @andrew-yian-sun ,
I can work on a fix. However, to test it, would you be able to provide a sample dataset?
Ideally, a very low resolution, just a few echoes, on a phantom.
Thanks,
-Pablo

andrew-yian-sun:
@pvelasco Yes, I'll try to get that to you tomorrow. Thanks!

tsalo:
@pvelasco It looks like the keyword you used in #542 didn't auto-close this issue. Now that #542 is merged, can this be closed?

pvelasco:
I think so.

tsalo:
Sounds good. I'll close this now, but @andrew-yian-sun if you find that the problems persists, please feel free to follow up here.

andrew-yian-sun:
@tsalo I recently tried to the recent version of heudiconv, 0.10.0 but echo is still appearing in the wrong place:
it should be `sub-001_ses-1_echo-1_part-mag_MEGRE`
but I am being given: `sub-001_ses-1_part-mag_echo-1_MEGRE`

Is this issue still being fixed?

tsalo:
The issue should be fixed, but the most recent release, 0.10.0, was in October of last year, while the fix was merged in February of this year.

@yarikoptic is it maybe time for 0.11.0?

pvelasco:
[...]

PR #542, opened 2022-02-01
BF: Fix the order of the 'echo' entity in the filename

pvelasco:
It addresses #541.


yarikoptic:
Thank you @pvelasco !  At some point we should provide some easy convenience for updating BIDS filenames in a more structured and schema driven way. For now this fix is great to have as is.

PR #546, opened 2022-02-24
ENH: specify POPULATE_INTENDED_FOR_OPTS within reproin following the doc

yarikoptic:
Just copy pasted from the doc.  I am afraid it might not work out as is for everyone, but ATM I think it is better to have it than not ;-)  I also tested locally on a sample conversion that `heudiconv --dbg -f reproin --command populate-intended-for` worked out nicely (for a single subj/sess conversion).


pvelasco (approved):
It looks good. I'm glad it works, at least for your dataset.

PR #547, opened 2022-02-28
Add test for the dataset that raised #541

pvelasco:


yarikoptic:
@pvelasco 
- note that travis isn't happy (didn't look inside)
- although not that heavy dicoms, we keep growing that tests/data and it starts worry me.  Should I just add MEGRE under http://datasets.datalad.org/?dir=/dicoms/velasco to stay near its MEEPI? ;)

pvelasco:
Yes, please, @yarikoptic; add MEGRE to datalad. Then I will correct the tests.

yarikoptic:
done: http://datasets.datalad.org/?dir=/dicoms/velasco/MEGRE

heudiconv/heuristics/bids_ME.py
```diff
@@ -21,9 +21,16 @@ def infotodict(seqinfo):
...
+    megre_mag = create_key('sub-{subject}/anat/sub-{subject}_part-mag_MEGRE')
+    megre_phase = create_key('sub-{subject}/anat/sub-{subject}_part-phase_MEGRE')
 
-    info = {bold: []}
+    info = {bold: [], megre_mag: [], megre_phase: []}
     for s in seqinfo:
         if '_ME_' in s.series_description:
             info[bold].append(s.series_id)
+        if 'GRE_QSM' in s.series_description:
+            if s.image_type[2] == 'M':
```
yarikoptic, line 32: note: this code block is not covered by tests... what data was supposed to have it covered?

heudiconv/tests/test_bids.py
```diff
@@ -1116,3 +1126,32 @@ def test_BIDSFile():
...
+    """ Unit test for the case of multi-echo GRE data with
+    magnitude and phase.
+    The different echoes should be labeled automatically.
+    """
+    tmpdir.chdir()
+    tmppath = tmpdir.strpath
+    try:
+        datadir = fetch_data(tmppath, "dicoms/velasco/{subID}")
+    except IncompleteResultsError as exc:
+        pytest.skip("Failed to fetch test data: %s" % str(exc))
```
yarikoptic, line 1142: note -- this line is "covered" which means that the test just skipped... I think I might see the reason

heudiconv/tests/test_bids.py
```diff
@@ -1116,3 +1126,32 @@ def test_BIDSFile():
...
+@pytest.mark.skipif(not have_datalad, reason="no datalad")
+def test_ME_mag_phase_conversion(tmpdir, subID='MEGRE', heuristic='bids_ME'):
+    """ Unit test for the case of multi-echo GRE data with
+    magnitude and phase.
+    The different echoes should be labeled automatically.
+    """
+    tmpdir.chdir()
+    tmppath = tmpdir.strpath
+    try:
+        datadir = fetch_data(tmppath, "dicoms/velasco/{subID}")
```
yarikoptic, line 1140: may be
```suggestion
        datadir = fetch_data(tmppath, f"dicoms/velasco/{subID}")
```
would explain why test seems to be skipping?  
why would you want to skip anyways if you had datalad installed (there is `@skipif`) , so let's just do without skipping and corrected (will give another bigger suggestion)

heudiconv/tests/test_bids.py
yarikoptic, line 1142: in datalad we trust? ;)

```suggestion
    datadir = fetch_data(tmppath, f"dicoms/velasco/{subID}")
```

yarikoptic (changes requested):
test seems in need of fixing.  To expedite, I will accept one of my changes, feel free to revert (well - just `push --force` your previous state) -- I just want to see quickly if may be it would resolve so I could accept

heudiconv/tests/test_bids.py
yarikoptic: my bad -- apparently (for some reason, didn't check) this is a common usage pattern, so let's keep as that for now, I will just fix the f-string

pvelasco: Sorry, I'm not really sure if I need to do anything on this PR...

yarikoptic: CI failures still to be checked, test fixed

pvelasco: I forgot to add the extension for the heuristic file in `tests.test_bids.test_ME_mag_phase_conversion`.
Addressed in https://github.com/cbinyu/heudiconv/commit/aee3d9ba81f55067a79de3ffcfb3bd39d74e6659.
It should be ready now.

yarikoptic:
coolio, thank you @pvelasco !

PR #558, opened 2022-04-06
Set up auto

### LINKED ISSUE #557: Add `auto` and building/pushing of docker image
Dear @jwodder -- need your help on this. We need to `auto`mate this poor soul since releasing is pain.  I can even go through already merged PRs since last release and label/adjust titles them as needed.  Please do the rest.

Also we need to automate building/pushing docker images to properly address https://github.com/nipy/heudiconv/issues/538 . Those are to be done at release time as well after successful releasing. `Dockerfile` is in the root of the repo and produced by  http://github.com/nipy/heudiconv/blob/HEAD/utils/gen-docker-image.sh script which I might need to adjust before next release.

jwodder:
Closes #557.

This PR also needs to be accompanied by the following changes:
- [x] Create the necessary labels for `auto` by running `GH_TOKEN=... auto create-labels` in a copy of this repository containing `.autorc`
- [x] All PRs merged since the previous release should receive appropriate labels. Done, this is the PRs: https://github.com/nipy/heudiconv/pulls?q=is%3Apr+is%3Aclosed+sort%3Acreated-desc+merged%3A%3E%3D2021-09-16
- [x] An auth token for uploading assets for the `heudiconv` project to PyPI must be saved as a secret named "`PYPI_TOKEN`"
- [x] A Docker Hub username and token for uploading `nipy/heudiconv` images must be saved as secrets named `DOCKER_LOGIN` and `DOCKER_TOKEN`, respectively (~~@yarikoptic is waiting for repronimservices to be added to the team to be used here~~ did for another account)


jwodder:
@yarikoptic Dockerization added.

.github/workflows/release.yml
```diff
@@ -0,0 +1,88 @@
...
+  build-docker:
+    runs-on: ubuntu-latest
+    needs: auto-release
+    if: needs.auto-release.outputs.auto-version != ''
+    steps:
+      - name: Get tag of latest release
+        id: latest-release
+        run: |
+          latest_tag="$(curl -fsSL https://api.github.com/repos/$GITHUB_REPOSITORY/releases/latest | jq -r .tag_name)"
+          echo "::set-output name=tag::$latest_tag"
```
yarikoptic, line 61: why `needs.auto-release.outputs.auto-version` can't be used here? I am afraid if api.github has some delay in propagating effects of `auto shipit`  we might end up using some wrong tag here.

.github/workflows/release.yml
```diff
@@ -0,0 +1,88 @@
...
+        run: bash gen-docker-image.sh
+        working-directory: utils
+
+      - name: Build Docker images
+        run: |
+          docker build \
+            -t nipy/heudiconv:master \
+            -t nipy/heudiconv:latest \
+            -t nipy/heudiconv:${{ steps.latest-release.outputs.tag }} \
+            .
```
yarikoptic, line 78: I think no harm, but just to be clear -- this would a "benign race" with a separate `.github/workflows/docker.yaml` workflow here to build/push one for `master`, thus in principle possibly making it a bit inconsistent (for some releases the same image used for all 3 and for others, `master` would be pointing to another built image although of the same content). correct?

Makefile
```diff
@@ -1,18 +0,0 @@
-PYTHON ?= python3
-
-all:
-	echo 'nothing by default'
-
-prep_release:
-	# take previous one, and replace with the next one
-	utils/prep_release
```
yarikoptic, line 8: if github workflow fails somehow, how could we prep release locally? i.e. how at least bumpversion would be done?

I wonder if either Makefile or some documentation should be left behind on that.

yarikoptic (commented):
may be to add some CONTRIBUTING.md with brief description of this release process, and what to do if it doesn't work out, i.e. is there a way to at least introduce version adjustments?

but also I thought if it would be possible to add versioneer or your versioningit while at it , or it better be a separate PR?  I am tired of `--version` to be not git informed in heudiconv.

.github/workflows/release.yml
jwodder: Because `auto version` outputs a patch level, not a version number.

.github/workflows/release.yml
jwodder: Yes.

.github/workflows/release.yml
yarikoptic: I feel that we had similar problem somewhere else and did find how to trick `auto` to tell us the full version, didn't we?  I would really prefer to not rely on the logic of jumping through github without any validation that the tag does correspond to this state. Wouldn't prior step tag it, so we could just use `git describe` here and get a clean version?

jwodder: Workflow adjusted.

jwodder:
@yarikoptic 

> but also I thought if it would be possible to add versioneer or your versioningit while at it

If we use versioningit, then presumably we would want to get rid of all the explicit version numbers bumped by `prep_release`/`bump2version`.  This would mean:

* Configuring whatever builds the docs to install the package beforehand so that the `__version__` can be imported in `docs/conf.py`
* Rewriting the Docker & Singularity `pull` commands in `docs/installation.rst` and the scripts in `docs/usage.rst` to no longer contain an explicit version number
  * Once we start pushing "latest" images to Docker Hub, we can use `latest` as the image tag in the `docker pull` command; would that work for Singularity as well?

yarikoptic:
All of the above sounds good to me. Yes, singularity should pull from docker hub, we will provide also images from ///repronim/containers later

jwodder:
@yarikoptic versioningit added; bump2version eliminated.

yarikoptic:
ok, let's proceed and see where it takes us

PR #566, opened 2022-05-12
Make versioningit write version to file; make setup.py read version as fallback

### LINKED ISSUE #565: versioningit: write/use static version
ATM it is fully PEP 518 driven package.  That is all nice for future/modern distributions but backporting to older debian/ubuntus would be painful since would require not only versioningit there (easy) but also backport of recent pybuild and also doing other dances.

https://github.com/jwodder/versioningit says that versioningit "Can optionally write the final version to a file for loading at runtime", so I think that should be done, such file shipped within sdist on pypi and then https://github.com/nipy/heudiconv/blob/master/heudiconv/__init__.py#L6 adjusted to fall back to using that file in case of versioningit not being found or version being "told" to be 0.0.0.

I guess may be setup.py also would need some tune up to provide that version for "old world" `python setup.py install` to get the right version?
 
That would allow to make it feasible to have heudiconv NeuroDebian packages updated for older releases of debian and ubuntu.

related ref:
- https://github.com/jwodder/versioningit/issues/28

jwodder:
Closes #565.

### DIFF
--- a/heudiconv/__init__.py
+++ b/heudiconv/__init__.py
@@ -2,13 +2,7 @@
 import logging
 import os
 from .info import __packagename__
-
-try:
-    from importlib.metadata import version
-except ImportError:
-    from importlib_metadata import version
-
-__version__ = version(__name__)
+from ._version import __version__
 
 # Rudimentary logging support.
 lgr = logging.getLogger(__name__)
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,17 @@ def main():
     with open(info_file) as infofile:
         exec(infofile.read(), globals(), ldict)
 
+    try:
+        import versioningit
+    except ImportError:
+        # versioningit isn't installed; assume we're building a Debian package
+        # from an sdist on an older Debian that doesn't support pybuild
+        vglobals = {}
+        with open(op.join(op.dirname(__file__), "heudiconv", "_version.py")) as fp:
+            exec(fp.read(), vglobals)
+        kwargs = {"version": vglobals["__version__"]}
+    else:
+        kwargs = {}
 
     def findsome(subdir, extensions):
         """Find files under subdir having specified extensions
@@ -58,7 +69,8 @@ def main():
                         op.join('data', '*.dcm'),
                         op.join('data', '*', '*.dcm')
             ],
-        }
+        },
+        **kwargs,
     )
 
 
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ _build/
 build/
 dist/
 .vscode/
+_version.py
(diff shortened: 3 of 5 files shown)


heudiconv/__init__.py
```diff
@@ -2,13 +2,7 @@
...
 import os
 from .info import __packagename__
-
-try:
-    from importlib.metadata import version
-except ImportError:
-    from importlib_metadata import version
-
-__version__ = version(__name__)
+from ._version import __version__
```
yarikoptic, line 5: I think it should just fall down to this if `__version__` is '0.0.0' ie not provided by versioningit, since I would not be getting `_version.__version__` automagically updating with every git commit and thus `heudiconv.__version__` even when I have versioningit (e.g. on my laptop) would not reflect actual version.

setup.py
```diff
@@ -21,6 +21,15 @@ def main():
     with open(info_file) as infofile:
         exec(infofile.read(), globals(), ldict)
 
+    try:
+        import versioningit
+    except ImportError:
+        # versioningit isn't installed; assume we're building a Debian package
```
yarikoptic, line 27: well -- not just Debian but any packaging environment (e.g. conda-forge) which doesn't have versioningit installed and/or not using PEP 518

heudiconv/__init__.py
jwodder: `versioningit` doesn't update with every Git commit either, unless you reinstall after each commit, but reinstalling causes `_version.py` to be updated anyway.

setup.py
jwodder: We should be using `versioningit` when building in conda-forge; the lack thereof is the cause of things like https://github.com/dandi/dandi-schema/issues/135.  The comment should not be generalized to other non-PEP 517 environments, as they should be discouraged as much as possible.

setup.py
```diff
@@ -21,6 +21,15 @@ def main():
...
+        # from an sdist on an older Debian that doesn't support pybuild
+        from heudiconv._version import __version__
```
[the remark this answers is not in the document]
jwodder: Adjusted.

heudiconv/__init__.py
yarikoptic: oh, right -- it relies on the generated PKG-INFO, so unlike versioneer it is not "dynamic"... that is a cons of versioningit then since I will see myself reporting "wrong" versions.  But then indeed this patch would be fine as far as versioningit goes I guess.

setup.py
yarikoptic: well -- silent discouragement is not a discouragement, it is "allowance" ;-)  Similarly I consider the fact that packages silently install with degenerate version -- it would be better if they failed.  So such plugs as the diff here now is generally needed not only in heudiconv and not only on Debian but IMHO in every versiongit using package.  But I will not fight over a comment here.

yarikoptic:
Thank you @jwodder ! In limited testing seems to work, will merge with `release` to come so I could test on debian package without going ad-hoc workflow to update it

PR #569, opened 2022-06-02
Minor face-lifts to ReproIn: align doc and code better to BIDS terms, address deprecation warnings etc

yarikoptic:
Should have no effect on functionality.


yarikoptic:
I don't think there could be negative effect on any users, so I will proceed with merging.

PR #570, opened 2022-06-06
BF(?): make _setter images be taken as scouts - only DICOMs are saved

yarikoptic:
Aiming to fix #40

@cookpa @tsalo -- may be such approach would be sufficient (i.e. would not cause some disturbance as e.g. not saving DICOM tarballs correctly for T1w)?  Would you be so kind to try on your data?

edit: an alternative -- completely ignore/skip them


yarikoptic:
have not got any feedback, and will just assume that it would all work ;)  we are at v0.12.2-93-g409e6bb so it is really a time to release.

PR #578, opened 2022-07-15
Clean session labels of non-alphanumeric characters #540

l-espana:
Most of the code from [keithcallenberg](https://github.com/keithcallenberg), with unit testing and small modifications per previous [PR](https://github.com/nipy/heudiconv/pull/540) comments made by me.

### DIFF
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -12,6 +12,7 @@ import csv
 from random import sample
 from glob import glob
 import errno
+import warnings
 
 from .external.pydicom import dcm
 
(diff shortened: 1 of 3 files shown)


yarikoptic:
Thank you!
- unfortunately there seems to be too many changes which are unrelated (some reformatting etc) to the actual fix to be implemented.  diff should be minimized to only what needs to be done to accomplish the mission of "Clean session labels of non-alphanumeric characters"
- ideally it should have retained commits done by @keithcallenberg to acknowledge his contribution as well, and then just had extra commits with extra changes (again -- just minimized to what matters)
- not yet sure why CI didn't run at all! very odd -- will need to check

Issue #580, opened 2022-07-18
question about using a dicom private tag in heuristic file


kcho:
### Summary

Hi,

I'm working with Siemens XA30 data, and I'm trying to use the information contained in the private dicom header in re-arranging output names. Is there a way to extract private tag information and use this in my `heuristic` file?

I can extract the information that I need using `pydicom` with the following commands

```
from pydicom import dcmread
ds = dcmread('dicom.dcm')
ds[0x52009230][0][0x002111fe][0][0x00211175].value
```

Thanks for the great tool.

### Platform details:

Choose one:
- [x] Local environment

CentOS 3.10.0-862.14.4.el7.x86_64
python 3.7.3

- [ ] Container

- Heudiconv version: 

0.11.3

yarikoptic:
there is https://github.com/nipy/heudiconv/pull/333 which I "resisted".  I guess it would have provided you with a remedy, right?

kcho:
Thanks!

PR #581, opened 2022-07-19
Add support for a  custom seqinfo to extract from DICOMs any additional metadata desired for a heuristic

yarikoptic:
Getting back to #333 (attn @bpinsard ) and recent #580 (attn @kcho) decided may be to add to the army of custom ad-hoc callbacks etc heuristics could provide.  So, WDYT if heuristic could provide `custom_seqinfo` callable (crude example is below in PR) where any information could be extracted from already loaded nibabel's dicom wrapper or just a full list of files for that sequence.  I think it would be a slightly better solution to #333 since would not *always* store potentially sensitive path in the extracted metadata record  and potentially could operate on already loaded DICOM metadata thus making it faster to extract extra metadata.

@bpinsard @kcho -- would smth like that work for your cases?

TODOs
- [ ] ~~replace~~compliment hashing check with some more adequate, may be round trip check through json + verification that no `\t` is present. if `\t` -- kaboom. Add to documentation that `\t` must be sanitized, or add such sanitization (and document it)
- [x] add basic test. Since `convertall` already got that custom field filled, test should just check if it is there.

### DIFF
--- a/docs/heuristics.rst
+++ b/docs/heuristics.rst
@@ -120,7 +120,7 @@ or::
         return seqinfos  # ordered dict containing seqinfo objects: list of DICOMs
 
 ---------------------------------------------------------------
-``custom_seqinfo(series_files, wrapper)``
+``custom_seqinfo(wrapper, series_files)``
 ---------------------------------------------------------------
 If present this function will be called on each group of dicoms with
 a sample nibabel dicom wrapper to extract additional information
--- a/docs/heuristics.rst
+++ b/docs/heuristics.rst
@@ -122,7 +122,7 @@ or::
 ---------------------------------------------------------------
 ``custom_seqinfo(series_files, wrapper)``
 ---------------------------------------------------------------
-If present this function will be called on eacg group of dicoms with
+If present this function will be called on each group of dicoms with
 a sample nibabel dicom wrapper to extract additional information
 to be used in ``infotodict``.
 
--- a/docs/heuristics.rst
+++ b/docs/heuristics.rst
@@ -119,6 +119,19 @@ or::
         ...
         return seqinfos  # ordered dict containing seqinfo objects: list of DICOMs
 
+---------------------------------------------------------------
+``custom_seqinfo(series_files, wrapper)``
+---------------------------------------------------------------
+If present this function will be called on eacg group of dicoms with
+a sample nibabel dicom wrapper to extract additional information
+to be used in ``infotodict``.
+
+Importantly the return value of that function needs to be hashable.
+For instance the following non-hashable types can be converted to an alternative
+hashable type:
+- list > tuple
+- dict > frozendict
+- arrays > bytes (tobytes(), or pickle.dumps), str or tuple of tuples.
 
 -------------------------------
 ``POPULATE_INTENDED_FOR_OPTS``
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -9,7 +9,18 @@ import os.path as op
 from pathlib import Path
 import sys
 import tarfile
-from typing import TYPE_CHECKING, Any, Dict, Hashable, List, NamedTuple, Optional, Union, Protocol, overload
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Hashable,
+    List,
+    NamedTuple,
+    Optional,
+    Protocol,
+    Union,
+    overload,
+)
 from unittest.mock import patch
 import warnings
 
@@ -90,14 +101,19 @@ def create_seqinfo(
     global total_files
     total_files += len(series_files)
 
-    custom_seqinfo_data = custom_seqinfo(wrapper=mw, series_files=series_files) \
-        if custom_seqinfo else None
+    custom_seqinfo_data = (
+        custom_seqinfo(wrapper=mw, series_files=series_files)
+        if custom_seqinfo
+        else None
+    )
     try:
         hash(custom_seqinfo_data)
     except TypeError:
-        raise RuntimeError("Data returned by the heuristics custom_seqinfo is not hashable. "
-                           "See https://heudiconv.readthedocs.io/en/latest/heuristics.html#custom_seqinfo for more "
-                           "details.")
+        raise RuntimeError(
+            "Data returned by the heuristics custom_seqinfo is not hashable. "
+            "See https://heudiconv.readthedocs.io/en/latest/heuristics.html#custom_seqinfo for more "
+            "details."
+        )
 
     return SeqInfo(
         total_files_till_now=total_files,
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -9,7 +9,8 @@ import os.path as op
 from pathlib import Path
 import sys
 import tarfile
-from typing import TYPE_CHECKING, Any, Dict, List, NamedTuple, Optional, Union, overload
+from typing import TYPE_CHECKING, Any, Dict, Hashable, List, NamedTuple, Optional, Union, overload
+from typing_extensions import Protocol
 from unittest.mock import patch
 import warnings
 
@@ -42,7 +43,16 @@ total_files = 0
 compresslevel = 9
 
 
-def create_seqinfo(mw: dw.Wrapper, series_files: list[str], series_id: str) -> SeqInfo:
+class CustomSeqinfoT(Protocol):
+    def __call__(self, wrapper: dw.Wrapper, series_files: list[str]) -> Hashable: ...
+
+
+def create_seqinfo(
+    mw: dw.Wrapper,
+    series_files: list[str],
+    series_id: str,
+    custom_seqinfo: CustomSeqinfoT | None = None,
+) -> SeqInfo:
     """Generate sequence info
 
     Parameters
@@ -109,6 +119,9 @@ def create_seqinfo(mw: dw.Wrapper, series_files: list[str], series_id: str) -> S
         date=dcminfo.get("AcquisitionDate"),
         series_uid=dcminfo.get("SeriesInstanceUID"),
         time=dcminfo.get("AcquisitionTime"),
+        custom =
+            custom_seqinfo(wrapper=mw, series_files=series_files)
+            if custom_seqinfo else None,
     )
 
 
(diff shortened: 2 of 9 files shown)


bpinsard:
That's way better than what I had PR-ed years back and should perfectly fit our needs.
Thanks!

kcho:
This will also work for me. Thanks for the PR!

bpinsard:
I finally tested the PR. I think that `group_dicoms_into_seqinfos` call in `parser.py` should also be given the `custom_seqinfo`  for it to be accessible in `infotoids`. Apart from that it fit the needs.

satra:
@yarikoptic - perhaps the return from `custom_seqinfo` could be a dictionary to add multiple columns instead of a single column named `custom`. would allow more generalizability to fields.

i would suggest decoupling anonymity from this though, since any custom return could potentially return things that are identifiable. i would leave that part to heuristic developers.

yarikoptic:
> @yarikoptic - perhaps the return from `custom_seqinfo` could be a dictionary to add multiple columns instead of a single column named `custom`. would allow more generalizability to fields.

ATM it indeed can be anything (e.g. that tuple in the [convertall example](https://github.com/nipy/heudiconv/pull/581/files#diff-6326c5f8ebb4d7cdb70ae391ddcf7e4741a3f1720d5e69bcb0d80d5475dd234cR10).  But indeed it would then be a single field `custom`, potentially not reloadable as a data structure in the round trip to that `.tsv` where it would be dumped -- I am yet to check what would happen/how it would work.

Indeed we could make it return a dict to formalize such extension, but I don't think we would be able to "dynamically" expand that  seqinfo_fields list of fields defined at the module level https://github.com/nipy/heudiconv/pull/581/files#diff-2aa4d95adec423a8c619a45d9676cb573bfdcbdb53d84af3d815ec34aee99d75R27 .  Pairing with aforementioned round trip we might though consider it to be limited to a data structure serializable through `json` and have round trip that way... (but because of ".tsv" might need special handling of tabs, uff)

> i would suggest decoupling anonymity from this though, since any custom return could potentially return things that are identifiable. i would leave that part to heuristic developers.

as far as I see it -- it is "decoupled" in a sense "it doesn't care".  What did you have in mind?

heudiconv/dicoms.py
```diff
@@ -62,6 +62,14 @@ def create_seqinfo(mw, series_files, series_id):
     global total_files
     total_files += len(series_files)
 
+    custom_seqinfo_data = custom_seqinfo(wrapper=mw, series_files=series_files) \
```
yarikoptic, line 65:
```suggestion
    custom_seqinfo_data = custom_seqinfo(series_files=series_files, wrapper=mw) \
```
to correspond order in the doc?

heudiconv/dicoms.py
```diff
@@ -62,6 +62,14 @@ def create_seqinfo(mw, series_files, series_id):
...
+        if custom_seqinfo else None
+    try:
+        hash(custom_seqinfo_data)
```
yarikoptic, line 68: @bpinsard remind me why `hashability` is the criteria instead of being serializable to string within .tsv?

bpinsard: Otherwise it crashes when it tries using the seqinfo as a dict key here:

```
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/basile/data/src/heudiconv/heudiconv/cli/run.py", line 177, in <module>
    main()
  File "/home/basile/data/src/heudiconv/heudiconv/cli/run.py", line 24, in main
    workflow(**kwargs)
  File "/home/basile/data/src/heudiconv/heudiconv/main.py", line 327, in workflow
    study_sessions = get_study_sessions(dicom_dir_template, files,
  File "/home/basile/data/src/heudiconv/heudiconv/parser.py", line 178, in get_study_sessions
    seqinfo_dict = group_dicoms_into_seqinfos(
  File "/home/basile/data/src/heudiconv/heudiconv/dicoms.py", line 310, in group_dicoms_into_seqinfos
    seqinfos[key][seqinfo] = series_files
TypeError: unhashable type: 'dict'
```

heudiconv/dicoms.py
```diff
@@ -62,6 +62,14 @@ def create_seqinfo(mw, series_files, series_id):
     global total_files
     total_files += len(series_files)
 
+    custom_seqinfo_data = custom_seqinfo(series_files=series_files, wrapper=mw) \
+        if custom_seqinfo else None
+    try:
+        hash(custom_seqinfo_data)
+    except TypeError:
+        raise RuntimeError("Data returned by the heuristics custom_seqinfo is not hashable. "
+        "See https://heudiconv.readthedocs.io/en/latest/heuristics.html#custom_seqinfo for more details.")
```
yarikoptic, line 71: @bpinsard so WDYT about adding some check for roundtrip serialization into/from .tsv? a "weaker" alternative could be to test for presence of `\t` or `\n` symbols which might break it, or they wouldn't ?  

apparently we did not add any test yet here, should have that.

heudiconv/heuristics/convertall.py
```diff
@@ -1,12 +1,29 @@
...
+    # for already loaded DICOM data, and including storing/returning
+    # the sample series file as was requested
+    # in https://github.com/nipy/heudiconv/pull/333
+    from nibabel.nicom.dicomwrappers import WrapperError
+    try:
+        affine = wrapper.affine.tostring()
+    except WrapperError:
+        lgr.exception("Errored out while obtaining/converting affine")
+        affine = None
+    return affine, series_files[0]
```
yarikoptic, line 24: sneaky sneaky @bpinsard !! ;) this might go into docs on `custom_seqinfo` but should not become part of `convertall` heuristic IMHO! ;-)

may be middle ground (and useful for another demonstration) could be creating a `converall-custom.py` heuristic which would be

```python
"""A demo convertall heuristic with custom_seqinfo extracting affine and sample DICOM path

This heuristic also demonstrates on how to create a "derived" heuristic which would augment
behavior of an already existing heuristic without complete rewrite.  Such approach could be
useful for heuristic like  reproin  to overload mapping etc.
"""

from .convertall import *

def custom_seqinfo(series_files, wrapper, **kw):
    # Just a dummy demo for what custom_seqinfo could get/do
    # for already loaded DICOM data, and including storing/returning
    # the sample series file as was requested
    # in https://github.com/nipy/heudiconv/pull/333
    from nibabel.nicom.dicomwrappers import WrapperError
    try:
        affine = wrapper.affine.tostring()
    except WrapperError:
        lgr.exception("Errored out while obtaining/converting affine")
        affine = None
    return affine, series_files[0]
```
?

satra:
the main intent is to give access to other fields in the dicom file as part of the heuristic. the only reason the tsv file existed is because that was produced by one of the early conversion tools we used, and acted as a cache between having to read all the dicom files and the heuristic. it is also the reason it had weird unused fields.

one could read a lot of dicom files very quickly now, much quicker than the original dicomconvert.py script :)

- one could make the process completely dynamic and allow any custom fields beyond the  ones hardcoded
- allow for arbitrary expansion of the columns knowing that only the heuristic can use them. 

you could add `custom_` prefix to any column name if the custom extractor returned a dictionary of fields and values. this would also force the heuristic writer to use the `custom_` prefix and provide round trip capabilities.

heudiconv/heuristics/convertall.py
yarikoptic: @bpinsard ping on above

bpinsard: Very sorry for the late follow-up.

I am not sure to understand what you are asking me.
In that commit https://github.com/nipy/heudiconv/pull/581/commits/f104fbcda275b2b8e336ba3cf1b424bb9fb8cef0 I  tried fixing that function which you added to the convertall heuristics in the PR, as the tests were crashing on the test data (due to a missing affine in the header). If we want the tests to pass it seems necessary. Or maybe I misunderstand your point here.

docs/heuristics.rst
```diff
@@ -93,6 +93,19 @@ or::
         ...
         return seqinfos  # ordered dict containing seqinfo objects: list of DICOMs
 
+---------------------------------------------------------------
+``custom_seqinfo(series_files, wrapper)``
```
yarikoptic, line 97:
```suggestion
``custom_seqinfo(wrapper, series_files)``
```

bpinsard:
Happy new year! 
Is there a chance this will get merged? 
What can I do to help finalize that feature that we have been using in production for a while now.
Thanks!

yarikoptic:
eh, we need to update (resolve conflicts) and follow original description to add tests

heudiconv/dicoms.py
bpinsard: Is there any part of heudiconv that might load the `dicominfo*.tsv` that get written in `.heudiconv`, in a second run with overwrite for instance?

For now it convert `custom` values to a str, eg. when using frozendict for hashable dict:
`frozendict.frozendict({'patient_name': 'p02_ana012', 'pe_dir': 'ROW', 'pe_dir_pos': 1, 'body_part': 'BRAIN', 'scan_options': "['IR', 'WE']", 'image_comments': '', 'slice_orient': 'Sag', 'echo_number': 'None', 'rescale_slope': None, 'receive_coil': 'HC1-7;NC1,2', 'image_history': 'ChannelMixing:ND=true_CMM=1_CDM=1;ACCAlgo:9', 'ice_dims': 'X_1_1_1_1_1_4_1_1_1_1_1_35'})` 

So not necessarily easily unserializable depending on what the custom_seqinfo is returning.

heudiconv/heuristics/convertall.py
yarikoptic: ha ha -- it indeed was me who added it in 3bc1f4e58ef574dcf6565b6c47a74be6c7eeab46 but I did mark it as TMP (i.e dropped for final version)

```
commit 3bc1f4e58ef574dcf6565b6c47a74be6c7eeab46
Author: Yaroslav Halchenko <debian@onerussian.com>
Date:   Tue Jul 19 16:10:25 2022 -0400

    TMP: just a demonstration on how custom_seqinfo could/should be used
```

now pushed as that proposed new heuristic

bpinsard:
Fantastic ! Many thanks !

PR #592, opened 2022-09-23
Beginning update reproin.py to deal with ses-X* instances

Michael-Sun:


heudiconv/heuristics/reproin.py
```diff
@@ -124,6 +124,9 @@
 
 - We replace all ( with { and ) with } to be able e.g. to specify session {date}
 - "WIP " prefix unconditionally added by the scanner is stripped
+
+
+-Michael-Sun Plans on adding a heuristic to deal with ses-X* instances
```
yarikoptic, line 129: make it happen!

Michael-Sun:
Made some edits to do some default behavior if it sees and X* or x*

heudiconv/heuristics/reproin.py
yarikoptic: add above , where `_ses-` marker described, description that `XXX` markers are dealt in some other way.

heudiconv/heuristics/reproin.py
```diff
@@ -780,6 +783,13 @@ def infotoids(seqinfos, outdir):
                     "Should have got a single session marker.  Got following: %s"
                     % ', '.join(map(repr, ses_markers))
                 )
+            if re.match("(X|x)*", ses_markers):
+               # replace X* or x* sessions with prior_sessions+1
+               prior_sessions = sorted(glob(os.path.join(sessions_dir, 'ses-*')))
+               if prior_sessions:
+                session = '%03d' % (len(prior_sessions) + 1)
+               else:
+                session='001'
```
yarikoptic, line 792: I don't think we should hardcode to `%03d` since intention on formatting might have been different etc.

- do not just take the length of `prior_sessions` but
  - extract labels (what comes after `ses-`), see what length they have.  if non-uniform (`1,...,12`) - use plain `'%d'`. If uniform - use `f'%0{len(ses_labels[0])}d`'`
  - do not base on `len(prior_sessions)` but convert all session labels into `int`. If errors - errors. Take max of those ints, so smth like `max(map(int, ses_labels))`
- make this logic into a function  `get_next_xxx_session(prior_session_dirs: list[str])`, for which also add unittests for it

heudiconv/heuristics/reproin.py
```diff
@@ -780,6 +783,13 @@ def infotoids(seqinfos, outdir):
                     "Should have got a single session marker.  Got following: %s"
                     % ', '.join(map(repr, ses_markers))
                 )
+            if re.match("(X|x)*", ses_markers):
```
yarikoptic, line 786:
```suggestion
            if re.match("XX+$", ses_markers) or re.match("xx+$", ses_markers):
```

to
- ensure consistent `xxx` or `XXX` and not `XxX`
- require at least 2 of those xs
- no other symbol there (anchor)

yarikoptic (changes requested):
left comments on how to proceed.  Will move to `draft` mode. Please undraft when you consider it ready and ping me

Issue #595, opened 2022-10-05
Dockerfile building fails


neurorepro:
### Summary

The Dockerfile provided in the Github repo fails to build

### Platform details:

Choose one:
- [x] Local environment
Ubuntu 20.04 LTS
- [ ] Container

- Heudiconv version: 

To reproduce:

```
cd /tmp
git clone git@github.com:nipy/heudiconv.git
cd heudiconv/
docker build -t deleteme/heudiconv .
```

<details>
<summary>Output</summary>

```console
Sending build context to Docker daemon  24.85MB
Step 1/9 : FROM neurodebian:bullseye
 ---> 98be26081cb0
Step 2/9 : ENV PATH="/opt/dcm2niix-v1.0.20211006/bin:$PATH"
 ---> Using cache
 ---> 3b7abc5dee8d
Step 3/9 : RUN apt-get update -qq     && apt-get install -y -q --no-install-recommends            ca-certificates            cmake            g++            gcc            git            make            pigz            zlib1g-dev     && rm -rf /var/lib/apt/lists/*     && git clone https://github.com/rordenlab/dcm2niix /tmp/dcm2niix     && cd /tmp/dcm2niix     && git fetch --tags     && git checkout v1.0.20211006     && mkdir /tmp/dcm2niix/build     && cd /tmp/dcm2niix/build     && cmake  -DCMAKE_INSTALL_PREFIX:PATH=/opt/dcm2niix-v1.0.20211006 ..     && make -j1     && make install     && rm -rf /tmp/dcm2niix
 ---> Using cache
 ---> 491d0894e9cf
Step 4/9 : RUN apt-get update -qq            && apt-get install -y -q --no-install-recommends                   gcc                   git                   git-annex-standalone                   libc-dev                   liblzma-dev                   netbase                   pigz            && rm -rf /var/lib/apt/lists/*
 ---> Using cache
 ---> a4b6618691c1
Step 5/9 : COPY [".",       "/src/heudiconv"]
 ---> 4af3f5e9a2af
Step 6/9 : ENV CONDA_DIR="/opt/miniconda-latest"     PATH="/opt/miniconda-latest/bin:$PATH"
 ---> Running in 3f8dbcbad2b5
Removing intermediate container 3f8dbcbad2b5
 ---> f334baf87595
[...]

PR #596, opened 2022-10-05
Bf docker: use python3.9 and newer dcm2niix

yarikoptic:
If merged, fixes #595

PR #597, opened 2022-10-05
Fix certificate issue as indicated in #595

neurorepro:


yarikoptic:
unfortunately we can't just do that -- Dockerfile is produced using neurodocker, as invoked from https://github.com/nipy/heudiconv/blob/master/utils/gen-docker-image.sh . See if you could set it to install specific version of `conda`. BTW -- in #596 I have not yet got to that error of certificates, but not sure if I reached that point ;)

neurorepro:
@yarikoptic i think i know how to do it with neurodocker, let me try quickly

neurorepro:
@yarikoptic done ! Luckily it was quite easy, setting the version of miniconda caused the conda update command to be skipped as you can see [here](https://github.com/ReproNim/neurodocker/blob/ad2f3a62afa9a8852bbf0ff3f1df1aadb3b74c8d/neurodocker/templates/miniconda.yaml#L44-L46)

yarikoptic:
Cool!
I have just merged my PR which also switched to newer dcm2niix.  If succeeds, I would prefer to stick to newer 3.9 - 3.7 will still be supported for 8 months but that isn't too long and just yesterday my issue known to be present in earlier pythons (iirc 3.7 included) was closed as "we know about the problem and it was addressed in 3.8", so - more anecdotal evidence to go forward with such a change.

With all that in mind -- would you mind updating/replacing this PR with the one which would fix conda installation into 3.9 instead of 3.7? I think it would be useful for robustness of our Dockerfile in the long run.

neurorepro:
Done ! I also tested the Dockerfile building with this PR and it works

neurorepro:
@yarikoptic Question though: shouldn't a dockerfile have fixed versions of dependencies, including for miniconda (i.e. not using latest but a set version as in this PR) ?

yarikoptic:
> Done ! I also tested the Dockerfile building with this PR and it works

cool, thanks!

> @yarikoptic Question though: shouldn't a dockerfile have fixed versions of packages and libraries, including for miniconda (i.e. not using `miniconda-latest` but a specific version such as `miniconda-py39_4.12.0` as in this PR) ?

"may be".  The problem is that complete listing of package versions is tedious  -- those which we list explicitly are just leaves, and they pull in bunch of other packages so versioning only our leaves is ... close to "no point".

In https://github.com/ReproNim/reproin/ I somewhat duplicate the effort done here with package building, but there I build relying on NeuroDebian, not conda, distribution: https://github.com/ReproNim/reproin/blob/master/generate_container.sh , and I use developed by us "nd_freeze" helper which freezes state of the Debian and NeuroDebian distribution to a specific date -- as a result with that single "version" specification I make build reproducible.

may be I should finally meld both approaches here in this repo (there is nothing "reproin" specific in reproin container), ... may be... some time

neurorepro:
@yarikoptic i think all your efforts of freezing packages, + setting packages version via conda, etc. shouldn't leave to many leaves unchecked, so i'd tend to be more optimistic :) or at least it leaves fewer to sweep when debugging irreproducible content...

PR #600, opened 2022-10-12
Set action step outputs via $GITHUB_OUTPUT

jwodder:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ for more information.


yarikoptic:
Thank you @jwodder!  current sweep of PRs shows that ideally the more WiP DataLad's `release-action` absorbs inside the better IMHO.

PR #607, opened 2022-11-03
Delete .dockerignore

### LINKED ISSUE #606: Gets unnecessary +dDATE added to a version corresponding to a clean release
```shell
smaug:~
$> docker run -it --rm --entrypoint bash -v /tmp/mount:/mnt:ro -v /tmp/mount/rw:/mnt/rw:rw nipy/heudiconv:v0.11.5 -c 'python -c "import heudiconv;print(heudiconv.__version__)"'
0.11.5+d20221103
```

I guess done by https://github.com/jwodder/versioningit . Not sure if it is a bug in versioningit or matter of some configuration, but ideally no `+d` would be added to a clean release version.

jwodder:
Closes #606.

Excluding version-controlled files from the Docker image causes the resulting copy of the repository inside the image to be dirty, which in turn results in "+dDATE" added to the version number.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,0 @@
-Dockerfile


yarikoptic:
oh, thank you!

PR #615, opened 2022-12-19
Add a bash anon-cmd to be used to incrementally anonymize sids

yarikoptic:
A simple bash script which would keep local ../.git/anon_sid_map.csv (assuming the script goes e.g. into code/ or some other subfolder like .heudiconv/), and could be used as anonymization script


yarikoptic:
I will just merge this goodie. Was using it in a study, seems to work nicely. Sure thing could be improved or made "built-in"

PR #619, opened 2023-01-25
Provide codespell config and workflow

yarikoptic:
we had fixed done before but never had CI to alert on adding new typos


yarikoptic:
remains green , let's proceed

PR #622, opened 2023-01-25
Duecredit dcm2niix

### LINKED ISSUE #132: duecredit support
Now that it is a package, would be easy to add basic duecredit support, with also entries for external tools we use (dcm2niix)

yarikoptic:
Closes #132 as we already have duecredit added, and just forgot to credit dcm2niix for all its hard work


yarikoptic:
Didn't break anything and closed an issue -- win win.

Issue #627, opened 2023-01-31
Test failures on `=heudiconv-0.11.6` (mainly `nipype.pipeline.engine.nodes.NodeExecutionError`)


TheChymera:
### Summary

[Here](https://ppb.chymera.eu/59b43c.log) is the full build and test log.
One error comes from datalad, on account of network sandboxing.
Any ideas what might be causing the nipype errors? @yarikoptic 

### Platform details:

- [x] Local environment, Gentoo Linux

TheChymera:
There also appears to be a test failing on account of `dcm2niix` not being present:

```
OSError: No command "dcm2niix" found on host localhost. Please check that the corresponding package is installed.
```

It doesn't seem to be listed among the deps, should I add it? I think it might have been omitted because it's used via nipype. But `dcm2niix` isn't a nipype dependency.

TheChymera:
Created https://github.com/nipy/heudiconv/pull/628 in case the comment above is an actual issue.

TheChymera:
for the other issues, a git identity needs to be declared in the sandbox, as seen e.g. in [team/heudiconv/-/blob/master/debian/rules#L11](https://salsa.debian.org/med-team/heudiconv/-/blob/master/debian/rules#L11)

TheChymera:
https://github.com/nipy/heudiconv/pull/631 fixes all remaining test issues, and might also help simplify the linked Debian build recipe.

PR #628, opened 2023-01-31
dcm2niix explicitly noted as a (PyPI) dependency and removed from being installed via apt-get etc

TheChymera:
Closes (part of) https://github.com/nipy/heudiconv/issues/627


yarikoptic:
FTR -- it was only recently that dcm2niix got "Python package" on pypi contributed: https://github.com/rordenlab/dcm2niix/pull/605/files  . What happens is that upon installation of the sources, it builds/installs the whl:

```
Building wheels for collected packages: dcm2niix
  Building wheel for dcm2niix (pyproject.toml) ... done
  Created wheel for dcm2niix: filename=dcm2niix-1.0.20220715-cp311-cp311-linux_x86_64.whl size=549462 sha256=336dea4b0cd2cef5754ed011566fb6f1d4792c9c056e6d595534ca9be6136609
  Stored in directory: /root/.cache/pip/wheels/36/8b/7d/26d61529a2df3157c9eecb9ea4ea4a45485fa28a1e9edae55f
```

I think it would only be great to become 'self pypi sufficient' so I do not see a reason to not do it. Do you see anything against it @satra @pvelasco and other @nipy/team-heudiconv ?

yarikoptic:
@TheChymera I have pushed a few more commits since we can not simplify some other aspects of installation/documentation. Please review and also check out more -- may be there is even more to it.

yarikoptic:
Let's give it a shot, thanks!

PR #631, opened 2023-02-02
Setting git author and email in test environment

TheChymera:
@yarikoptic as discussed on Tuesday. I tried a few different approaches, I think this is better than via a fixture since it ensures it will always be set throughout the test suite, without having to annotate individual tests.

### DIFF
--- a/heudiconv/tests/conftest.py
+++ b/heudiconv/tests/conftest.py
@@ -1,3 +1,8 @@
 import os
-os.environ["GIT_AUTHOR_EMAIL"] = "maxm@example.com"
-os.environ["GIT_AUTHOR_NAME"] = "Max Mustermann"
+import pytest
+@pytest.fixture(autouse=True, scope="session")
+def git_env():
+       os.environ["GIT_AUTHOR_EMAIL"] = "maxm@example.com"
+       os.environ["GIT_AUTHOR_NAME"] = "Max Mustermann"
+       os.environ["GIT_COMMITTER_EMAIL"] = "maxm@example.com"
+       os.environ["GIT_COMMITTER_NAME"] = "Max Mustermann"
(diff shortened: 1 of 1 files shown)


yarikoptic:
there will be no need to annotate individual tests if you make fixture session scope like the one I pointed you to in datalad: https://github.com/datalad/datalad/blob/maint/datalad/conftest.py#L27  pretty much just put those lines within such a function instead of module level since now if someone imports that module - that would be enough to ruin his day on his github dashboard while working within that python session ;)

heudiconv/tests/conftest.py
```diff
@@ -0,0 +1,5 @@
+import os
+os.environ["GIT_AUTHOR_EMAIL"] = "maxm@example.com"
+os.environ["GIT_AUTHOR_NAME"] = "Max Mustermann"
+os.environ["GIT_COMMITTER_EMAIL"] = "maxm@example.com"
+os.environ["GIT_COMMITTER_NAME"] = "Max Mustermann"
```
yarikoptic, line 5: no need to define both AUTHOR and COMMITTER. Just define AUTHOR -- then 2nd one would default to it AFAIK if not set explicitly.

TheChymera: nope, 2 tests still fail if I don't also set the committer, took me a while to figure that out as well...

https://ppb.chymera.eu/174ec2.log

TheChymera:
> pretty much just put those lines within such a function instead of module level since now if someone imports that module - that would be enough to ruin his day

Ok, will give the fixture another try, I think I missed the `autouse` argument.

heudiconv/tests/conftest.py
yarikoptic: 🤔 ok, let's see if fixture works

TheChymera:
@yarikoptic ok, it seems to work.

heudiconv/tests/conftest.py
TheChymera: it works :)

PR #637, opened 2023-02-15
add custom_info to group_dicoms_into_seqinfos call in parser

bpinsard:


yarikoptic:
are you trying to take over (which would be welcome to help bringing it over the finish line) the https://github.com/nipy/heudiconv/pull/581?

bpinsard:
Didn't meant to take over, just to influence the PR to include what makes it work for us. :-) 

Just added commits that should get the tests passing.
The main constraint for the custom_seqinfo is that is needs to be hashable:  that is likely be an error people using that function will run into, so it will need to be clearly documented. Or maybe should we add a check on the return value in the code with a more explicit error message?

yarikoptic:
> Didn't meant to take over,

if you can/want -- please do!

> The main constraint for the custom_seqinfo is that is needs to be hashable: that is likely be an error people using that function will run into, so it will need to be clearly documented. Or maybe should we add a check on the return value in the code with a more explicit error message?

Good point/aspect. The more checks to puke as early as possible the better instead of delaying inevitable crash with obscure error.

docs/heuristics.rst
```diff
@@ -86,6 +86,19 @@ or::
         ...
         return seqinfos  # ordered dict containing seqinfo objects: list of DICOMs
 
+---------------------------------------------------------------
+``custom_seqinfo(series_files, wrapper)``
+---------------------------------------------------------------
+If present this function will be called on eacg group of dicoms with
```
yarikoptic, line 92:
```suggestion
If present this function will be called on each group of dicoms with
```

we have `codespell` workflow which should pickup typos. If you rebase -- we better see it working on this PR too (for some reason it did not even though I think it was back 2 weeks ago when you started this PR)

yarikoptic: oh, this is PR to my PR -- I will take it on and then rebase my PR. Thanks!

yarikoptic:
I have fasfforwaded my branch to these changes (thanks @bpinsard ), then fixed up for my suggestions, rebased and force pushed in my PR (https://github.com/nipy/heudiconv/pull/581). because of rebase github did not figure out to close this PR. Let's continue from that newer state

PR #641, opened 2023-02-15
Pull request for contributing guide.

TheChymera:


yarikoptic:
as quickly discussed
- avoid duplicating info, so just point to info.py I guess where all are described
- note that it is .rst not .md
- mentioned that `pip install heudiconv[all]` would install with all dependencies for testing etc, and for local development should be `pip install -e .[all]` **but** `git-annex` installation would also be required for DataLad.  Refer to http://handbook.datalad.org/ installation section

README.rst
```diff
@@ -76,15 +76,7 @@ all relevant citations via `DueCredit <http://duecredit.org>`_.
...
-HeuDiConv requires installation of `dcm2niix`_ and optionally `DataLad`_.
-
-For development you will need a non-shallow clone (so there is a
-recent released tag) of the aforementioned repository. You can then
-install all necessary development requirements using ``pip install -r
-dev-requirements.txt``.  Testing is done using `pytest
-<https://docs.pytest.org/>`_.  Releases are packaged using Intuit
-auto.  Workflow for releases and preparation of Docker images is in
-``.github/workflows/release.yml``.
+For a detailed into, see our `contributing guide <CONTRIBUTING.rst>`_.
```
yarikoptic, line 79: please also include in documentations similarly to how we do for README.rst:

```
docs/index.rst:.. include:: ../README.rst

```

CONTRIBUTING.rst
```diff
@@ -0,0 +1,94 @@
...
+
+
+Additional Hints
+----------------
+
+It is recommended to check that your contribution complies with the following
+rules before submitting a pull request:
+
+* All public functions (i.e. functions whose name does not start with an underscore) should have
+  informative docstrings with sample usage presented as doctests when appropriate.
```
yarikoptic, line 88: please say that docstrings should be `numpy` style , point to https://numpydoc.readthedocs.io/en/latest/format.html

Also max line length - let's make 120?

yarikoptic:
coolio, thanks!

PR #642, opened 2023-02-15
Docker images: tag also as "unstable", strip "v" prefix, and avoid building in non-release workflow for releases.

yarikoptic:
We had on docker hub

image

so we were building also there. In consultation with @asmacdo we decided to just build on github actions since we have more control/logging (yet to establish con/tinuous I guess for the archive).  So I have disabled building on docker hub, and tuned the workflow so we provide continuity here.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,8 +67,9 @@ jobs:
         run: |
           docker build \
             -t nipy/heudiconv:master \
+            -t nipy/heudiconv:unstable \
             -t nipy/heudiconv:latest \
-            -t nipy/heudiconv:"$(git describe)" \
+            -t nipy/heudiconv:"$(git describe | sed -e 's,^v,,g')" \
             .
 
       - name: Push Docker images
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -67,8 +67,9 @@ jobs:
         run: |
           docker build \
             -t nipy/heudiconv:master \
+            -t nipy/heudiconv:unstable \
             -t nipy/heudiconv:latest \
-            -t nipy/heudiconv:"$(git describe)" \
+            -t nipy/heudiconv:"$(git describe | sed -e 's,^v,,g')" \
```
asmacdo, line 72:
```suggestion
            # remove "v" before version
            -t nipy/heudiconv:"$(git describe | sed -e 's,^v,,g')" \
```
optional nitpick

asmacdo (approved):
lgtm

.github/workflows/release.yml
yarikoptic: that comment might break that `bash` multiline invocation so I better not risk it ;-)

PR #644, opened 2023-02-15
Reword and correct punctuation on usage.rst

candleindark:


docs/usage.rst
```diff
@@ -2,7 +2,7 @@
 Usage
 =====
 
-``heudiconv`` processes DICOM files and converts the output into user defined
+``heudiconv`` processes DICOM files and structures the output into user defined
```
yarikoptic, line 5: I would keep converts here since structures doesn't hint on "conversion" part

yarikoptic:
```suggestion
``heudiconv`` processes DICOM files and converts the output into user defined
```

PR #654, opened 2023-03-29
Replace py.path with pathlib

jwodder:
The `py` library should not be used in modern code, and it's not even listed as dependency of heudiconv, so I don't believe the current code even works.


yarikoptic:
yeah -- that `monitor` code is somewhat stale/unmaintained/not-used , so indeed could be just full of such oddities. Thank you @jwodder

PR #658, opened 2023-04-10
Add & apply pre-commit and lint job

jwodder:


jwodder:
@nipy/team-heudiconv Please review this PR and comment any desired configuration changes.  I am using this as the base for several other PRs, so it would be good to get this PR processed quickly.

tox.ini
```diff
@@ -1,6 +1,34 @@
 [tox]
-envlist = py27,py33,py34,py35
+envlist = lint,py27,py33,py34,py35
```
asmacdo, line 2: These versions don't line up with our requirements. 

IMO something to fix separately, lmk if you'd like me to file an issue.

tox.ini
```diff
@@ -1,6 +1,34 @@
...
+commands =
+    flake8 heudiconv
+
+[flake8]
+doctests = True
+exclude = .*/,build/,dist/,test/data,venv/
+hang-closing = False
+unused-arguments-ignore-stub-functions = True
+select = A,B,B902,C,E,E242,F,U100,W
+ignore = A003,B005,E203,E262,E266,E501,W503
```
asmacdo, line 24: Nitpick (shouldn't block merge) 

IMO it's convenient to comment what these are.

asmacdo (approved):
Looks good to me!

tox.ini
jwodder: That's handled by #660.

asmacdo:
IMHO, I prefer a linter job to a precommit hook. (I haven't used them extensively, so please correct me if I'm missing something.)

The reason is that I sometimes want to commit and push a WIP (for example, to commit where I am so I can switch branches to reproduce a big) and I wouldn't want the linter failure to slow me down.

jwodder:
@asmacdo 

> The reason is that I sometimes want to commit and push a WIP (for example, to commit where I am so I can switch branches to reproduce a bug) and I wouldn't want the linter failure to slow me down.

That's what `git commit`'s `-n`/`--no-verify` option is for.

jwodder:
@yarikoptic It will cause conflicts, but that's only a problem for me.

PR #659, opened 2023-04-10
Don't call `logging.basicConfig()` in `__init__.py`

jwodder:
Calling `logging.basicConfig()` when your library is imported is incredibly rude, as it infringes upon the importing code's ability to configure the logging itself.  The only time (if any) the `heudiconv` package should call `logging.basicConfig()` is in the `main()` function for a command-line entry point.

See also <https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library>.

### DIFF
--- a/heudiconv/__init__.py
+++ b/heudiconv/__init__.py
@@ -1,16 +1,9 @@
-# set logger handler
 import logging
-import os
 
 from ._version import __version__
 from .info import __packagename__
 
 __all__ = ["__packagename__", "__version__"]
 
-# Rudimentary logging support.
 lgr = logging.getLogger(__name__)
-logging.basicConfig(
-    format="%(levelname)s: %(message)s",
-    level=getattr(logging, os.environ.get("HEUDICONV_LOG_LEVEL", "INFO")),
-)
 lgr.debug("Starting the abomination")  # just to "run-test" logging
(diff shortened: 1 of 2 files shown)


heudiconv/__init__.py
```diff
@@ -1,16 +1,9 @@
...
 from .info import __packagename__
 
 __all__ = ["__packagename__", "__version__"]
 
-# Rudimentary logging support.
 lgr = logging.getLogger(__name__)
-logging.basicConfig(
-    format="%(levelname)s: %(message)s",
-    level=getattr(logging, os.environ.get("HEUDICONV_LOG_LEVEL", "INFO")),
-)
```
yarikoptic, line 15: Please add it then to https://github.com/nipy/heudiconv/blob/master/heudiconv/cli/run.py

jwodder: @yarikoptic For some reason, `monitor.py` [already configures the logger for that module to use a different log format than the rest of heudiconv](https://github.com/nipy/heudiconv/blob/5e6c7ea6f2b5fb91a123fa52890a8c1d5e2cc725/heudiconv/cli/monitor.py#L25).  Should this be retained?

yarikoptic: heh, ok then, let's just proceed for now. Not yet sure if we ever get back to monitor to make it actually used.

PR #666, opened 2023-04-11
Make `test_private_csa_header` test write to temp dir

jwodder:
The `runner()` call in `test_private_csa_header` in `test_dicoms.py` did not specify an outdir, and so the test was creating files in the current directory, which is the repository root when testing locally with tox.  This PR fixes the test to instead write to a temporary directory.


yarikoptic:
Thank you! It remains green, so all good

PR #668, opened 2023-04-12
Use `pydicom.dcmread()` instead of `pydicom.read_file()`

jwodder:
`read_file()` was deprecated in pydicom v2.2.0 and will be removed in v3.0.  Code should instead use `dcmread()`, which has been an alias of `read_file()` since pydicom v1.0.0.


yarikoptic:
Thank you!

PR #669, opened 2023-04-17
ENH: set and expose global dicom compresslevel var

octomike:
Python's tarfile is using a compression level of 9 by default (back to and including python 3.5): https://docs.python.org/3/library/tarfile.html#tarfile.open

We see cases where this leads to a large waste of cpu cycles with minimal gain as opposed to compress level 5 or even 1. 

Setting a global `compresslevel` variable makes it easy to override from our heuristics, keeping the default with minimal code impact.

### DIFF
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -20,6 +20,7 @@ with warnings.catch_warnings():
 
 lgr = logging.getLogger(__name__)
 total_files = 0
+compresslevel = 9
 
 
 def create_seqinfo(mw, series_files, series_id):
@@ -459,7 +460,8 @@ def compress_dicoms(dicom_list, out_prefix, tempdirs, overwrite):
         try:
             if op.lexists(outtar):
                 os.unlink(outtar)
-            with tarfile.open(outtar, "w:gz", dereference=True) as tar:
+            with tarfile.open(outtar, "w:gz", compresslevel=compresslevel,
+                              dereference=True) as tar:
                 for filename in dicom_list:
                     outfile = op.join(tmpdir, op.basename(filename))
                     if not op.islink(outfile):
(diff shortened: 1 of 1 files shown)


heudiconv/dicoms.py
```diff
@@ -20,6 +20,7 @@
 
 lgr = logging.getLogger(__name__)
 total_files = 0
+compresslevel = 9
```
yarikoptic, line 23: ha -- a cute workaround!  I don't mind accepting.

I only worry that eventually it can get moved thus breaking this "feature" for you unless we 
1. at least annotate it here
```suggestion
# Might be monkey patched by user heuristic to tune desired compression level
# Preferably do not move/rename unless 
compresslevel = 9
```
2. in your heuristic you better first check `if hasattr(dicoms, compresslevel)` before assigning or someone moves it away -- you might be out of luck.

sure thing this is just a quite an ad-hoc workaround. another one could have been for you to just overload `tarfile.open` and provide your custom compresslevel there... smth like

```
import tarfile

def _tarfile_open(*args, **kwargs):
    kwargs.set_default('compresslevel', 3)
    return tarfile.open(*args, **kwargs)

tarfile.open = _tarfile_open
```
or alike.  Just an idea

heudiconv/dicoms.py
```diff
@@ -20,6 +20,9 @@
 
 lgr = logging.getLogger(__name__)
 total_files = 0
+# Might be monkey patched by user heuristic to tune desired compression level
+# Preferably do not move/rename unless
```
yarikoptic, line 24:
```suggestion
# Preferably do not move/rename unless
```

yarikoptic: hey github -- where is merge button???

PR #671, opened 2023-05-01
Declare `custom_grouping` return type instead of casting

jwodder:
This has the added benefit that mypy will error if a callable with a different return type is passed as `custom_grouping`.

### DIFF
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -9,17 +9,7 @@ import os.path as op
 from pathlib import Path
 import sys
 import tarfile
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Dict,
-    List,
-    NamedTuple,
-    Optional,
-    Union,
-    cast,
-    overload,
-)
+from typing import TYPE_CHECKING, Any, Dict, List, NamedTuple, Optional, Union, overload
 from unittest.mock import patch
 import warnings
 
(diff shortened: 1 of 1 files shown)


heudiconv/dicoms.py
```diff
@@ -202,7 +204,8 @@ def group_dicoms_into_seqinfos(
     flatten: Literal[False, True] = False,
     custom_grouping: str
     | Callable[
-        [list[str], Optional[Callable[[dcm.dataset.Dataset], Any]], type[SeqInfo]], Any
+        [list[str], Optional[Callable[[dcm.dataset.Dataset], Any]], type[SeqInfo]],
+        dict[SeqInfo, list[str]],
```
yarikoptic, line 208: grr... elderly me fails to find where I have added some type defined for another callback in heudiconv (may be I never finished/submitted that PR) following the https://mypy.readthedocs.io/en/stable/protocols.html#callback-protocols  . Because this signature is quite long - I wonder if we should just define `CustomGrouppingT` or alike to be used in all instances here, WDYT?

jwodder: Unless we're going to use this type outside of `group_dicoms_into_seqinfos()`, I don't see the need for an alias.

yarikoptic:
sorry, now we got a conflict (`|` to `Union` change I guess) - would you be kind to refresh and feel welcome to merge when passes or I will do (unless forget again ;))

jwodder:
@yarikoptic Conflict resolved.

Issue #667, opened 2023-04-12
pydicom warnings encountered during `test_b0dwi_for_fmap`


jwodder:
When the `test_b0dwi_for_fmap` test in `heudiconv/tests/test_convert.py` is run, pydicom emits two UserWarnings with the message "The value length (22) exceeds the maximum length of 16 allowed for VR SH." (presumably because one or more files in `heudiconv/tests/data/b0dwiForFmap/` are ill-formed), and then pytest's `filterwarnings` setting causes these warnings to be converted into errors that are caught & logged [here](https://github.com/nipy/heudiconv/blob/2c6d228516dda41cdc9f91e7bea51b3acdb0f5b7/heudiconv/dicoms.py#L612) — and yet, the test does not fail.

For the record, the output from running just this test as of commit 2c6d228 is:

<details>

```
============================= test session starts ==============================
platform darwin -- Python 3.11.3, pytest-7.3.0, pluggy-1.0.0 -- /Users/jwodder/dartmouth/heudiconv/.tox/py3/bin/python
cachedir: .tox/py3/.pytest_cache
rootdir: /Users/jwodder/dartmouth/heudiconv
configfile: tox.ini
collecting ... collected 120 items / 119 deselected / 1 selected

heudiconv/tests/test_convert.py::test_b0dwi_for_fmap WARNING: Heuristic is missing an `infotoids` method, assigning empty method and using provided subject id b0dwiForFmap. Provide `session` and `locator` fields for best results.
230412-10:38:52,754 nipype.workflow INFO:
	 [Node] Setting-up "convert" in "/private/var/folders/l7/wrkq93d133d8zpn36fmqrq0r0000gn/T/dcm2niix25vaxluk/convert".
INFO: [Node] Setting-up "convert" in "/private/var/folders/l7/wrkq93d133d8zpn36fmqrq0r0000gn/T/dcm2niix25vaxluk/convert".
230412-10:38:52,762 nipype.workflow INFO:
	 [Node] Executing "convert" <nipype.interfaces.dcm2nii.Dcm2niix>
INFO: [Node] Executing "convert" <nipype.interfaces.dcm2nii.Dcm2niix>
230412-10:38:52,949 nipype.interface INFO:
	 stdout 2023-04-12T10:38:52.949521:Compression will be faster with 'pigz' installed http://macappstore.org/pigz/
INFO: stdout 2023-04-12T10:38:52.949521:Compression will be faster with 'pigz' installed http://macappstore.org/pigz/
[...]

PR #673, opened 2023-05-05
Make an `embed_dicom_and_nifti_metadata()` annotation work on Python 3.7

### LINKED ISSUE #672: typing related error in metadata embedding.
<!-- DO NOT DELETE THIS!
This template is used to facilitate issue resolution.
All text in <!-> tags will not be displayed.
-->

### Summary
<!-- If you are having conversion troubles, please share as much
relevant information as possible. This includes, but is not limited to:
- log of conversion
- heuristic
-->
Updating heudiconv from master branch, I have the following error showing up which seems related to #656  changes. 


```
230505-14:27:39,460 nipype.workflow WARNING:                                                                                                              
         [Node] Error on "embedder" (/scratch/tmp/embedmetab5y9u753/embedder)                                                                             
Embedding failed: Exception raised while executing Node embedder.                                                                                         
                                                                                                                                                          
  File "/data/neuromod/virtualenvs/neuromod_data/lib/python3.7/site-packages/nipype/utils/functions.py", line 34, in create_function_from_source          
    exec(function_source, ns)                                                                                                                             
  File "<string>", line 4, in <module>                                                                                                                    
TypeError: unsupported operand type(s) for |: 'type' and 'type'                                                                                           
                                                                                                                                                          
The above exception was the direct cause of the following exception:                                                                                      
                                                                                                                                                          
  File "/data/neuromod/virtualenvs/neuromod_data/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 398, in run                            
    runtime = self._run_interface(runtime)                                                                                                                
  File "/data/neuromod/virtualenvs/neuromod_data/lib/python3.7/site-packages/nipype/interfaces/utility/wrappers.py", line 133, in _run_interface          
    self.inputs.function_str, self.imports                                                                                                                
  File "/data/neuromod/virtualenvs/neuromod_data/lib/python3.7/site-packages/nipype/utils/functions.py", line 44, in create_function_from_source          
    raise RuntimeError(msg) from e                                                                                                                        
RuntimeError: Error executing function                                                                                                                    
def embed_dicom_and_nifti_metadata(                                                                                                                       
    dcmfiles: List[str],                                                                                                                                  
    niftifile: str,                                                                                                                                       
    infofile: str | Path,                                                                                                                                 
    bids_info: Optional[Dict[str, Any]],                                                                                                                  
) -> None:                                                                                                                                                
    """Embed metadata from nifti (affine etc) and dicoms into infofile (json)                                                                             
                                                                                                                                                          
    `niftifile` should exist. Its affine's orientation information is used while                                                                          
    establishing new `NiftiImage` out of dicom stack and together with `bids_info`                                                                        
    (if provided) is dumped into json `infofile`                                                                                                          
                                                                                                                                                          
    Parameters                                                                                                                                            
    ----------                                                                                                                                            
    dcmfiles                                                                                                                                              
    niftifile                                                                                                                                             
    infofile                                                                                                                                              
    bids_info: dict                                                                                                                                       
      Additional metadata to be embedded. `infofile` is overwritten if exists,                                                                            
      so here you could pass some metadata which would overload (at the first                                                                             
      level of the dict structure, no recursive fancy updates) what is obtained                                                                           
      from nifti and dicoms                                                                                                                               
                                                                                                                                                          
    """                                                                                                                                                   
    # These imports need to be within the body of the function so that they                                                                               
    # will be available when executed by nipype:                                                                                                          
    import json                                                                                                                                           
    import os.path                                                                
...
...
```

### Platform details:

Choose one:
- [x] Local environment: ubuntu-22.04 python3.7.16
<!-- If selected, please provide OS and python version -->
- [ ] Container
<!-- If selected, please provide container name and tag"-->

- Heudiconv version:
<!-- To check: run heudiconv with just the --version flag -->
`0.5.3.post808+g6cc9f29`

jwodder:
Fixes #672.

### DIFF
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -9,7 +9,17 @@ import os.path as op
 from pathlib import Path
 import sys
 import tarfile
-from typing import TYPE_CHECKING, Any, Dict, List, NamedTuple, Optional, cast, overload
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    List,
+    NamedTuple,
+    Optional,
+    Union,
+    cast,
+    overload,
+)
 from unittest.mock import patch
 import warnings
 
@@ -264,8 +274,9 @@ def group_dicoms_into_seqinfos(
         if custom_grouping is None:
             raise RuntimeError("Custom grouping is not defined in heuristic")
         if callable(custom_grouping):
-            return cast(Dict[SeqInfo, List[str]],
-                        custom_grouping(files, dcmfilter, SeqInfo))
+            return cast(
+                Dict[SeqInfo, List[str]], custom_grouping(files, dcmfilter, SeqInfo)
+            )
         grouping = custom_grouping
         study_customgroup = None
 
@@ -541,8 +552,9 @@ def compress_dicoms(
         try:
             if op.lexists(outtar):
                 os.unlink(outtar)
-            with tarfile.open(outtar, "w:gz", compresslevel=compresslevel,
-                              dereference=True) as tar:
+            with tarfile.open(
+                outtar, "w:gz", compresslevel=compresslevel, dereference=True
+            ) as tar:
                 for filename in dicom_list:
                     outfile = op.join(tmpdir, op.basename(filename))
                     if not op.islink(outfile):
@@ -570,7 +582,7 @@ def compress_dicoms(
 def embed_dicom_and_nifti_metadata(
     dcmfiles: List[str],
     niftifile: str,
-    infofile: str | Path,
+    infofile: Union[str, Path],
     bids_info: Optional[Dict[str, Any]],
 ) -> None:
     """Embed metadata from nifti (affine etc) and dicoms into infofile (json)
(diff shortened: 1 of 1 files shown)


heudiconv/dicoms.py
```diff
@@ -570,7 +582,7 @@ def _assign_dicom_time(ti: tarfile.TarInfo) -> tarfile.TarInfo:
 def embed_dicom_and_nifti_metadata(
     dcmfiles: List[str],
     niftifile: str,
-    infofile: str | Path,
+    infofile: Union[str, Path],
```
yarikoptic, line 585: how come we had tests passed on python 3.7? we do have `from __future__ import annotations` on top

jwodder: * Read the comment above `embed_dicom_and_nifti_metadata()` about why `from __future__ ...` in our source file is irrelevant.

* I'm not sure why the tests passed, but I suspect it may be related to #667.

yarikoptic:
ok, let's proceed since indeed the comment already describes the reason.

PR #680, opened 2023-05-23
Check if acq_time has ms and add .0 if not

crcox:
* In python 3.11.2, `strptime_micr(acq_date + acq_time, "%Y%m%d%H%M%S[.%f]")` does not make ms optional
* A simple solution to make the code work regardless of whether ms are recorded or not is:
```
if not '.' in acq_time:
    acq_time = acq_time + '.0'
```

### DIFF
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -487,6 +487,9 @@ def get_datetime_from_dcm(dcm_data: dcm.FileDataset) -> Optional[datetime.dateti
     """
     acq_date = dcm_data.get("AcquisitionDate")
     acq_time = dcm_data.get("AcquisitionTime")
+    if not '.' in acq_time:
+        acq_time = acq_time + '.0'
+
     if not (acq_date is None or acq_time is None):
         return strptime_micr(acq_date + acq_time, "%Y%m%d%H%M%S[.%f]")
 


yarikoptic:
we already do that since today's release 0.13.1 at https://github.com/nipy/heudiconv/blob/719ad9d4688e625643d01766653789abdf7bc5ca/heudiconv/utils.py#L669:5 

did you find it not working ?

yarikoptic:
sorry, I misread your report -- you are already stating that our helper doesn't do what it supposed to ... I will check later in detail.

yarikoptic:
> * In python 3.11.2, `strptime_micr(acq_date + acq_time, "%Y%m%d%H%M%S[.%f]")` does not make ms optional

why exactly you think so, i.e. do you have an example to demonstrate that? 

from me reading the code [here](https://github.com/nipy/heudiconv/blob/719ad9d4688e625643d01766653789abdf7bc5ca/heudiconv/utils.py#LL685C33-L685C44) we add `%f` only if provided date has ms and otherwise uses the pattern to match without `%f` so I do think this patch is needed.  FWIW [here](https://github.com/nipy/heudiconv/blob/719ad9d4688e625643d01766653789abdf7bc5ca/heudiconv/tests/test_utils.py#L173) is the test which tests this code

crcox:
Two apologies: I must not have tested appropriately--your test proves that it works. Maybe I was not using the version of Python that I thought I was.

I can say for certain that when running the main script, the code would crash when calling this function; manually revising the `fmt` string resolved the problem, as did my solution above. So really, the only thing that makes sense is I was not running the version of Python I thought I was.

Second apology is for the slow reply.

yarikoptic:
so some mystery remains -- thus please reopen or file a new one if you somehow see this issue to come up.

PR #681, opened 2023-06-01
Add description of placeholders which could be used in the produced templates

yarikoptic:
Inspired (and found to be missing in the docs) by https://github.com/nipy/heudiconv-joss-paper/pull/31/files#diff-0403ef06adf405f7b310b4518bd6a3559854f54c61676f676ce9cbfee7172ab6R173

### DIFF
--- a/docs/heuristics.rst
+++ b/docs/heuristics.rst
@@ -28,9 +28,23 @@ Conversion outputs are defined as keys, a `tuple` consisting of three elements:
 - `None` - a historical artifact (corresponds to some notion of
   ``annotation_class`` no living human is aware about)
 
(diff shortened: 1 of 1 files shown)


yarikoptic:
hm, RTD did not report to the status although it built it fine: https://heudiconv--681.org.readthedocs.build/en/681/heuristics.html . So I will proceed with merging

yarikoptic:
aha -- I followed the https://docs.readthedocs.io/en/stable/guides/pull-requests.html#troubleshooting which lead me to my apps permissions page and apparently for RTD, for `nipy` organization there were no permissions. So I granted them. Hopefully next PR would report back on building docs

PR #682, opened 2023-06-07
Fix inconsistent behavior of existing session when using -d compared to --files option: raise an AssertionError instead of just a warning

neurorepro:
This is described in #683

### DIFF
--- a/heudiconv/parser.py
+++ b/heudiconv/parser.py
@@ -278,15 +278,10 @@ def get_study_sessions(
             )
             lgr.info("Study session for %r", study_session_info)
 
-            if study_session_info in study_sessions:
-                if grouping != "all":
-                    # MG - should this blow up to mimic -d invocation?
-                    lgr.warning(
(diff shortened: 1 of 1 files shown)


heudiconv/parser.py
```diff
@@ -278,15 +278,9 @@ def infotoids(
...
-                        "Existing study session with the same values (%r)."
-                        " Skipping DICOMS %s",
-                        study_session_info,
-                        seqinfo.values(),
-                    )
-                    continue
+            if grouping != "all":
+                assert (study_session_info not in study_sessions), (
+                    f"Existing study session {study_session_info} "
+                    f"already in analyzed sessions {study_sessions.keys()}")
```
yarikoptic, line 284: I have checked 1366 heudiconv.log files from conversions and found none with "Existing study session with the same values". So I think at least I should be safe ;) let's see how error treats us.

yarikoptic:
change of behavior so adding `minor` label instead of just a `patch`

Issue #624, opened 2023-01-26
Fieldmaps acquired during each part of a session splits into two not differentiated


neurorepro:
It is relatively common to have a single theoretical session splits practically into two sessions due to the length of the protocol, so e.g. `anat`, `task 1`, `task 2`, `fmap` in the morning and `DWI`, `task 3`, `fmap` in the afternoon.

heudiconv can deal with this with the option `-g all` to group all files no matter the series UID which changes after each practical session and would give an error otherwise.

The issue I am having is that when the grouping is applied to a dataset with a fieldmap having the same series id (e.g. `2_myfieldmap`) during each practical session, the two fieldmaps are not differentiated as can be seen in `dicominfo.tsv`. The resulting detected fieldmap magnitude ('M' in protocol type) has then twice the number of files, and same for the phase, resulting in this kind of output for conversion:
```
│       │   ├── sub-s01_ses-01_magnitude1.json
│       │   ├── sub-s01_ses-01_magnitude1.nii.gz
│       │   ├── sub-s01_ses-01_magnitude2.json
│       │   ├── sub-s01_ses-01_magnitude2.nii.gz
│       │   ├── sub-s01_ses-01_magnitude3.json
│       │   ├── sub-s01_ses-01_magnitude3.nii.gz
│       │   ├── sub-s01_ses-01_magnitude4.json
│       │   ├── sub-s01_ses-01_magnitude4.nii.gz
│       │   ├── sub-s01_ses-01_phasediff1.json
│       │   ├── sub-s01_ses-01_phasediff1.nii.gz
│       │   ├── sub-s01_ses-01_phasediff2.json
│       │   └── sub-s01_ses-01_phasediff2.nii.gz
```

Would there be any suggestion on how to tackle this issue ? 

Would the only solution be to convert each practical session (i.e. with a unique study UID) separately and group them later in a single theoretical session ? This would be quite cumbersome and a shame considering the great `-g` option normally made for this kind of case.

yarikoptic:
> The issue I am having is that when the grouping is applied to a dataset with a fieldmap having the same series id (e.g. `2_myfieldmap`) during each practical session, the two fieldmaps are not differentiated as can be seen in `dicominfo.tsv`.

[...]

PR #684, opened 2023-06-07
Seriesgrouping

neurorepro:


neurorepro:
@yarikoptic As discussed in #624 I implemented a hash to shorten the name, but the full Series UID could be used too

yarikoptic:
In general I think the approach makes sense in general.  Re why not before: I guess (@satra and @mih might correct if have any recoll of the motivations) is that series id and name (instead of UID) were chosen for human accessibility since they make it easier to understand which particular series in question while annotating `.edit.txt` and just debugging the operation.  So far I do not see immediate problem with adding series UID since AFAIK indeed it should be the same for all files within a single acquisition.  Might be worth checking though on more of combined ones like T1PD etc. 

TL;DR: we need to approach it with caution and thus need more checking/work. Some reservations/concerns and thinking out loud:

- the fact that we ran into it just now suggests that **situation is not typical thus questions the change of the default behavior**: 
   - It seems it would apply ONLY in the case of `-g all` or `-g accession` i.e. where we force multiple scanning sessions into a single conversion study.  That would partially explain why we have not ran into it I guess. Such grouppings were added only few years back IIRC and aren't default mode of operation
   - as you note in https://github.com/nipy/heudiconv/issues/624#issuecomment-1406097247 `dcm2niix` would still do "the right thing" and convert those incorrectly "groupped" files into different files since it would rely on series UUID. So it is just that neither a heuristic nor heudiconv would know how to tease them apart.
- it **would introduce the core change of behavior** 
  - possibly making it then impossible for `heudiconv` to operate on prior produced annotations in `-c none` mode for the subsequent conversion with `-c dcm2niix` since all groups names would change. Needs a check:
  - may be `heudiconv` needs to become smart and detect in the mode of work while reloading already existing mapping file and then not bother adding UID? might help in transition period but generally not needed/too cumbersome... may be we would just add a check while loading an existing mapping that all the series ids do include that UID hash and error out if they don't with informative message.  I think the latter is better.
  - but it also might break some heuristics, which might have relied on those series ids (did any which we ship? needs to be checked)
  - it could also be made optional - either via command line option (more preferable) or automagicly enabled only in `-g all` or `-g accession` (I don't like that).  But then it adds differences in behavior based on CLI options, not good

neurorepro:
@yarikoptic thank you very much for the feedback

Regarding scope of use case: i think it is actually common to have artificially split a single BIDS session into two when the scanning protocol is too long (e.g. AM: anat, resting state, ... PM: diffusion MRI, ...) or the subject needs to exit. It is actually mentioned explicitely in the [BIDS definition of session (def number 5)](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#definitions):

> if a subject has to leave the scanner room and then be re-positioned on the scanner bed, the set of MRI acquisitions will still be considered as a session and match sessions acquired in other subjects. Similarly, in situations where different data types are obtained over several visits (for example fMRI on one day followed by DWI the day after) those can be grouped in one session.

The fact that not many people reported it i think is because users did not know what went wrong and would not think this may be due to heudiconv (who would think of blaming such a magnificent tool ?)

Regarding change of behavior i totally agree that this should be done carefully. I think what seems to be your preferred option would be best:

> may be we would just add a check while loading an existing mapping that all the series ids do include that UID hash and error out if they don't with informative message. I think the latter is better.

I will update the PR with that implemented.

If the error is too much for users (for which the use case of needing to keep previous data without uid is justified) then maybe we could add later a temporary option `--with-deprecated-series-id`  ?

neurorepro:
@yarikoptic I modified the code to check that the conversion table does not use deprecated series IDs. Otherwise it recomputes the conversation table with the non-deprecated series IDs.

I also set the hash length to 8 as you suggested.

neurorepro:
Hello @yarikoptic does it look like the code is ok now ? Cheers !

heudiconv/utils.py
```diff
@@ -222,6 +223,41 @@ def load_json(filename: str | Path, retry: int = 0) -> Any:
...
+        return True
+    # If all previous tests passed, then the series_id is not deprecated
+    return False
+
+
+def has_deprecated_seriesid(info_file: str) -> bool:
+    """Return True if any series_id in the info_file is in deprecated format."""
+    info = read_config(info_file)
+    series_ids = [series_id for sublist in info.values() for series_id in sublist]
+    if any(is_deprecated_seriesid(series_id) for series_id in series_ids):
```
yarikoptic, line 251: Could be shorter 
```suggestion
    if any(map(is_deprecated_seriesid, series_ids)):
```
but it is ok either way.

heudiconv/utils.py
```diff
@@ -222,6 +223,41 @@ def load_json(filename: str | Path, retry: int = 0) -> Any:
...
+        lgr.warning(
+            "Found deprecated series identifier in info file in the format"
+            "<series num>-<protocol name> instead of <series num>-<protocol name>-<UID hash>"
```
yarikoptic, line 254:
```suggestion
            "Found deprecated series identifier in info file in the format "
            "<series num>-<protocol name> instead of <series num>-<protocol name>-<UID hash>. "
```

heudiconv/utils.py
```diff
@@ -222,6 +223,41 @@ def load_json(filename: str | Path, retry: int = 0) -> Any:
...
+    series_number = series_id.split('-')[0]
+    if not series_number.isdigit():
+        return True
+    # Check the last part of the series_id is a hash in hexadecimal format of length len_hash_hex
+    hash_hex = series_id.split('-')[-1]
+    hex_digits = set(string.hexdigits)
+    if len(hash_hex) != len_hash_hex or not all(c in hex_digits for c in hash_hex):
+        return True
+    # If all previous tests passed, then the series_id is not deprecated
+    return False
```
yarikoptic, line 244:
```suggestion
    return len(hash_hex) != len_hash_hex or not all(c in hex_digits for c in hash_hex)
```

yarikoptic: or may be better to use `re` (might need an import)

```suggestion
    return re.match(f"[{string.hexdigits}]{{{len_hash_hex}}}$")
```

this function needs a unittest which would be trivial to do

yarikoptic, line 244: overall can be

```suggestion
    return not re.match(f"[{string.digits}]+-.+-[{string.hexdigits}]{{{len_hash_hex}}}$")
```
to just match regex... again -- just needs a unittest

heudiconv/utils.py
```diff
@@ -222,6 +223,41 @@ def load_json(filename: str | Path, retry: int = 0) -> Any:
...
+    A non-deprecated series ID is in the format "series_number-protocol_name-seriesUID_hash_hex"
+    """
+    # Check at least two '-' in the series_id
+    if series_id.count('-') <= 1:
+        return True
+    # Check the first part of the series_id is a number
+    series_number = series_id.split('-')[0]
+    if not series_number.isdigit():
+        return True
+    # Check the last part of the series_id is a hash in hexadecimal format of length len_hash_hex
```
yarikoptic, line 238:
```suggestion
    # Check the last part of the series_id is a hash in hexadecimal format of length len_hash_hex
    # Note that protocol_name could contain -
```

is that correct and that is why `[-1]`?

yarikoptic (changes requested):
> @yarikoptic I modified the code to check that the conversion table does not use deprecated series IDs. Otherwise it recomputes the conversation table with the non-deprecated series IDs.

I worry that it is a bit too intrusive, in my [comment above](https://github.com/nipy/heudiconv/pull/684#issuecomment-1584635957) I thought it might be better to be more polite and not to just abolish the work people might have done already by just saying that "now we need another way". But Ok, I guess I can swallow this, we would just announce some "major"ish release.

But one major concern about the PR is absent "setting in stone" for such new behaviors in tests: not a single test was changed or added, thus suggesting actual problem of us not bothering to test that behavior before.  Could you please add/change some test so it does get duplicated seqinfo (I guess we could just copy some subfolder with dicoms into a new one thus causing duplicates) and seeing it fail?

I also left some comments on how to possibly shorten that function to determine if outdated using `re` -- can become just 1 liner.  But we would need to have a test either way.

yarikoptic:
@neurorepro ping on above comments.

neurorepro:
Yes totally for the tests. I actually tested it on two datasets (already having old data, and newly converted) and it went fine. But indeed tests need to be included for continuous testing.

I'll get back to you!

yarikoptic:
a gentle ping

PR #701, opened 2023-08-30
Implement run+? which would make increment within that sequence

yarikoptic:
request by @dnkennedy

TODOs
- [ ] make sure does not break existing
- [ ] add tests for +? mode


dnkennedy:
Is there something I should do regarding testing this?

PR #713, opened 2023-10-09
added asl-support content in utils.py and dicoms.py files

pbbaba:


heudiconv/dicoms.py
```diff
@@ -109,6 +109,11 @@ def create_seqinfo(mw: dw.Wrapper, series_files: list[str], series_id: str) -> S
         date=dcminfo.get("AcquisitionDate"),
         series_uid=dcminfo.get("SeriesInstanceUID"),
         time=dcminfo.get("AcquisitionTime"),
+        # ASL support
+        in_plane_phase_encoding_direction = dcminfo.get("InPlanePhaseEncodingDirection") if dcminfo.get("InPlanePhaseEncodingDirection") else None,
```
yarikoptic, line 113: just
```suggestion
        in_plane_phase_encoding_direction = dcminfo.get("InPlanePhaseEncodingDirection"),
```
and so on would be sufficient.  But first please elaborate why these particular are necessary for ASL support?
Note: we do not really want to extend list of extracted fields to get unnecessarily long.

NB I should really look into finishing https://github.com/nipy/heudiconv/pull/581 some days soon to provide means for easy extraction of any metadata desired.

pbbaba:
I added the `if..else` condition so as to avoid the code break if any dicom image does not contain these headers. It did break for 'Image Orientation Patient' tag. 

I added these extra tags so as to make them accessible for naming the `perf` files during the BIDS conversion.

yarikoptic:
> I added the `if..else` condition so as to avoid the code break if any dicom image does not contain these headers. It did break for 'Image Orientation Patient' tag.

`.get("key")` should  work and default to `None` if there is no "key".  So code should not break and if..else hence not needed

yarikoptic:
ping @pbbaba - do you have time to tune up the PR following the review?

PR #728, opened 2024-01-11
properly remove GE multiecho bvals/bvecs

bpinsard:
GE Hyperband (at least some versions of it) multi-echo data get diffusion vector/values in the dicom tags used by dcm2niix to generate bvecs/bvals. 

While heudiconv is properly coded to remove such data when there is a single bvec/bval per series, multi-echo data outputs as many as there are echoes causing it to crash.

```
An error occur processing s398
expected str, bytes or os.PathLike object, not TraitListObject
  File "/app/convert.py", line 54, in convert_pipeline
    heudiconv_workflow(**heudiconv_params)
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/main.py", line 463, in workflow
    prep_conversion(
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/convert.py", line 258, in prep_conversion
    convert(
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/convert.py", line 623, in convert
    bids_outfiles = save_converted_files(
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/convert.py", line 891, in save_converted_files
    if bvals_are_zero(res.outputs.bvals):
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/convert.py", line 1083, in bvals_are_zero
    with open(bval_file) as f:
TypeError: expected str, bytes or os.PathLike object, not TraitListObject
```

The current PR fixes that by checking if all files only contain zeros bvals when not suffixed `_dwi`, and remove all of them appropriately.

FYI @miltoncamachocamacho


heudiconv/convert.py
```diff
@@ -1077,6 +1083,10 @@ def bvals_are_zero(bval_file: str) -> bool:
     True if all are all 0 or 5; False otherwise.
     """
 
+    # GE hyperband multi-echo containing diffusion info
+    if isinstance(bval_file, TraitListObject):
+        return all([bvals_are_zero(bvf) for bvf in bval_file])
```
yarikoptic, line 1088:
```suggestion
        return all(bvals_are_zero(bvf) for bvf in bval_file)
```
so we deal with generator and quit as soon as can . another alternative

```suggestion
        return all(map(bvals_are_zero, bval_file))
```

heudiconv/convert.py
```diff
@@ -1077,6 +1083,10 @@ def bvals_are_zero(bval_file: str) -> bool:
     True if all are all 0 or 5; False otherwise.
     """
 
+    # GE hyperband multi-echo containing diffusion info
+    if isinstance(bval_file, TraitListObject):
```
yarikoptic, line 1087: note that we have `: str` types annotation for `bval_file`.  Here the premise is that it could be a list (like) as well.  We need to probably cast to regular list as soon as we get value there and adjust types annotations.

Also it would be great to gain some unittest on this aspect, otherwise this remains an untested and thus likely to break code.

PR #729, opened 2024-01-12
rst paths must be relative to the docs/ dir

asmacdo:


README.rst
```diff
@@ -53,7 +53,7 @@ into structured directory layouts.
 
 Heudiconv can be inserted into your workflow to provide automatic conversion as part of a data acquisition pipeline, as seen in the figure below:
 
-.. image:: figs/environment.png
+.. image:: ../figs/environment.png
```
yarikoptic, line 56: nope -- it is in the root so cannot `../`. For the documentation where we link via `../README.rst` which I guess causes the issue you are trying to solve... but there there is a symlink

```
❯ ls -ld docs/figs
lrwxrwxrwx 1 yoh yoh 8 Jan 12 17:30 docs/figs -> ../figs//

```

so likely should just work.

So what problems are you trying to solve???

asmacdo: This was my initial investigation into what was broken about the docs build. You're correct, this broke the root README.

Christian fixed this another way today, so I'll close

PR #739, opened 2024-02-20
datalad sensitive marking fixes

bpinsard:
# Mark sensitive files added in the commit not all globbed.

Currently all globbed files in the current tree get marked as sensitive which could overwrite previously removed metas (eg after defacing), it now filter file in the commit where converted files are added.

# Add metadata rather than init to fix reruns issue.

If one runs heudiconv on a branch, remove the metadata (eg after defacing) then remove that branch and then rerun the same conversion on the same dataset, it seems like the new files with same names inherit the absence of metadata from the previous files (from git-annex branch). This is unexpected as the git-annex hash should differ so `.met` file should not be picked-up, but there is some git-annex black magic going on here. The issue is that sensitive files do not get marked as non-sensitive.

### DIFF
--- a/heudiconv/external/dlad.py
+++ b/heudiconv/external/dlad.py
@@ -152,7 +152,7 @@ def add_to_datalad(
         # not in effect! ?
         # annex_add_opts=['--include-dotfiles']
     )
-    annexed_files = [sr["path"] for sr in save_res if sr["key"]]
+    annexed_files = [sr["path"] for sr in save_res if sr.get("key", None)]
 
     # Provide metadata for sensitive information
     sensitive_patterns = [
--- a/heudiconv/external/dlad.py
+++ b/heudiconv/external/dlad.py
@@ -146,23 +146,27 @@ def add_to_datalad(
                 message="Added gitattributes to place all .heudiconv content"
                 " under annex",
             )
-    ds.save(
+    save_res = ds.save(
         ".",
         recursive=True
         # not in effect! ?
         # annex_add_opts=['--include-dotfiles']
     )
+    annexed_files = [sr["path"] for sr in save_res if sr["key"]]
 
     # Provide metadata for sensitive information
-    last_commit = "HEAD"
-    mark_sensitive(ds, "sourcedata", last_commit)
-    mark_sensitive(ds, "*_scans.tsv", last_commit)  # top level
-    mark_sensitive(ds, "*/*_scans.tsv", last_commit)  # within subj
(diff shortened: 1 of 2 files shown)


yarikoptic (changes requested):
good idea, thank you!

But I think we can avoid referring to last_commit since we are not even 100% certain it has happened just now, and we can explicitly have the list of just committed files ;-)  

Above in a call to `ds.save` - get the results and go through them -- they should include all files which were add'/committ'ed .  Eg.

```shell
❯ rm existed && echo "replaced" >| existed; touch newfile; python -c "import datalad.api as dl; print(dl.save('.'))"
<frozen importlib._bootstrap>:1047: ImportWarning: _SixMetaPathImporter.find_spec() not found; falling back to find_module()
add(ok): newfile (file)                                                                                                                                                              
add(ok): existed (file)                                                                                                                                                              
save(ok): . (dataset)                                                                                                                                                                
action summary:                                                                                                                                                                      
  add (ok: 2)
  save (ok: 1)
[{'action': 'add', 'path': '/tmp/testds/newfile', 'type': 'file', 'refds': '/tmp/testds', 'status': 'ok', 'message': '', 'key': 'MD5E-s0--d41d8cd98f00b204e9800998ecf8427e'}, {'action': 'add', 'path': '/tmp/testds/existed', 'type': 'file', 'refds': '/tmp/testds', 'status': 'ok', 'message': '', 'key': 'MD5E-s9--d908d26cac8092d475f40a5179ca6347'}, {'action': 'save', 'type': 'dataset', 'path': '/tmp/testds', 'refds': '/tmp/testds', 'status': 'ok'}]
```

so go through those with `action: 'add'` and it should give you all the paths which were just saved if any

bpinsard:
I don't think it should be limited to the one added, eg. if heudiconv is run with overwrite (and overwritten annexed files are unlocked) we want to (re) mark the updated annexed files as sensitive in case the metadata was removed on the overwritten files.

yarikoptic:
> I don't think it should be limited to the one added,

here "add" is more like `git add` -- something was committed since was modified so it should have been fine too

```
❯ echo 123 >> 123; datalad save 123
add(ok): sub-YutaMouse20/123 (file)                                                                 
save(ok): . (dataset)                                                                               
action summary:                                                                                     
  add (ok: 1)
  save (ok: 1)
❯ echo 123 >> 123; datalad save 123
add(ok): sub-YutaMouse20/123 (file)                                                                 
save(ok): . (dataset)   
```

well, let's hope that with filtering on globs nothing unexpected would happen -- let's try it out!

PR #743, opened 2024-02-26
Make it possible to review auto version -v output during release + adjust that workflow step description

yarikoptic:


.github/workflows/release.yml
```diff
@@ -24,10 +24,12 @@ jobs:
...
           chmod a+x ~/auto
 
-      - name: Check whether a release is due
+      - name: Query 'auto' on type of the release
         id: auto-version
         run: |
-          version="$(~/auto version)"
+          # to be able to debug if something goes wrong
+          auto version -v | tee /tmp/auto-version
+          version="$(tail -n 1 /tmp/auto-version)"
```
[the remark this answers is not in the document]
yarikoptic: not 100% -- but on my test run it was. 

ok -- let's try another way:

```
sed -ne '/Calculated SEMVER bump:/s,.*: *,,p' < /tmp/auto-version
```

will push now...

.github/workflows/release.yml
```diff
@@ -24,10 +24,13 @@ jobs:
...
           chmod a+x ~/auto
 
-      - name: Check whether a release is due
+      - name: Query 'auto' on type of the release
         id: auto-version
         run: |
-          version="$(~/auto version)"
+          # to be able to debug if something goes wrong
+          set -o pipefail
+          auto version -v | tee /tmp/auto-version
```
jwodder, line 32:
```suggestion
          auto version -vv | tee /tmp/auto-version
```

Use the same verbosity as in https://github.com/dandi/dandi-cli/blob/aa55b8914c8ff2216780892e5ff1796d350fd0a4/.github/workflows/release.yml#L34

.github/workflows/release.yml
```diff
@@ -24,10 +24,13 @@ jobs:
...
+          version=$(sed -ne '/Calculated SEMVER bump:/s,.*: *,,p' < /tmp/auto-version)
```
jwodder, line 33:
```suggestion
          version="$(sed -ne '/Calculated SEMVER bump:/s,.*: *,,p' /tmp/auto-version)"
```

jwodder (changes requested):

.github/workflows/release.yml
yarikoptic: according to shellcheck/trial it is not needed

```
❯ shellcheck /tmp/1.sh
❯ /tmp/1.sh
patch with space
❯ cat /tmp/1.sh
#!/bin/sh

version=$(sed -ne '/Calculated SEMVER bump:/s,.*: *,,p' /tmp/auto-version)
echo "$version"
```

jwodder (approved):

PR #755, opened 2024-04-29
fix code failure when dicom time fields are empty

jennan:
Hello,

This pull request is meant to fix an error we are encountering when processing data with empty date/time fields due to anonymization. Please note the fields are not missing but contain an empty string.

I have adapted the code of `get_datetime_from_dcm` to handle this scenario the same way as if the date/time was missing.

Please let me know if you need more details or changes to this pull request.

Thank you :).

### DIFF
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -526,18 +526,18 @@ def get_datetime_from_dcm(dcm_data: dcm.FileDataset) -> Optional[datetime.dateti
     3. SeriesDate & SeriesTime  (0008,0021); (0008,0031)
 
     """
-    acq_date = dcm_data.get("AcquisitionDate")
-    acq_time = dcm_data.get("AcquisitionTime")
-    if not (acq_date is None or acq_time is None):
+    acq_date = dcm_data.get("AcquisitionDate", "").strip()
+    acq_time = dcm_data.get("AcquisitionTime", "").strip()
+    if len(acq_date) > 0 and len(acq_time) > 0:
         return strptime_micr(acq_date + acq_time, "%Y%m%d%H%M%S[.%f]")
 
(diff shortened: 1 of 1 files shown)


heudiconv/dicoms.py
```diff
@@ -526,18 +526,18 @@ def get_datetime_from_dcm(dcm_data: dcm.FileDataset) -> Optional[datetime.dateti
...
+    acq_dt = dcm_data.get("AcquisitionDateTime", "").strip()
+    if len(acq_dt) > 0:
         return strptime_micr(acq_dt, "%Y%m%d%H%M%S[.%f]")
 
-    series_date = dcm_data.get("SeriesDate")
-    series_time = dcm_data.get("SeriesTime")
-    if not (series_date is None or series_time is None):
+    series_date = dcm_data.get("SeriesDate", "").strip()
+    series_time = dcm_data.get("SeriesTime", "").strip()
+    if len(series_date) > 0 and len(series_time) > 0:
```
yarikoptic, line 540:
```suggestion
    if series_date and series_time:
```

yarikoptic (changes requested):
Thank you for the PR.
I wonder if we could figure out some unittest.

Left some recommended changes to make code a little more generic/shorter and thus easier to read. If you do not see a gotcha with the proposed approach, please accept.

## Note to myself:

Overall this is inline with the logic we had before -- if either of them is not defined, we do not return.

But when we are talking about anonymization we might then expect the case where date is null-ed out but time is left untouched.  But since I do not think it makes sense to choose some random base for the case when only time is known -- we are ok.

jennan:
Hi @yarikoptic , sorry for the delay. I have included your changes, but I am equally happy if you merge https://github.com/nipy/heudiconv/pull/756 and close this PR :-).

PR #760, opened 2024-05-16
Specify license in CITATION.cff

yarikoptic:
Is important for JOSS paper requirements and hopefully would set license on zenodo instead of (default?) CC 4.0


yarikoptic:
Well, linters succeeded, we can proceed.

PR #761, opened 2024-05-30
[ENH] update_complex_name to handle real+imaginary complex data

bpinsard:
Someone recently shared with me some GE Hyperband multi-echo data from a SIGNA UHP, where they exported mag + real + imaginary data (weird combination but why not), all 3 ends up in the same series. So that should ideally be handled in heudiconv within `update_complex_name` which only covers mag+phase case for now.

Though I hope it won't be the final export format for that project, it could make sense to cover that usecase in heudiconv.

I don't know if there are any open test data with real+imaginary data within the same series, @neurolabusc ?


bpinsard:
Codespell get called in both Codespell and Linters? It fails in both on some files that has nothing to do with that PR.

bpinsard:
This will require https://github.com/rordenlab/dcm2niix/pull/826 to work, otherwise we might need to assume that data are "MAGNITUDE" by default in `update_complex_name`

yarikoptic:
> Codespell get called in both Codespell and Linters? It fails in both on some files that has nothing to do with that PR.

thanks!  now there is
- https://github.com/nipy/heudiconv/pull/762

heudiconv/convert.py
```diff
@@ -332,18 +332,22 @@ def update_complex_name(metadata: dict[str, Any], filename: str) -> str:
...
+    if "M" in img_type or "MAGNITUDE" in img_type:
+        part = "mag"
+    elif "P" in img_type or "PHASE" in img_type:
+        part = "phase"
+    elif "REAL" in img_type:
+        part = "real"
+    elif "IMAGINARY" in img_type:
+        part = "imag"
     else:
         raise RuntimeError("Data type could not be inferred from the metadata.")
```
yarikoptic, line 344: while at it

```suggestion
        raise RuntimeError(f"Data type could not be inferred from the ImageType={img_type}.")
```
?

heudiconv/convert.py
```diff
@@ -332,18 +332,22 @@ def update_complex_name(metadata: dict[str, Any], filename: str) -> str:
...
-    elif "P" in img_type:
-        mag_or_phase = "phase"
+    if "M" in img_type or "MAGNITUDE" in img_type:
+        part = "mag"
+    elif "P" in img_type or "PHASE" in img_type:
+        part = "phase"
+    elif "REAL" in img_type:
+        part = "real"
+    elif "IMAGINARY" in img_type:
+        part = "imag"
```
yarikoptic, line 342: if we are approaching this , could we make it more robust right away? i.e. raise and error whenever multiple/conflicting ones are found? smth like (didn't try but sketching idea)

```python
   type_to_part = {
         "M": "mag",
         "MAGNITUDE": "mag",
         "P": "phase", 
      ...
     }
     present_types = set(type_to_part).intersection(img_type)
     if not present_types:
           raise RuntimeError(f"Data type could not be inferred from the ImageType={img_type}. Known types are: {sorted(type_to_part)}")
     elif len(present_types) == 1:
           part = type_to_part[present_types[0]]
     else:
           raise RuntimeError(f"Data type could not be inferred from the ImageType={img_type}. Multiple types matched: {present_types}")
```

or if both "M" and "MAGNITUDE" could be provided (let's assume that -- more robust):

```python
     present_parts = set(type_to_part[type] for type in img_type if type in type_to_part)
     if not present_parts:
           raise RuntimeError(f"Data type could not be inferred from the ImageType={img_type}. Known types are: {sorted(type_to_part)}")
     elif len(present_parts) == 1:
           part = present_parts[0]
     else:
           raise RuntimeError(f"Data type could not be inferred from the ImageType={img_type}. Multiple BIDS parts matched: {present_parts}")
```

heudiconv/convert.py
```diff
@@ -975,9 +979,17 @@ def save_converted_files(
         is_uncombined = (
             len(set(filter(bool, channel_names))) > 1
         )  # Check for uncombined data
-        is_complex = (
-            "M" in image_types and "P" in image_types
-        )  # Determine if data are complex (magnitude + phase)
+        CPLX_PARTS = ["MAGNITUDE", "PHASE", "IMAGINARY", "REAL"]
```
yarikoptic, line 982: I need to "turn off" for today -- logic escapes me on how to possibly  reuse that `type_to_part` here ;) may be nohow ;)

yarikoptic (changes requested):
I guess let's wait on dcm2niix decision but meanwhile provided some recommenations to changes/around... can't do more today -- better go do smth as dumb as fix a car

heudiconv/convert.py
bpinsard: thanks a lot for the feedback, will try to wrap my head around that.

bpinsard:
ImageType is a mess:  just remembered that diffusion data on Siemens won't have that info in the ImageType, see https://github.com/CMRR-C2P/MB/issues/305 for the CMRR sequence where they were able to fix that.

I wonder if dcm2niix should add a new [ComplexImageComponent](https://dicom.nema.org/medical/dicom/2016e/output/chtml/part03/sect_C.8.13.3.html#table_C.8-85) json tag rather than editing the `ImageType` to match what inconsistently exists on Siemens.

heudiconv/convert.py
```diff
@@ -975,9 +979,17 @@ def save_converted_files(
...
+        is_complex = len(
+            set(
+                [
+                    part
+                    for its in image_types
+                    for part in CPLX_PARTS
+                    if part in its or part[0] in its
+                ]
+            )
+        )  # Determine if data are complex (magnitude + phase or real + imag or all-4)
```
miltoncamachocamacho, line 992:
```suggestion
        is_complex = len(
        set(
            its
            for its in image_types
            if any(part == its for part in CPLX_PARTS)
        )
        ) # Determine if data are complex (magnitude + phase or real + imag or all-4)
```
Otherwise there will be some image_types (e.g., "IMAGINARY") for which the first letter "I" is in "PRIMARY" or in "ORIGINAL" resulting in is_complex = True. Trowing errors when the update_complex_name function is run.

heudiconv/convert.py
miltoncamachocamacho, line 982: Perhaps include "M" and "P"
```suggestion
        CPLX_PARTS = ["M", "MAGNITUDE", "P", "PHASE", "IMAGINARY", "REAL"]
```

heudiconv/convert.py
miltoncamachocamacho: Thanks for all the help @yarikoptic and @bpinsard. I tried the second suggestion and it is indeed more robust.

PR #770, opened 2024-06-21
Replace deprecated "_phase" suffix in reproin heuristic with "part-phase" entity

### LINKED ISSUE #769: ReproIn heuristic uses `phase` suffix instead of `part-phase`
### Summary

The ReproIn heuristic (`reproin`) currently uses suffixes to distinguish complex-valued fMRI data (see below). However, the `phase` suffix is deprecated in favor of the `part` entity.

https://github.com/nipy/heudiconv/blob/4c2c2c06835f26474bc2c710a228242d2cdd45b3/heudiconv/heuristics/reproin.py#L501-L504

The difficult thing is that `infotodict` is run separately on each run, and magnitude and phase data from the same run are typically (at least with the CMRR MB sequence) split into separate folders of DICOMs, so they're processed separately. I don't know how to flexibly identify magnitude BOLD runs that need `part-mag` (i.e., runs with phase data also reconstructed) vs. those that don't (magnitude-only reconstruction).

### Platform details:

n/a

tsalo:
Closes #769. This is working on DICOMs associated with https://openneuro.org/datasets/ds005250.

Changes proposed:

- Track previous and next sequence info in ReproIn's `infotodict`.
- ~~Rename `s` variable to `curr_seqinfo` for improved clarity.~~ Moved to #779.
- For non-SBRef, non-fieldmap scans, set "part" entity to "phase" for phase data instead of setting the suffix to "phase".
- For magnitude non-SBRef, non-fieldmap scans, check if the _next_ sequence has the same `series_description` and is a phase scan. If it is, set "part" to "mag".

### DIFF
--- a/heudiconv/heuristics/reproin.py
+++ b/heudiconv/heuristics/reproin.py
@@ -403,64 +403,61 @@ def infotodict(
     run_label: Optional[str] = None  # run-
     dcm_image_iod_spec: Optional[str] = None
     skip_derived = False
-    for i_acq, s in enumerate(seqinfo):
+    for i_acq, curr_seqinfo in enumerate(seqinfo):
         # XXX: skip derived sequences, we don't store them to avoid polluting
         # the directory, unless it is the motion corrected ones
         # (will get _rec-moco suffix)
-        if skip_derived and s.is_derived and not s.is_motion_corrected:
-            skipped.append(s.series_id)
-            lgr.debug("Ignoring derived data %s", s.series_id)
+        if (
+            skip_derived
+            and curr_seqinfo.is_derived
+            and not curr_seqinfo.is_motion_corrected
+        ):
+            skipped.append(curr_seqinfo.series_id)
+            lgr.debug("Ignoring derived data %s", curr_seqinfo.series_id)
             continue
 
         if i_acq == 0:
+            prev_seqinfo = None
             prev_dcm_image_iod_spec = None
-            prev_image_type_datatype = None
         else:
-            prev_dcm_image_iod_spec = seqinfo[i_acq - 1].image_type[2]
-            prev_image_type_datatype = {
-                # Note: P and M are too generic to make a decision here, could be
-                #  for different datatypes (bold, fmap, etc)
-                "FMRI": "func",
-                "MPR": "anat",
-                "DIFFUSION": "dwi",
-                "MIP_SAG": "anat",  # angiography
-                "MIP_COR": "anat",  # angiography
-                "MIP_TRA": "anat",  # angiography
-            }.get(prev_dcm_image_iod_spec, None)
+            prev_seqinfo = seqinfo[i_acq - 1]
+            for f in "series_description", "protocol_name":
+                prev_seqinfo = prev_seqinfo._replace(
+                    **{f: getattr(prev_seqinfo, f).format(**prev_seqinfo._asdict())}
+                )
+
+            prev_dcm_image_iod_spec = prev_seqinfo.image_type[2]
 
         if i_acq == (len(seqinfo) - 1):
+            next_seqinfo = None
             next_dcm_image_iod_spec = None
-            next_image_type_datatype = None
         else:
-            next_dcm_image_iod_spec = seqinfo[i_acq + 1].image_type[2]
-            next_image_type_datatype = {
-                # Note: P and M are too generic to make a decision here, could be
-                #  for different datatypes (bold, fmap, etc)
-                "FMRI": "func",
-                "MPR": "anat",
-                "DIFFUSION": "dwi",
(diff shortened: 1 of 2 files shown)


heudiconv/heuristics/reproin.py
```diff
@@ -493,15 +522,30 @@ def infotodict(
...
+                    # if next one is phase fMRI, we should set part to mag
+                    if (
+                        (
+                            next_seqinfo.series_description
+                            == curr_seqinfo.series_description
+                        )
+                        and (next_dcm_image_iod_spec == "P")
+                        and not curr_seqinfo.series_description.endswith("_SBRef")
+                    ):
+                        series_info["part"] = "mag"
```
tsalo, line 548: Here's the main change. If necessary, I can try extending it to other suffixes (e.g., DWI).

yarikoptic:
> Rename s variable to curr_seqinfo for improved clarity.

yeah, but complicates review a little due to diff now including changes I can ignore or not and need to decide when looking...

Could you submit a separate PR with that typo fix and such a rename (so no functionality change) and then let's improve this one -- yet to grasp on all the changes, and as we do not have any nice testing for reproin heuristic, I wonder how we could add one :-/  I would really be shy to accept without some way to test.  Locally I could do using smth like https://github.com/nipy/heudiconv/blob/master/utils/test-compare-two-versions.sh but could we do better?!

tsalo:
:+1: I've opened #779 with the variable name change.

heudiconv/heuristics/reproin.py
```diff
@@ -523,6 +546,19 @@ def infotodict(
...
                 datatype_suffix = "dwi"
 
+            # Add "part" entity as needed
+            if datatype != "fmap" and not curr_seqinfo.series_description.endswith(
+                "_SBRef"
+            ):
+                if "P" in curr_seqinfo.image_type:
+                    series_info["part"] = "phase"
+                elif "M" in curr_seqinfo.image_type:
+                    # if next one is phase from same scan, we should set part to mag
```
yarikoptic, line 556: is it guaranteed to be the "next", could it be "prior"?

heudiconv/heuristics/reproin.py
```diff
@@ -523,6 +546,19 @@ def infotodict(
                 # label for dwi as well
                 datatype_suffix = "dwi"
 
+            # Add "part" entity as needed
+            if datatype != "fmap" and not curr_seqinfo.series_description.endswith(
+                "_SBRef"
+            ):
+                if "P" in curr_seqinfo.image_type:
+                    series_info["part"] = "phase"
```
yarikoptic, line 554: codecov says that this line is not covered, not yet sure how wide implication is ;-)

heudiconv/heuristics/reproin.py
```diff
@@ -643,8 +681,8 @@ def from_series_info(name: str) -> Optional[str]:
             "_Scout" in curr_seqinfo.series_description
             or (
                 datatype == "anat"
-                and datatype_suffix
-                and datatype_suffix.startswith("scout")
+                and filename_suffix_parts[-1]
+                and filename_suffix_parts[-1].startswith("scout")
```
yarikoptic, line 685: what does this have to do with phase suffix? nothing in short commit messages hints on possible relation to this change... why is it?

heudiconv/heuristics/reproin.py
```diff
@@ -402,15 +402,43 @@ def infotodict(
...
+        else:
+            next_seqinfo = seqinfo[i_acq + 1]
+            for f in "series_description", "protocol_name":
+                next_seqinfo = next_seqinfo._replace(
+                    **{f: getattr(next_seqinfo, f).format(**next_seqinfo._asdict())}
+                )
+
+            next_series_description = next_seqinfo.series_description
+            next_dcm_image_iod_spec = next_seqinfo.image_type[2]
+
```
yarikoptic, line 441: this duplicates that `_replace` hack 2 more times, causes me duplication-allergy...

Since we are to do that anyways, do it once then -- just in a loop before this loop apply the replacements to all first.

Then avoid that `i_acq` - do following 'zip'ing, as chatgpt suggested

```
my_list = [1, 2, 3, 4, 5]

for previous, current, next in zip([None] + my_list[:-1], my_list, my_list[1:] + [None]):
    print(f"Previous: {previous}, Current: {current}, Next: {next}")
```

which would avoid all the if's and else's etc and make code easier to read IMHO.

yarikoptic (changes requested):
Thank you! some feedback to improve it.

yarikoptic:
@tsalo -- ping on this. Would you happen to get time to address comments?

tsalo:
Sorry, I've been really busy. I might have time in the next couple of weeks though.

PR #772, opened 2024-07-03
[JOSS] FIX split names into given-names/surname/suffix

mvdoc:
As per https://joss.readthedocs.io/en/latest/paper.html#names and https://github.com/openjournals/joss-reviews/issues/5839#issuecomment-2207175596

This should create a correct crossref and CITATION file

### DIFF
--- a/paper.md
+++ b/paper.md
@@ -9,85 +9,113 @@ tags:
  - open science
  - FOSS
 authors:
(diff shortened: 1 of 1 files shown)


yarikoptic:
oh, I was too old fashioned!

Does it have affect on article "Authors" rendering?

not sure what are procedures at JOSS level to adjust records.

Issue #777, opened 2024-08-07
Feature Request: support dicom datetime with timezone information


AlanKuurstra:
### Summary
Dicom conversion fails if the dicom datetime contains timezone information, eg. "20230905141258.340-0400"

>Traceback (most recent call last):
  File "/usr/local/bin/heudiconv", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/cli/run.py", line 30, in main
    workflow(**kwargs)
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/main.py", line 479, in workflow
    prep_conversion(
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/convert.py", line 259, in prep_conversion
    convert(
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/convert.py", line 637, in convert
    save_scans_key(item, bids_outfiles)
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/bids.py", line 494, in save_scans_key
    rows[f_name] = get_formatted_scans_key_row(item[-1][0])
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/bids.py", line 579, in get_formatted_scans_key_row
    acq_datetime = dicoms.get_datetime_from_dcm(dcm_data=dcm_data)
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/dicoms.py", line 536, in get_datetime_from_dcm
    return strptime_micr(acq_dt, "%Y%m%d%H%M%S[.%f]")
  File "/usr/local/lib/python3.9/dist-packages/heudiconv/utils.py", line 689, in strptime_micr
    return datetime.strptime(date_string, fmt)
  File "/usr/lib/python3.9/_strptime.py", line 568, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
  File "/usr/lib/python3.9/_strptime.py", line 352, in _strptime
    raise ValueError("unconverted data remains: %s" %
ValueError: unconverted data remains: .372-0400

Perhaps [strptime_micr](https://github.com/nipy/heudiconv/blob/41b6d83d16bf35e701f1d86cf80d21abc81b3aa5/heudiconv/utils.py#L671) could support a fmt with an additional optional [%z]?

PR #780, opened 2024-08-14
timezone aware

AlanKuurstra:
addresses #777 

Adds capability to handle datetime stripping for all datetimes allowed in the dicom and bids standards.

### DIFF
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -536,11 +536,15 @@ def get_datetime_from_dcm(dcm_data: dcm.FileDataset) -> Optional[datetime.dateti
     3. SeriesDate & SeriesTime  (0008,0021); (0008,0031)
 
     """
-    if "AcquisitionDate" in dcm_data and "AcquisitionTime" in dcm_data:
+
+    def check_tag(x):
+        return x in dcm_data and dcm_data[x].value.strip()
+
+    if check_tag("AcquisitionDate") and check_tag("AcquisitionTime"):
         return strptime_dcm_da_tm(dcm_data, "AcquisitionDate", "AcquisitionTime")
-    if "AcquisitionDateTime" in dcm_data:
+    if check_tag("AcquisitionDateTime"):
         return strptime_dcm_dt(dcm_data, "AcquisitionDateTime")
-    if "SeriesDate" in dcm_data and "SeriesTime" in dcm_data:
+    if check_tag("SeriesDate") and check_tag("SeriesTime"):
         return strptime_dcm_da_tm(dcm_data, "SeriesDate", "SeriesTime")
     return None
 
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -32,7 +32,8 @@ from .utils import (
     get_typed_attr,
     load_json,
     set_readonly,
-    strptime_micr,
+    strptime_dcm_da_tm,
+    strptime_dcm_dt
 )
 
 if TYPE_CHECKING:
@@ -535,19 +536,12 @@ def get_datetime_from_dcm(dcm_data: dcm.FileDataset) -> Optional[datetime.dateti
     3. SeriesDate & SeriesTime  (0008,0021); (0008,0031)
 
     """
-    acq_date = dcm_data.get("AcquisitionDate", "").strip()
-    acq_time = dcm_data.get("AcquisitionTime", "").strip()
-    if acq_date and acq_time:
-        return strptime_micr(acq_date + acq_time, "%Y%m%d%H%M%S[.%f]")
-
-    acq_dt = dcm_data.get("AcquisitionDateTime", "").strip()
-    if acq_dt:
-        return strptime_micr(acq_dt, "%Y%m%d%H%M%S[.%f]")
-
-    series_date = dcm_data.get("SeriesDate", "").strip()
-    series_time = dcm_data.get("SeriesTime", "").strip()
-    if series_date and series_time:
-        return strptime_micr(series_date + series_time, "%Y%m%d%H%M%S[.%f]")
+    if "AcquisitionDate" in dcm_data and "AcquisitionTime" in dcm_data:
+        return strptime_dcm_da_tm(dcm_data, "AcquisitionDate", "AcquisitionTime")
+    if "AcquisitionDateTime" in dcm_data:
+        return strptime_dcm_dt(dcm_data, "AcquisitionDateTime")
+    if "SeriesDate" in dcm_data and "SeriesTime" in dcm_data:
+        return strptime_dcm_da_tm(dcm_data, "SeriesDate", "SeriesTime")
     return None
 
 
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -537,7 +537,7 @@ def get_datetime_from_dcm(dcm_data: dcm.FileDataset) -> Optional[datetime.dateti
 
     """
 
-    def check_tag(x):
+    def check_tag(x: str) -> bool:
         return x in dcm_data and dcm_data[x].value.strip()
 
     if check_tag("AcquisitionDate") and check_tag("AcquisitionTime"):
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -33,7 +33,7 @@ from .utils import (
     load_json,
     set_readonly,
     strptime_dcm_da_tm,
-    strptime_dcm_dt
+    strptime_dcm_dt,
 )
 
 if TYPE_CHECKING:
--- a/heudiconv/tests/test_utils.py
+++ b/heudiconv/tests/test_utils.py
@@ -9,6 +9,7 @@ from pathlib import Path
 from typing import IO, Any
 from unittest.mock import patch
 
+import pydicom as dcm
 import pytest
 
 from heudiconv.utils import (
@@ -23,6 +24,9 @@ from heudiconv.utils import (
     remove_suffix,
     save_json,
     strptime_micr,
+    strptime_bids,
+    strptime_dcm_da_tm,
+    strptime_dcm_dt,
     update_json,
 )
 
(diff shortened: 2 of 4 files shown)


yarikoptic:
dang, I have managed to miss this PR and CI managed to not run test for some reason!
I will attend to it within a few days.  If you could rebase or otherwise refresh it to re-trigger the CI, would be appreciated.

AlanKuurstra:
I rebased, CI is awaiting approval from a maintainer.

heudiconv/utils.py
```diff
@@ -662,32 +664,107 @@ def get_datetime(date: str, time: str, *, microseconds: bool = True) -> str:
...
+    """
+    # https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html
+    datetime_str = dcm_data.get(dt_tag)
+    fmts = ["%Y%z", "%Y%m%z", "%Y%m%d%z", "%Y%m%d%H%z", "%Y%m%d%H%M%z", "%Y%m%d%H%M%S%z", "%Y%m%d%H%M%S.%f%z",
+            "%Y", "%Y%m", "%Y%m%d", "%Y%m%d%H", "%Y%m%d%H%M", "%Y%m%d%H%M%S", "%Y%m%d%H%M%S.%f"]
+    datetime_obj = strptime(datetime_str, fmts)
+
+    if not datetime_obj.tzinfo and (0x0008, 0x0201) in dcm_data:
+        utc_offset = dcm_data[0x0008, 0x0201].value
+        datetime_obj = datetime_utc_offset(datetime_obj, utc_offset) if utc_offset else datetime_obj
```
yarikoptic, line 766: I am not sure who we can receive a datetime_obj with timezone from above, but I think it better be explicitly checked and warned about so we do not quietly overload it:

```suggestion
    if utc_offset_dcm := dcm_data.get([0x0008, 0x0201]):
        if utc_offset := utf_offset_dcm.value:
             datetime_obj2 = datetime_utc_offset(datetime_obj, utc_offset)
             if datetime_obj.tzinfo and datetime_obj2 != datetime_obj:
                 lgr.warning("Unexpectedly previously parsed datetime %s contains zoneinfo which is different from the one obtained from DICOMs UTFOffset field: %s", datetime_obj, datetime_obj2)
             else:
                 datetime_obj = datetime_obj2 
```

or alike?

heudiconv/utils.py
```diff
@@ -662,32 +664,107 @@ def get_datetime(date: str, time: str, *, microseconds: bool = True) -> str:
...
 
+def datetime_utc_offset(datetime_obj: datetime, utc_offset: str):
+    """set the datetime's tzinfo by parsing an utc offset string"""
+    sign, hours, minutes = re.match(r"([+\-]?)(\d{2})(\d{2})", utc_offset).groups()
+    sign = -1 if sign == '-' else 1
+    hours, minutes = int(hours), int(minutes)
+    tzinfo = datetime.timezone(sign * datetime.timedelta(hours=hours, minutes=minutes))
+    return datetime_obj.replace(tzinfo=tzinfo)
 
-def strptime_micr(date_string: str, fmt: str) -> datetime:
```
yarikoptic, line 671: Please keep prior implementation around, but add a deprecation warning. It could be that someone used it in their heuristic, so we better do not just rip it from under them.

heudiconv/tests/test_utils.py
```diff
@@ -173,19 +176,85 @@ def test_get_datetime() -> None:
...
+    target = None
+    if dt_fmt[-2:] == "%z" and offset:
+        target = datetime.strptime(dt, dt_fmt)
+    else:
+        target = datetime.strptime(dt + offset, dt_fmt + offset_fmt)
+    ds = dcm.dataset.Dataset()
+    ds["AcquisitionDateTime"] = dcm.DataElement("AcquisitionDateTime","DT", dt)
+    if offset:
+        ds[(0x0008, 0x0201)] = dcm.DataElement((0x0008, 0x0201), "SH", offset)
+    assert strptime_dcm_dt(ds, "AcquisitionDateTime") == target
```
yarikoptic, line 256: It is unfortunate that we do not carry any "proper" dicom file with timezone info. But FWIW our dependencies do:

```
❯ /usr/bin/find -iname *.dcm | while read d; do off=$(dcmdump +P 0008,0201 $d 2>/dev/null) ; [ -n "$off" ] || continue; echo -e "$d\t$off"; done | head
./.tox/typing/lib/python3.11/site-packages/pydicom/data/charset_files/chrArab.dcm	(0008,0201) SH [-0400]                                  #   6, 1 TimezoneOffsetFromUTC
./.tox/typing/lib/python3.11/site-packages/pydicom/data/charset_files/chrFren.dcm	(0008,0201) SH [-0400]                                  #   6, 1 TimezoneOffsetFromUTC
./.tox/typing/lib/python3.11/site-packages/pydicom/data/charset_files/chrFrenMulti.dcm	(0008,0201) SH [-0400]                                  #   6, 1 TimezoneOffsetFromUTC

```

but I do not think we should rely on them.  But we could copy some I guess. WDYT, worth it?

heudiconv/dicoms.py
```diff
@@ -531,19 +532,12 @@ def get_datetime_from_dcm(dcm_data: dcm.FileDataset) -> Optional[datetime.dateti
...
-    acq_dt = dcm_data.get("AcquisitionDateTime", "").strip()
-    if acq_dt:
-        return strptime_micr(acq_dt, "%Y%m%d%H%M%S[.%f]")
-
-    series_date = dcm_data.get("SeriesDate", "").strip()
-    series_time = dcm_data.get("SeriesTime", "").strip()
-    if series_date and series_time:
-        return strptime_micr(series_date + series_time, "%Y%m%d%H%M%S[.%f]")
+    if "AcquisitionDate" in dcm_data and "AcquisitionTime" in dcm_data:
+        return strptime_dcm_da_tm(dcm_data, "AcquisitionDate", "AcquisitionTime")
```
yarikoptic, line 536: FWIW this is not functionally equivalent to above since we did `strip()` on values, so if string of spaces is a value for some reason (don't want to dig now why we even did such safe guarding), it would not be considered, which is the case here.  Could you adjust so we stay 100% consistent with prior implementation?

yarikoptic (changes requested):
Thanks again for the PR! Let's get "zone aware!" ;-)
See individual comments for some minor tune ups requested

heudiconv/tests/test_utils.py
AlanKuurstra: I'm not sure I see the benefit to adding dicom files to the git repository. Generating the dicom from scratch facilitates testing every tag allowed by the standard without keeping numerous files in the repository.

yarikoptic: there are pros/cons for both approaches. e.g. for existing/committed (to git or via git-annex) files -- ensuring that we are operating correctly on the files as provided by manufacturers. Not surprising that dcm2niix has all those collections linked as submodules: https://github.com/rordenlab/dcm2niix/blob/master/.gitmodules

heudiconv/dicoms.py
yarikoptic: not sure why it was marked resolved since  code remains the same -- if empty value we would proceed to the function and return

heudiconv/utils.py
```diff
@@ -681,12 +684,123 @@ def strptime_micr(date_string: str, fmt: str) -> datetime:
...
+    datetime_str = dcm_data[dt_tag].value.strip()
+    fmts = ["%Y%z", "%Y%m%z", "%Y%m%d%z", "%Y%m%d%H%z", "%Y%m%d%H%M%z", "%Y%m%d%H%M%S%z", "%Y%m%d%H%M%S.%f%z",
+            "%Y", "%Y%m", "%Y%m%d", "%Y%m%d%H", "%Y%m%d%H%M", "%Y%m%d%H%M%S", "%Y%m%d%H%M%S.%f"]
+    datetime_obj = strptime(datetime_str, fmts)
+
+    if utc_offset_dcm := dcm_data.get((0x0008, 0x0201)):
+        if utc_offset := utc_offset_dcm.value.strip():
+            datetime_obj2 = datetime_utc_offset(datetime_obj, utc_offset)
+            if datetime_obj.tzinfo and datetime_obj2 != datetime_obj:
+                lgr.warning("Unexpectedly previously parsed datetime %s contains zoneinfo which is different from the one obtained from DICOMs UTFOffset field: %s", datetime_obj, datetime_obj2)
```
yarikoptic, line 800: please use `black` and/or ideally `pre-commit` which we have for the project -- it should take care about reformatting those long lines etc

heudiconv/utils.py
```diff
@@ -681,12 +684,123 @@ def strptime_micr(date_string: str, fmt: str) -> datetime:
...
+    datetime_obj = datetime.datetime.combine(date.date(), time.time())
+
+    if utc_offset_dcm := dcm_data.get((0x0008, 0x0201)):
+        utc_offset = utc_offset_dcm.value.strip()
+        datetime_obj = datetime_utc_offset(datetime_obj, utc_offset) if utc_offset else datetime_obj
+    return datetime_obj
+
+
+def strptime_dcm_dt(dcm_data: dcm.Dataset, dt_tag: TagType) -> datetime.datetime:
+    r"""
```
yarikoptic, line 779: I know that you might have followed a trend of some docstrings sneaking in with `r` for no reason, but if no reason here in new ones -- please do not use them since adds cognitive burden to later possibly look for something needing this `raw` annotation

heudiconv/tests/test_utils.py
yarikoptic: but you are correct that on the fly generation could be quite handy for how you have done them above - in case of parametric tests etc. I was just wondering if we could add testing on some existing file with such metadata in it... but if none there -- let's forget about this particular aspect

yarikoptic (approved):
thank you @AlanKuurstra , let's proceed!

PR #788, opened 2024-09-24
feat: rename BIDS files for multi-orientation series.

bpinsard:
(rebase of https://github.com/nipy/heudiconv/pull/441 , many years later I know slightly more what I am doing.)
The purpose of that PR is to rename files adding `acq-orient`  when there are multiple files with varying orientation generated by dcm2niix ( `...._i00000[0-9].json/nii.gz`) . Otherwise heudiconv want to rename the separate files to the same BIDS name and crash. 
 
The obvious main use-case would be 3-plane localizers. 
You're gonna tell me: why do you want to bidsify localizers? 
Well, why not? Also, we started exporting the localizers combined+uncombined as a potential way to QA the head-coil elements. And I would like to have that future tool running on clean BIDS, not on messy dicoms.


heudiconv/convert.py
```diff
@@ -523,6 +523,35 @@ def update_uncombined_name(
     return filename
 
 
+def update_multiorient_name(
+    metadata: dict[str, Any],
+    filename: str,
+) -> str:
+    if "acq-" in filename:
+        lgr.warning(
+            "Not embedding multi-orientation information as prefix already uses acq- parameter."
```
yarikoptic, line 532:
```suggestion
            "Not embedding multi-orientation information as `%r` already uses acq- parameter.", filename
```

so user gets a clue on what you are warning about.

heudiconv/convert.py
```diff
@@ -523,6 +523,35 @@ def update_uncombined_name(
     return filename
 
 
+def update_multiorient_name(
+    metadata: dict[str, Any],
+    filename: str,
+) -> str:
+    if "acq-" in filename:
```
yarikoptic, line 530:
```suggestion
    if "_acq-" in filename:
```

so doesn't catch some possible (now or in the future) `megaacq-`?

heudiconv/convert.py
```diff
@@ -523,6 +523,35 @@ def update_uncombined_name(
...
+        lgr.warning(
+            "Not embedding multi-orientation information as prefix already uses acq- parameter."
+        )
+        return filename
+    iop = metadata.get("ImageOrientationPatientDICOM")
+    # iop = [round(x) for x in iop]
```
yarikoptic, line 536:
```suggestion
```

heudiconv/convert.py
```diff
@@ -523,6 +523,35 @@ def update_uncombined_name(
...
+    assert isinstance(iop, list)
+    cross_prod = [
+        iop[1] * iop[5] - iop[2] * iop[4],
+        iop[2] * iop[3] - iop[0] * iop[5],
+        iop[0] * iop[4] - iop[1] * iop[3],
+    ]
+    cross_prod = [abs(x) for x in cross_prod]
+    slice_orient = ["sagittal", "coronal", "axial"][cross_prod.index(1)]
+    bids_pairs = filename.split("_")
+    # acq needs to be inserted right after sub- or ses-
```
yarikoptic, line 546: according to
https://github.com/bids-standard/bids-specification/blob/master/src/schema/rules/entities.yaml
it should be e.g. after a possible `_task-`  .  But overall -- reminded me about unfinished 
- https://github.com/nipy/heudiconv/pull/544
and that we have base implementation in
- https://github.com/nipy/heudiconv/blob/383525d7d3bdff7fbe52287aa7ee1d091532e39f/heudiconv/bids.py#L1069

couldn't you use it here?

yarikoptic (changes requested):
Overall - makes sense, thanks!  Left some recommendations/comments.

My main concern is that we would do that "all the time" and unconditionally to anything else , i.e. may be there is some other entity for which we would already separate by and thus this orientation would not be needed?

what I mean -- what other (if any) scans other than localizer/scouts could have multiple `ImageOrientationPatientDICOM `?

yarikoptic:
Thanks @bpinsard . But could you add anything about that concern of "all the time"? ;-)

bpinsard:
For now, multi-orient series cause heudiconv to crash unless the overwrite flag is used, but in that case only one nifti (the last) is kept. 

So the only hypothetical case where we would add the `acq-` entity while deduplication of filename was not required, would be if each orientation would be already renamed with `-part`,  `echo-` or `ch-` entity, which IMO would be surprising to differentially acquire (echo index) or reconstruct (part|ch) each acquired orientation.

The only series other than localizers that could maybe result in multiple ImageOrientationPatientDICOM that I could think of would be non-product sequences:
- maybe using FIRMM, or sequences that use navigators to estimate motion and adjust the FoV accordingly (like in-utero fMRI/DWI with large motion amplitude)
- a custom sequence that would acquire multiple orientations and outputs in a single series for offline hyper-resolution.
- something non brain that I am not aware of.

However, depending on how the dicoms contain that dynamic orientation, they might not be converted to separate nifti by dcm2niix.
Of course there is still a case not accounted for, where let say mutliple axial series are acquired with slightly varying IOP, would all be name `acq-axial`. Which reminds me, colleagues in my old PhD lab used  https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.25143 IIRC to acquire fMRI simultaneously in brain and cervical spine. I will try to retrieve some of these data and check what is the dcm2niix outputs.
Maybe we should then rename by adding an index (as is done by dcm2niix) to an existing BIDS entity.

I doubt that all the possible cases of fancy custom WIP/C2P can be planned, as it seems there is room for developers to decide how dicoms are formatted.

(sorry for the lengthy response)

bpinsard:
Retrieved some of this fancy multi-FoV sequence dicoms, and the IOP is indeed different for the 2 resulting niftis (brain, cspine), and would likely be both labelled  as `axial`. :(

yarikoptic:
I will move it to draft until we decide on the way forward (correct me if I am mistaken and it is all clear)

bpinsard:
Thanks for the follow-up!
The simpler way forward would be to have `acq-orient{i}` to be added to the files, as this is the only sensible entity to be available to all bids modalities AFAIK (otherwise `chunk` would have been a good candidate). When `acq` is already assigned a value, we could suffix it to the existing value (camelcasing is the only possible option with BIDS). Not ideal, but better than heudiconv crashing on these rare instance of multi-FoV-single-series acquisitions. 
WDYT?

bpinsard:
Changed the strategy: when multiple FoVs in same series generates multiple niftis with dcm2niix, we add `chunk-` entity with an increasing index. Not all MRI suffixes are allowed to use `chunk-` though, so that won't always be BIDS-valid. 
It required adding a few missing "known entities" in `bids.py`. 
Tested that on the 3-planes localizer and old dual-FoV bold C2P data.

yarikoptic-gitmate (commented):
edit by @yarikoptic: I let claude run a little too wild here... migth take time to review that #869 for plausibility...
---- 

Went and checked what dcm2niix does now and where BIDS stands, since that was the open question. **`chunk` is the right call** — details below. I've pushed a rebase of this onto master plus fixes for what I found: #869.

## dcm2niix has no answer to align with

With `dcm2niix==1.0.20260724` from PyPI, on this PR's own localizer fixture:

- plain `-b y` → `..._i00001/2/3`, a non-BIDS postfix, plus a new sidecar hint: `"BidsGuess": ["discard", "_acq-fl2_run-1_localizer"]`
- `-f %h` / `-f %H -br .` (its own BIDS / ReproIn naming, both warn "experimental") → a full tree, but the split is *still* disambiguated by `_i0000N` appended **after** the suffix: `sub-…_acq-fl2_run-1_localizer_i00001.nii.gz`, under a `discard/` datatype. `--terse` does not drop it.
- the `dataset_description.json` it writes declares `"BIDSVersion": "1.8.0"`.

So it punts, and heudiconv has to decide.

## BIDS says `chunk`

Against `bidsschematools` 1.2.7 (BIDS 1.11.1): `chunk` was generalized from microscopy to MRI in **1.9.0** (bids-standard/bids-specification#1586), and its definition is this case verbatim — *"images of the same physical sample with different fields of view acquired in the same imaging experiment … may be used to indicate different anatomical structures or regions of the same structure"*. That covers the brain+c-spine dual-FoV BOLD too.

Allowed in `anat` (all suffix groups), `func`, `dwi`, `fmap`, `micr`. **Not** in `perf`, `pet`, `mrs` — so your "won't always be BIDS-valid" caveat is real but narrow.

`acq-orient{i}` would have been worse: `acq` is a free-form label the user's heuristic owns, and CamelCase-appending to an existing value makes it unparseable. `chunk` is an index entity meant for exactly this.

Two things worth stating plainly in the PR: there is still no `localizer`/`scout` suffix in BIDS at all, so the flagship use case is non-BIDS whatever entity we pick (reproin emits scouts as DICOM-only, dcm2niix says `discard`) — and it'd be good to warn for the `perf`/`pet`/`mrs` suffixes rather than silently emit invalid names.

## Correcting the premise

On current master this does **not** crash — it silently falls through to the numeric fallback and mangles the *suffix*:

```
master:  sub-loc_T1w1.nii.gz   sub-loc_T1w2.nii.gz   sub-loc_T1w3.nii.gz
this PR: sub-loc_chunk-1_T1w.nii.gz  sub-loc_chunk-2_T1w.nii.gz  sub-loc_chunk-3_T1w.nii.gz
```

Which if anything raises the priority: silent invalid output is worse than a crash.

## What I fixed in #869

1. **The index is arbitrary.** `sorted(iops)` is a lexicographic sort of the *string* rendering of a list of floats. On your own fixture that gives Sag/Cor/Tra = chunk-1/2/3 although Tra was acquired before Cor — and it is not stable across subjects: any obliquity or float-formatting difference reshuffles it, so `chunk-1` is the sagittal plane for one subject and the coronal one for the next, and your brain/c-spine FoVs could swap. Now indexed by distinct orientation in file order — deterministic, and still groups files sharing an orientation (which matters when the series is *also* split by echo or mag/phase; a file-position index would have got that wrong).

2. **`_known_entities` ordering.** The five entities added here are appended at the end, but the entity table interleaves them: `… task, tracksys, acq, nuc, voi, ce, trc, stain, rec, …`. Since `__str__` rebuilds the name from that list, any name carrying one came out non-compliant. (`chunk` itself happened to land right.) Filled the list in from `rules/entities.yaml` instead.

3. **Routing through `BIDSFile` loses information** — and this PR is the first production path to feed real output basenames through `parse`/`__str__`. `parse` dropped entities it didn't know (`sub-1_space-T1w_desc-x_bold.nii.gz` → `sub-1_bold.nii.gz`), and it looked for `key-value` pairs anywhere, assuming the last preceded the suffix — so reproin's `sub-1_task-rest_bold__dup-01` lost `bold` outright. Also `ch-`, heudiconv's own entity, got relocated to the end and warned about.

4. **Two hard aborts and one collision.** A basename with no entities → `IndexError`, one without `sub-` → `ValueError`, both out of `save_converted_files`; a sidecar missing `ImageOrientationPatientDICOM` while its siblings have it → `KeyError`. And since dcm2niix splits on more than orientation, a series can yield two images sharing one — both get `chunk-1`, the numeric fallback is skipped because the name *was* modified, and `safe_movefile` then dies moving the second onto the first (reproduced with a 4-image localizer; master converts it fine).

Minor: copy-pasted docstrings in `bids_localizer.py` ("Multi-Echo … `_ME_`") and in `test_convert_multiorient` ("labeled in `acq` entity", "checks that the echo entity comes before part"); fixture is 2.0MB of 512² pixel data for a filename test, and still carried `InstitutionAddress`/`OperatorsName`/`ReferringPhysicianName`. All handled in #869.

Left alone, worth a thought: `is_multiorient` compares orientations as exact strings, so two files of the same plane differing in the last decimal read as multi-orient.

---

PR #791, opened 2024-10-15
[ENH]  lighter alpine multi-stage build docker image, ~65% smaller

bpinsard:
*Disclaimer: I don't know enough about Neurodocker and Reproenv to know if these are critical to run heudiconv image in specific contexts or if there are any advantages compared to a bare alpine version.*

Alpine allow:
 - way lighter image 
```
$ docker image ls | grep heudiconv
heudiconv                         alpine             6c44f1a0115f   10 minutes ago   517MB
nipy/heudiconv                    latest             241fc4f2610c   13 days ago      1.44GB
```
 - more recent git-annex version than what is shipped by apt. 

Also: 
- Size reduction are in part due to multi-stage build.
- Updated dcm2niix in the meantime.

I might have forgotten some non-pip managed heudiconv reqs, it would be safer for the `docker.yml` workflow to run the tests on the image before pushing to dockerhub. Also, if `bash` or other term stuffs are used in specific deployment, it can be added.


bpinsard:
Why in hell do the 3.9+ tests crash when just changeing the dockerfile that is not used in the tests!? 🤯

bpinsard:
Nibabel https://github.com/nipy/nibabel/pull/1336 broke heudiconv tests via dcmstack. 
See https://github.com/moloney/dcmstack/issues/94

bpinsard:
Heudiconv itself could be made lighter if this wasn't 12MB https://github.com/nipy/heudiconv/tree/master/heudiconv/tests/data/b0dwiForFmap

bpinsard:
forced-push to get the tests to pass with newer nibabel version.

yarikoptic:
Thanks for cooking it up. For now I just merged "my smaller PR #790" progressing forward base versions. I wonder if we could/should position this one as `heudiconv-alpine` instead of completely replacing one with debian/conda happen someone relies on having those components already... That would require moving your recipe into a separate Dockerfile (e.g. Dockerfile.alpine) and adjusting our CI though to build/push multiple images.

bpinsard:
Make sense, I will work on the changes.

PR #793, opened 2024-10-21
Fix assignment of sensitive git-annex metadata data via glob patterns (regression introduced by #739)

bpinsard:
I introduced a regression in #739, and it was not flagged by the tests.

This PR fixes the glob pattern to properly flag the files as sensitive. 
This is critical as if any data management pipeline rely on heudiconv setting this metadata, then sensitive data might not be handled appropriately/exposed.
I added that feature to the regression test, as this already require datalad to download data.
@yarikoptic let me know if you prefer this test to be it in a separate function.

### DIFF
--- a/heudiconv/external/dlad.py
+++ b/heudiconv/external/dlad.py
@@ -156,12 +156,12 @@ def add_to_datalad(
 
     # Provide metadata for sensitive information
     sensitive_patterns = [
-        "sourcedata",
+        "sourcedata/**",
         "*_scans.tsv",  # top level
         "*/*_scans.tsv",  # within subj
         "*/*/*_scans.tsv",  # within sess/subj
-        "*/anat",  # within subj
-        "*/*/anat",  # within ses/subj
+        "*/anat/*",  # within subj
+        "*/*/anat/*",  # within ses/subj
     ]
     for sp in sensitive_patterns:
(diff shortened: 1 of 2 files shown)


yarikoptic:
> I introduced a regression in #739, and it was not flagged by the tests.

for my own sake -- what was there the regression, since I can't catch what was done differently besides making a loop over a list?

it is ok to complement that test indeed.

bpinsard:
The changes in #739 meant to apply the metadata only on the newly converted files. This was done by passing the list of converted files to `external.dlad.mark_sensitive` and checking against the glob patterns provided, but the glob pattern gives anat folders, not the files in there. While git-annex is able to set metadata recursively when filtering with the files, none of the folder pattern were kept. So it ended up not adding metadata to these files. Globbing on the files in `anat` folder solves that.

yarikoptic:
gy - #793 to fix #739 ;-) Thanks!

bpinsard:
symmetrical PR ;) Thanks!
I might be the main user of the datalad sensitive marking.

yarikoptic:
nope -- just today I had to reset it for defaced images we pushed for https://github.com/spatialtopology/ds005256/ which we are yet to finalize into publishable state on openneuro ;)

PR #799, opened 2024-10-30
Add Python 3.12 into github testing CI

yarikoptic:


yarikoptic:
3.13 -- is something scary

```
  import traits.ctraits
E   ImportError: /opt/hostedtoolcache/Python/3.13.0/x64/lib/python3.13/site-packages/traits/ctraits.cpython-313-x86_64-linux-gnu.so: undefined symbol: Py_TRASHCAN_SAFE_BEGIN
```

let's tame the horses a little.

Issue #455, opened 2020-05-22
nibabel can't read my data and heudiconv doesn't explain why


Terf:
### Summary

When I run
```
docker run -v $PWD:/base nipy/heudiconv \
-d "/base/Data/{subject}-{session}/Hopkins/dicom/*/*.dcm" \
-o /base/Nifti/ -f convertall -s 01 -ss 001 -c none --overwrite
```
I get an error that `nibabel.nicom.dicomwrappers.WrapperError: File contains more than one StackID. Cannot handle multi-stack files`. Would anyone know what a `StackID` even is? I can't find much about it either in the heudiconv docs or elsewhere online, but the stack trace indicates the problem occurs on [line 105](https://github.com/nipy/heudiconv/blob/1f6ae74f7f06c0e7d2702623794eff62a3bd50ca/heudiconv/dicoms.py#L105) in the heudiconv source where you can see a try-catch block, but only KeyErrors are caught. So, I tried catching all errors by using sed to remove ` KeyError`:
```
docker run -v $PWD:/base --entrypoint="" nipy/heudiconv bash -c \
"sed -i 's/ KeyError//g' /src/heudiconv/heudiconv/dicoms.py && heudiconv -d '/base/Data/{subject}-{session}/Hopkins/dicom/*/*.dcm' -o /base/Nifti/ -f convertall -s 01 -ss 001 -c none --overwrite"
```
which does bypass the error in that particular spot but moves the problem downstream so it simply fails a bit later in the run. I’ve never seen this issue with other datasets I’ve converted, so I assume the issue is actually with the data, not the code, but I have no idea what it could be. Does anyone else? Ultimately I hope to submit a PR that catches the nibabel exception and raises our own exception with more of an explanation.

### Platform details:

Choose one:
- [ ] Local environment

- [x] Container: nipy/heudiconv (latest tag as of 5/22/20)

- Heudiconv version: 0.8.0

yarikoptic:
I am away from laptop ATM, but if you try with --minmeta (check for correct spelling in --help) would it work?
Any chance you could share a sample of such dicoms?

Terf:
I tried again with the `--minmeta` option (the help says that is the correct spelling), but got the same multi-stack files error. I will ask again but unfortunately I almost certainly cannot share any data. 

[...]

PR #800, opened 2024-10-31
Require nibabel >= 5.3.1 (fixes issues with enhanced DICOMs), and drop Python 3.8 support 

bpinsard:
to avoid a bunch of enhanced dicoms issues, notably through dcmstack. 
eg. 
```
dcmstack.dcmmeta.InvalidExtensionError: The extension is not valid: Missing required base classification time
``` 
should be fixed.

### DIFF
--- a/heudiconv/info.py
+++ b/heudiconv/info.py
@@ -11,7 +11,6 @@ CLASSIFIERS = [
     "Environment :: Console",
     "Intended Audience :: Science/Research",
     "License :: OSI Approved :: Apache Software License",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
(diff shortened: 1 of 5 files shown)


bpinsard:
looks like nibabel drop 3.8 support, should heudiconv  do the same? @yarikoptic

yarikoptic:
> looks like nibabel drop 3.8 support, should heudiconv do the same? @yarikoptic

I guess we could.
- https://github.com/nipy/heudiconv/pull/801

heudiconv/info.py
```diff
@@ -27,7 +27,7 @@
     "dcmstack>=0.8",
     "etelemetry",
     "filelock>=3.0.12",
-    "nibabel",
+    "nibabel>=5.3.1",
```
yarikoptic, line 30: remind me -- why we did not have that issue before? if that is just a buggy singular nibabel version - we better just exclude that one

bpinsard: Forgot to note the version that was causing this issue before upgrading, but has been there for a while https://github.com/nipy/heudiconv/issues/455#issuecomment-633105835 and my environment wasn't that old.  Also we had https://github.com/nipy/nibabel/pull/1378 recently merge that solved failing tests.

yarikoptic: well, this is a fresh version released not even a month ago. Some distributions might not have it yet, and would cause troubles.   I would just keep in mind that a recent nibabel is needed for processing enhanced dicoms?

May be even get some "catch all" handler that if exception happens and nibabel is below this version -- recommend upgrading?

Issue #576, opened 2022-06-13
Filenames in scans.tsv have `.nii.gz` extension although converted to uncompressed `.nii`


catetrai:
### Summary

When converting with output format `.nii`, the filenames listed in `scans.tsv` have `.nii.gz` extension.

The files themselves are correctly converted to uncompressed NIFTIs (thanks to the dcm2niix `compress` JSON option, see #365 ). So the only problem is in the `scans.tsv` file listing.

As a result, the BIDS validator fails with error `SCANS_FILENAME_NOT_MATCH_DATASET` (code 129): "The filename in scans.tsv file does not match what is present in the BIDS dataset."

### Platform details:

Choose one:
- [ ] Local environment

- [x] Container: `nipy/heudiconv:v0.11.3`

- Heudiconv version: 0.11.3

PR #802, opened 2024-11-11
[WIP] Handle dcm2niix not compressing nifti files

### LINKED ISSUE #365: Invalid data when outputting uncompressed NIfTIs
When setting a key to `outtype=('nii',)`, heudiconv executes successfully but produces invalid data.

A workaround is to use the `--dcmconfig` flag with the following config option:
```
{
    "compress": "n"
}
```

octomike:
So, we were hit by https://github.com/rordenlab/dcm2niix/issues/124, when converting a *very* long epi scan.

It seems that:

+ dcm2niix still refuses to compress (pigz or not) niftis > 4GB
+ heudiconv does not realize the file is now `.nii` instead of `.nii.gz` and produces invalid file extensions after moving
+ we can make dcm2niix happy by supplying `dcmconfig.json` with `"compress": "n"`, but heudiconv is oblivious about that change

I added some code that tries to catch these corner cases in `convert.py`

+ it warns the user if the conversion failed to compress and changes outtype to `nii`
+ also changes the outtype to `nii` if the user supplied `dcmconfig` as stated above
+ removed some hard coded `nii.gz` file extension snippets in `bids.py` and infer the effective extension by checking if a bids_file (json) has an accompanying `nii` or `nii.gz` file

I think this should also fix #365 and #576.

Tests were only run manually with a `dcmconfig` file and without, and with source data of mixed file length (>4GB and <4GB). Conversion works fine, validator is happy and `scans.tsv` now contain `.nii` and `.nii.gz` respectively.

But I am very nervous about this, because I assume other side effects by just switch `outtype` from `nii.gz` to `nii` mid conversion.

### DIFF
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -491,7 +491,7 @@ def save_scans_key(
     for bids_file in bids_files:
         # get filenames
         f_name = "/".join(bids_file.split("/")[-2:])
-        f_name = f_name.replace("json", "nii.gz")
+        f_name = f_name.replace("json", outtype_from_bidsfile(bids_file))
         rows[f_name] = get_formatted_scans_key_row(item[-1][0])
         subj_, ses_ = find_subj_ses(f_name)
         if not subj_:
@@ -938,9 +938,9 @@ def select_fmap_from_compatible_groups(
     # acq_times for the compatible fmaps:
     acq_times_fmaps = {
         k: acq_times[
-            # remove session folder and '.json', add '.nii.gz':
+            # remove session folder and '.json', add outtype:
             remove_suffix(remove_prefix(v[0], sess_folder + op.sep), ".json")
-            + ".nii.gz"
+            + '.' + outtype_from_bidsfile(v[0])
         ]
         for k, v in compatible_fmap_groups.items()
     }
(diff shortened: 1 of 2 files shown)


yarikoptic:
as for this particular PR: have you considered simply adding code which would just add a check that produced by dcm2niix .nii.gz is actually not compressed, and if so -- double check that it is ok'ish (opens up using nibabel and can access last volume? just to ensure that it is not totally borked conversion exiting early with `1`... just to be safe), rename/compress from within heudiconv python process using [gzip](https://docs.python.org/3/library/gzip.html) "battery included"?

octomike:
> Hi @octomike I totally forgot about that
> 
>     * [(Silently) produces .nii (not .nii.gz) despite   -z i   switch rordenlab/dcm2niix#124](https://github.com/rordenlab/dcm2niix/issues/124)
> 
> 
> I wonder -- did you try `dcm2niix` built with `-dmyDisableGzSizeLimits` as @neurolabusc suggested us to try (in 2017!)? I think I might just try to enable that flag for neurodebian builds of dcm2niix

I just did and the warning (`Warning: Saving uncompressed data: internal compressor unable to process such large files.`) went away, but it still produced uncompressed nifits (silently now).

After digging through the dcm2niix code a litte, it *appears* as if https://github.com/rordenlab/dcm2niix/commit/3c7f26be#diff-210e732ea7e4240e3ca012e3668da51f0e389a975fbc5b0edab430afbe27e314R9926 disabled pigz compression altogether, but that's [the only code path](https://github.com/rordenlab/dcm2niix/blob/master/console/nii_dicom_batch.cpp#L5188) to compress something if this define is set.

What a mess :see_no_evil:

octomike:
> rename/compress from within heudiconv python process using [gzip](https://docs.python.org/3/library/gzip.html) "battery included"?

That sounds much more reasonable than passing uncompress `nii` forward. I'll try to create a different PR with this approach.

PR #803, opened 2024-11-12
[ENH] Recompress files where dcm2niix failed

octomike:
As threatened in #802 an alternative approach.

+ try to open **and read** the nifit to  verify it wasn't truncated
+ recompress with gzip
+ fail if we couldn't - as opposed to just emit invalid file names (at least fmriprep fails to handle nii.gz that are not gz at all)

I decided to leverage the fact that heudiconv blindly moves all converted output files to `nii.gz`, even if they are just `nii` by renaming the successfully re-compressed image to `.nii`. This feels ugly, but results in the shortest/least-invasive PR.

### DIFF
--- a/heudiconv/convert.py
+++ b/heudiconv/convert.py
@@ -628,6 +628,19 @@ def convert(
                         item_dicoms, prefix, with_prov, bids_options, tmpdir, dcmconfig
                     )
 
+                    # try to handle compression failures from dcm2niix
+                    if outtype == 'nii.gz':
+                        converted_files = res.outputs.converted_files
+                        if not isinstance(converted_files, list):
+                            converted_files = [converted_files]
+                        uncompressed = [x for x in converted_files if x.endswith('.nii')]
+                        if len(uncompressed) > 0:
+                            lgr.warning("Conversion returned uncompressed nifti (>4GB?) - "
+                                        "trying to salvage by recompressing ourselves. "
+                                        "This might take a while ")
+                            if not recompress_failed(uncompressed):
+                                raise RuntimeError("Error compressing nifti")
+
                     bids_outfiles = save_converted_files(
                         res,
                         item_dicoms,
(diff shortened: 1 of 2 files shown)


heudiconv/convert.py
```diff
@@ -630,6 +630,19 @@ def convert(
...
+                        converted_files = res.outputs.converted_files
+                        if not isinstance(converted_files, list):
+                            converted_files = [converted_files]
+                        uncompressed = [x for x in converted_files if x.endswith('.nii')]
+                        if len(uncompressed) > 0:
+                            lgr.warning("Conversion returned uncompressed nifti (>4GB?) - "
+                                        "trying to salvage by recompressing ourselves. "
+                                        "This might take a while ")
+                            if not recompress_failed(uncompressed):
+                                raise RuntimeError("Error compressing nifti")
```
yarikoptic, line 644: why to delay this crash here?  this way you are hiding from user information on individual file and corresponding error -- why not just to just call `recompress` here and let error (if any happens) to bubble up?

I would also just loop here and have `recompress` to operate on a single file.

FWIW, I think it would be worth adding a test scenario to test this code path. I would have just mock patched the `nipype_convert` so that I would decompress some output file(s) and thus triggering this code path.

octomike: Thanks, I'll look into it!

yarikoptic: Hi @octomike , have time to finish this one up?

octomike: Wow, completely forgot. Will definitely look again

octomike: Took me just 6 months..

I addressed your suggestions and also added two test cases as you described (disclaimer: those were generated in a claude session - I don't know the policy here, but I'm open to redo those manually). The test cases check that recompress_failed actually recompresses the non-gz nifti returned from a mocked nipype_convert and also assert an error when loading a truncated/corrupt file because in the implementation I validate the uncompressed files are actually okay with nibabel.load().get_fdata().

yarikoptic:
does it remain the need, as dcm2niix fails to compress large files? then we would need to get back to it and de-slop AI abit: de-dup in the tests, move imports up, etc.  Up for it @octomike ?

PR #806, opened 2024-11-29
[ENH] filter dicoms earlier to avoid nibabel crash with XA ill-formed mutli-planar localizer dicoms

bpinsard:
see https://github.com/nipy/nibabel/issues/1392#issuecomment-2508066687

Apparently some of the validation of the dicom triggers nibabel.nicom that crashes on ill-formed dicoms.
Moving the dicom filter a bit before, allow to easily filter these with the heuristics `filter_dicom`. 
I should not change anything to the results.

### DIFF
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -169,6 +169,9 @@ def validate_dicom(
     Parse DICOM attributes. Returns None if not valid.
     """
     mw = dw.wrapper_from_file(fl, force=True, stop_before_pixels=True)
+    if dcmfilter is not None and dcmfilter(mw.dcm_data):
+        lgr.warning("Ignoring %s because of DICOM filter", fl)
+        return None
     # clean series signature
     for sig in ("iop", "ICE_Dims", "SequenceName"):
         try:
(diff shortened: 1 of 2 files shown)


dangom:
Affected by this as well. Any change this could be merged?!

hbraunDSP:
I am also affected by this; adding my voice to the merge requests.

Are we waiting on test coverage?

heudiconv/heuristics/reproin.py
```diff
@@ -226,7 +226,7 @@ def _delete_chars(from_str: str, deletechars: str) -> str:
 
 def filter_dicom(dcmdata: dcm.dataset.Dataset) -> bool:
     """Return True if a DICOM dataset should be filtered out, else False"""
-    return True if dcmdata.StudyInstanceUID in dicoms2skip else False
+    return True if dcmdata.get("StudyInstanceUID") in dicoms2skip else False
```
yarikoptic, line 229: why this change was necessary?  (the others in the file are just formatting, and should not be needed)

heudiconv/dicoms.py
```diff
@@ -203,6 +194,15 @@ def validate_dicom(
     except AttributeError:
         lgr.info("File {} is missing any StudyInstanceUID".format(fl))
         file_studyUID = None
+    if dcmfilter is not None and dcmfilter(mw.dcm_data):
+        lgr.warning("Ignoring %s because of DICOM filter", fl)
+        return None
```
yarikoptic, line 199: I am a little confused -- it seems that you moved filtering **later** not earlier!

I do see that you made it so it clears/deletes signatures now **later** than filtering is done, so would that suffice -- just to move deletion of signatures and keeping the rest of the code as is?

yarikoptic (changes requested):
Thanks for the PR! Sorry it takes awhile to get to them.

I left some questions to clarify the changes.  Also I wonder if there is an easy way for a test, even if it would just be a smoke test to show that it crashed before and no longer? ;)

PR #816, opened 2025-03-05
Fixup script for anonymization

yarikoptic:
TODOs
- [ ] make git-annex invocations optional/autodetected based on presence of .git/annex
- [ ] review git-annex behavior -- I think due to dotfiles. we might want some explicit setting to be tuned up in git annex config

### DIFF
--- a/utils/anon-cmd
+++ b/utils/anon-cmd
@@ -1,6 +1,7 @@
 #!/bin/bash
 # Generic anonymization script which would anonymize sid based on what it had
 # seen in the past or simply what the translation dict already has.
+# TODO: make git-annex use optional
 
 set -eu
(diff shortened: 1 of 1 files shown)


yarikoptic:
@claude review the script modified by this PR - its purpose is to maintain and update mapping for subject ids to new sequentially growing ids.  Then 
- tune to use git-annex  only if it is a git-annex repo (e.g. when there is `.git/annex`)
- if file starts with period, make script ensure that `annex.dotfiles` is set to true (`git annex config --set annex.dotfiles true` )
- add testing (could use e.g. bats or even code in python via pytest etc) that mapping happens as expected and file contains desired mapping

Issue #814, opened 2025-02-23
Siemens XA30: heudiconv fails to handle partial volumes


mvdoc:
Whenever a run is manually halted, Siemens XA30 generates a valid enhanced DICOM with fewer slices. If that happens, dcm2niix splits the series into a nifti with the full volumes and another nifti with the partial volume (see https://github.com/rordenlab/dcm2niix/issues/742). Currently, heudiconv doesn't handle this case, and crashes because it tries to overwrite the intact volume with the partial volume.

Possible solutions:
1. Since dcm2niix will flag the partial volumes as derived, a solution would be to allow calling dcm2niix with the flag `-i y`, which ignores all derived series. But this is not really a great solution, since there are cases where we may want to convert also derived series.
2. dcm2niix adds a `derived` string under the bidsguess key in the output json. We could check if that string exists, but I guess it would work only when heudiconv is run with the flag `--bids`
3. The partial volumes are valid nifti files with fewer slices than prescribed. If multiple files are generated from the same series, we could maybe add a check on the number of slices to detect the partial volumes.

bpinsard:
I see 3 options to fix that issue:
- a heuristic specific `dicom_filter` that checks dicoms tags are reported in the issue above, needs to be used for all XA data
- heuristics to allow specifying per-series the parameters for dcm2niix call: in that case `-i y` .
- a more global fix that would remove the partial volumes after dcm2niix conversion in `convert.save_converted_files`, partial volumes would be detected if the same series produced 2 files, and only one with `"RawImage": False` metadata. see #817

mvdoc:
Closing this as #817 fixed this issue (and I also just verified that heudiconv 1.3.4 works on a dataset with partial volumes). Thanks again @bpinsard!

PR #817, opened 2025-03-13
[FIX] remove XA partial volumes, generated by dcm2niix as derived data

bpinsard:
Attempt at fixing #814 , following new dcm2niix strategy to fix https://github.com/rordenlab/dcm2niix/issues/742 

TODO:
- [x] add tests data ( I have phantom data that I can add here).


yarikoptic:
could it be some really minimal phantom data, like masterfully interrupted 2 volumes with 1 having 2 slices and another 1? ;)  if not -- may be we should add to https://datasets-tests.datalad.org/ -- just share somewhere and I would add there

heudiconv/convert.py
```diff
@@ -922,6 +923,42 @@ def save_converted_files(
...
+        # be merged/postprocessed later
+        bids_files = (
+            sorted(bids_files)
+            if len(bids_files) == len(res_files)
+            else [None] * len(res_files)
+        )
+        # preload since will be used in multiple spots
+        bids_metas = [load_json(b) for b in bids_files if b]
+
+        # interrupted scans on XA: see dcm2niix #742
```
yarikoptic, line 936: this code block is worth moving into some helper function e.g. `handle_partial_volumes(bids_files, bids_metas)` which would potentially mutate those structures.

heudiconv/convert.py
```diff
@@ -1040,7 +1067,7 @@ def save_converted_files(
         safe_movefile(res_files, outname, overwrite)
         if isdefined(res.outputs.bids):
             try:
-                safe_movefile(res.outputs.bids, outname_bids, overwrite)
+                safe_movefile(bids_files, outname_bids, overwrite)
```
yarikoptic, line 1070: would need to double-check on what is the effect here and why e.g. we did not use bids_files to start with here, or why checking `isdefined` for res.output.bids and not bids_files to be defined...

mvdoc:
thank you for this @bpinsard . I tried to run it on the dataset that gave the error, but it still fails. This is a multiecho (3 echoes) dataset, and the weird thing is that the partial volumes for echoes 1 and 2 get recognized as partial with your PR, but not echo 3.

unfortunately I cannot share the dataset and it will take me a while to get some phantom data

bpinsard:
Thanks for the feedback, I will acquire some phantom multi-echo tomorrow to have test data. Are you using the CMRR sequence?

mvdoc:
yep. thanks @bpinsard !

bpinsard:
Well obviously removing items from the lists while iterating on it was a bad idea, but was only a problem with more than 2 converted files, such as with multi-echo. Works on the multi-echo phantom data I got this morning. 
Refactored to a separate function, too.

mvdoc:
I just tried on my dataset, it works great 💯

heudiconv/convert.py
```diff
@@ -851,6 +851,57 @@ def nipype_convert(
...
+    nii_files
+        filtered niftis
+    bids_files
+        filtered BIDS jsons
+    bids_metas
+        filtered BIDS metadata
+
+    """
+    partial_volumes = [not metadata.get("RawImage", True) for metadata in bids_metas]
+    no_partial_volumes = not any(partial_volumes) or all(partial_volumes)
```
yarikoptic, line 881: sorry, my had hurts a little trying to do this `bool` and need for that `or`. Wouldn't it make it easier to talk in terms of 

```suggestion
    has_partial_volumes = any(partial_volumes)
```
? (and then flip `if` below)  or it wouldn't be sufficient?

heudiconv/convert.py
```diff
@@ -851,6 +851,57 @@ def nipype_convert(
...
+    -------
+    nii_files
+        filtered niftis
+    bids_files
+        filtered BIDS jsons
+    bids_metas
+        filtered BIDS metadata
+
+    """
+    partial_volumes = [not metadata.get("RawImage", True) for metadata in bids_metas]
```
yarikoptic, line 880: so, a partial volume is iff `RawImage = False`? i.e. is every volume with `RawImage = False` is a partial volume?

heudiconv/convert.py
bpinsard: I am not sure if "RawImage":False is also present for other derived data, so I tried to narrow down the case where a series produced niftis with both False and not present (or True if ever). 
There might still be derived data that we want to convert.

yarikoptic:
fwiw -- typing seem needs fixing

bpinsard:
- typing error present in main.
- tests setup broken for git-annex install.

heudiconv/convert.py
yarikoptic: FTR, `datalad foreach-dataset -r --output-streams relpath   git grep RawImage` across our datasets found  251 of .json files among which
- 241 of _rec-moco files (motion corrected on the scanner I assume).  **So we should not just skip those AFAIK!**
- other 10 are mix of T1w and 1 dwi without having __dup in name with some __dup (very likely to be the canceled ones)

surprised that we didn't see that many of __dup files which should be the interrupted runs.  But likely because conversion was done with earlier versions of dcm2niix so it didn't detected as many cases before

heudiconv/convert.py
```diff
@@ -849,6 +849,61 @@ def nipype_convert(
...
+        filtered niftis
+    bids_files
+        filtered BIDS jsons
+    bids_metas
+        filtered BIDS metadata
+
+    """
+    partial_volumes = [
+        not metadata.get("RawImage", True)
+        and metadata.get("SeriesNumber", 0) > 1000
```
yarikoptic, line 880: could you please add a comment above to point to current dcm2niix code which relies on this heuristic.

Meanwhile, I filed

- https://github.com/rordenlab/dcm2niix/issues/972

PR #818, opened 2025-03-17
Do not create README if there is other allowed README.md or alike

yarikoptic:


heudiconv/bids.py
```diff
@@ -196,6 +196,8 @@ def populate_bids_templates(
         op.join(path, "README"),
         "TODO: Provide description for the dataset -- basic details about the "
         "study, possibly pointing to pre-registration (if public or embargoed)",
+        # TODO: get from schema
+        glob_suffixes=[".md", ".txt", ".rst"],
```
jwodder, line 200: Shouldn't you also add the empty string as a possible suffix?

yarikoptic: d'oh, right -- there could be indeed a suffix less match, good catch!

PR #823, opened 2025-05-08
DOC: fix the description of filter_files

mslw:
I might be confused, but the docs say "Any files where this function returns True will be filtered out." and I am convinced that it is the opposite.

The filter_files is used in the reproin heuristic (where it always returns True), with this docstring: "Return True if a file should be kept, else False". This commit adjusts the docs to effectively state the same (it used to say the opposite).

The docstring in the heudiconv heuristic was right: filter_files is fed to `dicoms.py::group_dicoms_into_seqinfos()` as `file_filter`, where it is applied with built-in `filter()` to the to-be-processed file list; `filter()` yields items for which `function(item) == True`.

N.B. This change makes it clear that `filter_files` and `filter_dicom` have opposite logic. The latter is fed to
`dicoms.py::group_dicoms_into_seqinfos()` as `dcmfilter`; `dcmfilter` is passed into `dicoms.py::validate_dicom()`; validate_dicom returns `None` (meaning: *not valid*) if `dcmfilter(mw.dcm_data)` is `True`.

So for a file to be kept, `filter_files` must return `True`, and `filter_dicom` must return `False`. This commit changes the documentation rather than the logic, assuming that the logic has been present long enough to warrant preservation (it is also easier to only change the docs).

### DIFF
--- a/docs/heuristics.rst
+++ b/docs/heuristics.rst
@@ -83,7 +83,8 @@ But it is not used directly by HeuDiConv.
 A utility function used to filter any input files.
 
 If this function is included, every file found will go through this filter. Any
-files where this function returns ``True`` will be filtered out.
+files where this function returns ``False`` will be filtered out (files where
+this function returns ``True`` will be kept).
 
 --------------------------
 ``filter_dicom(dcm_data)``
--- a/docs/heuristics.rst
+++ b/docs/heuristics.rst
@@ -85,6 +85,8 @@ A utility function used to filter any input files.
 If this function is included, every file found will go through this filter. Any
 files where this function returns ``False`` will be filtered out (files where
 this function returns ``True`` will be kept).
+Note: this function's logic is opposite to ``filter_dicom`` (see
+below). Think of it as an input for Python's built-in ``filter()``.
 
 --------------------------
 ``filter_dicom(dcm_data)``


docs/heuristics.rst
```diff
@@ -83,7 +83,8 @@ But it is not used directly by HeuDiConv.
 A utility function used to filter any input files.
 
 If this function is included, every file found will go through this filter. Any
-files where this function returns ``True`` will be filtered out.
+files where this function returns ``False`` will be filtered out (files where
+this function returns ``True`` will be kept).
```
yarikoptic, line 87: oh... Thank you @mslw !  I would like to go one more step and make an explicit note, smth like

```suggestion
this function returns ``True`` will be kept).
**Note: mention the opposite meaning with `filter_dicom`.**
```

WDYT?

yarikoptic (approved):
Thanks again!  For this PR we are all set, just let's decide on extra note.

But going forward -

- https://github.com/nipy/heudiconv/issues/824

interested to take on it @mslw ?

docs/heuristics.rst
mslw: How about simply saying it like this?

```suggestion
this function returns ``True`` will be kept).
Note: this function's return value is interpreted conversely to ``filter_dicom`` (see below).
```

mslw: Feel free to commit either of those, or anything else that sounds good to you!

mslw:
> But going forward -
> 
>     * [add handling of `filter_dicoms` to parallel `filter_files` and deprecated `filter_dicom` #824](https://github.com/nipy/heudiconv/issues/824)
> 
> 
> interested to take on it @mslw ?

Thanks! I like the idea of making them both consistent, but for now I have to pass - too much happening at once. Maybe the next time I'm running heudiconv, unless somebody else tackles that before.

PR #825, opened 2025-07-09
Use ubuntu-22.04 until we fixup NeuroDebian APT for debian and ubuntu

yarikoptic:


yarikoptic:
type checking is unrelated

PR #830, opened 2025-08-29
Sanitize "locator" value

yarikoptic:
We observed the Study Description to contain spaces in it. Due to unresolved bug in nipype (https://github.com/nipy/nipype/issues/3604) we must avoid spaces in the file names since then external execution of dcm2niix would fail.  While at it, I decided to sanitize it more and replace all other "funny" characters with the special treatment by shells.


yarikoptic:
Do you rely on locator values extracted from DICOMs StudyDescription, @nipy/team-heudiconv ? 
if so -- please verify that this PR doesn't affect behavior for you!

I am planing to merge/release within a week or so

PR #846, opened 2025-12-18
Add enhanced DICOM support with CLI integration and tests

Dhananjhay:
**Summary**

This PR adds support for Enhanced DICOM datasets, enabling robust metadata extraction and CLI-level integration for workflows that rely on modern multi-dimensional DICOM formats.

**Note**: This PR is intended to be merged after the sanitize-feature PR and will be rebased accordingly once that PR lands.

**Key changes**

Introduces an enhanced DICOM metadata extraction module
Adds --use-enhanced-dicom CLI option and threads it through the codebase
Refactors enhanced DICOM handling into independent, testable functions
Supports extraction of multi-dimensional parameters (e.g., multiple echo times)
Fixes TR/TE extraction from Enhanced DICOM functional groups
Adds comprehensive integration tests for enhanced DICOM workflows
Applies formatting and linting cleanups where relevant

**Dependency on prior PR**

This PR is logically independent but touches adjacent code paths affected by the sanitize-feature PR. To keep each change reviewable:
The sanitize-feature PR is submitted first
This PR will be rebased on top of it once merged

**Testing**

Added integration tests covering --use-enhanced-dicom
Existing test suite passes
Tests cover CLI wiring and metadata extraction paths

### DIFF
--- a/heudiconv/dicom/enhanced.py
+++ b/heudiconv/dicom/enhanced.py
@@ -272,14 +272,15 @@ def create_seqinfo_enhanced(
     TR = float(TR) if TR is not None else 0.0
     empty_list = []
     if TE is None:
-        if "SharedFunctionalGroupsSequence" in ds:
+        try:
             TE = ds.SharedFunctionalGroupsSequence[0].MREchoSequence[0].EffectiveEchoTime
-        elif "PerFrameFunctionalGroupsSequence" in ds:
-            for i in range(0,len(ds.PerFrameFunctionalGroupsSequence)):
-                    for j in range(0, len(ds.PerFrameFunctionalGroupsSequence[i].MREchoSequence)):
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -11,3 +11,4 @@ build/
 dist/
 sample_nifti.json
 venvs/
+heudiconv/_version.py
```
yarikoptic: unrelated, not needed, should not be here

akhanf:
Thanks @yarikoptic, we will actually retract this back to the drawing board as we've discovered a number of issues.. I have a better idea of enhanced dicoms now so alot of this will be re-conceptualized. 

Overall the issues we are trying to resolve are for certain enhanced dicoms: 1) metadata (eg seqinfo) not extracted properly, and 2) nifti conversion not working properly. I am thinking a solution to 1) could be a PR here (some pointers on how to do this in the cleanest way would be helpful!), and 2) either try to tackle in dcm2niix, or make an alternative optional converter.

PR #848, opened 2026-01-15
BF: support _ses-DATE as alternative to _ses-{date} for Siemens XA60

yarikoptic:
Siemens X60 software does not allow {} characters in sequence names, so _ses-{date} cannot be used. This adds support for _ses-date (without braces) which behaves identically - resolving to the acquisition date.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/heuristics.rst
+++ b/docs/heuristics.rst
@@ -43,7 +43,7 @@ The following details of the sequences could also be used as a ``{detail}`` in t
 
 - ``item``: an index of seqinfo (e.g., ``1``),
 - ``subject``: a subject label (e.g., ``qa``)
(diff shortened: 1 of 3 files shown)


jcf2:
Note: X60 --> XA60. 

It is looking like this is a change due to the "Numaris X" platform that the XA releases are built on, so would apply to other XA releases as well -- and is an intentional change and not a bug.

bpinsard:
I never used that feature; I didn't even knew it existed.
I am not even sure to understand what is does from the docs: one can set the sequence on the Numaris console to contain `{date}` placeholder that is then in the dicoms SeriesDescription/ProtocolName or in the Patient registration PatientName or PatientID fields ?
How does that relate to series keys?

Shouldn't that be in the `infotoids` section rather than `infotodict` as this is more to determine session value?

PR #851, opened 2026-02-24
[ENH] add MultipleLabels IntendedFor method

octomike:
This is a suggestion for yet another IntendedFor method. We had some success using `PlainAcquisitionLabel` merged a while back, but there are a number of cases where our site-wide heuristic requires more than that.

We often get duplicated files using the prescan normalization method of our scanner. Currently we name it `rec-norm` (I'm aware of [this beast](https://github.com/bids-standard/bids-specification/pull/105), we decided to use rec to be bids valid for now). The issue with this however is that we cannot base our IntendedFor matching decision solely on `acq`, because we have two groups now:

```
fmap/sub-foo_acq-se_dir-AP_rec-norm_epi.json
fmap/sub-foo_acq-se_dir-PA_rec-norm_epi.json
fmap/sub-foo_acq-se_dir-AP_epi.json
fmap/sub-foo_acq-se_dir-PA_epi.json
```

and 

```
func/sub-foo_task-t_acq-se_rec-norm_run-01_bold.json
func/sub-foo_task-t_acq-se_rec-norm_run-02_bold.json
func/sub-foo_task-t_acq-se_run-01_bold.json
func/sub-foo_task-t_acq-se_run-02_bold.json
```

And as I understand, fmaps with `rec-norm` are intended only for the corresponding functionals with `rec-norm` .

We could introduce a second label `acq-senorm` and use that, however, the entity modification with `rec-` happens way after our protocol modifications (we heavly rely on `reproin.fix_dbic_protocol()` and a set of regexes.

---

Using a combination of `acq` and `rec`  as they key would help us out here. I decided to create a slightly more generic approach with this PR, instead of just introducing `PlainAcquisitionRecLabel` or something similar.

Using `MultipleLabels`, a heuristic user can now base the decision on one ore more entities, configurable with a new paramter `parameter_opts` in `POPULATE_INTENDED_FOR_OPTS`. (I added an example to the docs)

It's mildly invasive, because we have to carry these options along in `bids.py` - however, I tried to make sure that nothing breaks with `reproin` or any other heuristic that is unaware of this new option.

### DIFF
--- a/docs/heuristics.rst
+++ b/docs/heuristics.rst
@@ -184,9 +184,15 @@ The parameters that can be specified and the allowed options are defined in ``bi
    * ``'PlainAcquisitionLabel'``: similar to ``'CustomAcquisitionLabel'``, but does not change
      behavior for ``func`` modality and always bases decision on the ``_acq-`` label. Helps in
      cases when there are multiple tasks and a shared ``fmap`` for some of them.
+   * ``'MultipleLabels'``: in cases where an ``_acq-`` label is not sufficient to distinguish
+     fmap groups. Useful in scenarios where ``_rec-`` is automatically assigned. Configure the
+     list of entities to consider in the ``'parameter_opts'`` dictionary.
+     ``''``
    * ``'Force'``: forces ``heudiconv`` to consider any ``fmaps`` in the session to be
      suitable for any image, no matter what the imaging parameters are.
 
+ - ``'parameter_opts'``: Some ``'matching_parameters'`` like ``'MultipleLabels'`` can be
+   configured using this dictionary.
 
  - ``'criterion'``: Criterion to decide which of the candidate ``fmaps`` will be assigned to
    a given file, if there are more than one. Allowed values are:
@@ -195,10 +201,16 @@ The parameters that can be specified and the allowed options are defined in ``bi
    * ``'Closest'``: The closest in time to the beginning of the image acquisition.
 
 .. note::
-  Example::
+  Examples::
 
     POPULATE_INTENDED_FOR_OPTS = {
             'matching_parameters': ['ImagingVolume', 'Shims'],
+            'criterion': 'Closest'
+    }
+
+    POPULATE_INTENDED_FOR_OPTS = {
+            'matching_parameters': ['MultipleLabels'],
+            'parameter_opts': {'MultipleLabels': {'entities': ['acq', 'rec']}},
             'criterion': 'Closest'
     }
 
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -78,6 +78,7 @@ AllowedFmapParameterMatching = [
     "ModalityAcquisitionLabel",
     "CustomAcquisitionLabel",
     "PlainAcquisitionLabel",
+    "MultipleLabels",
     "Force",
 ]
 # Key info returned by get_key_info_for_fmap_assignment when
@@ -697,7 +698,9 @@ def find_fmap_groups(fmap_dir: str) -> dict[str, list[str]]:
 
 
 def get_key_info_for_fmap_assignment(
-    json_file: str, matching_parameter: str
+    json_file: str,
+    matching_parameter: str,
+    parameter_opts: dict = {},
 ) -> list[Any]:
     """
     Gets key information needed to assign fmaps to other modalities.
@@ -710,6 +713,9 @@ def get_key_info_for_fmap_assignment(
         path to the json file
     matching_parameter : str in AllowedFmapParameterMatching
         matching_parameter that will be used to match runs
+    parameter_opts : dict
+        possible options for the matching_parameter
+        default: {}
 
     Returns
     -------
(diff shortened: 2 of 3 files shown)


octomike:
Many tests are currently failing because of [requests](https://github.com/psf/requests/commit/4bd79e397304d46dfccd76f36c07f66c0295ff82) and [chardet](https://github.com/chardet/chardet/releases) - I guess a release is imminent.

heudiconv/bids.py
```diff
@@ -762,6 +771,14 @@ def get_key_info_for_fmap_assignment(
...
         plain_label = BIDSFile.parse(op.basename(json_file))["acq"]
         key_info = [plain_label]
+    elif matching_parameter == "MultipleLabels":
+        # use a heuristic specific list of labels and match on the
+        # concatenation of their values
+        combined_labels = ""
+        for entity in parameter_opts.get("entities", []):
+            label = BIDSFile.parse(op.basename(json_file))[entity]
+            combined_labels += label or ""
+        key_info = [combined_labels or None]
```
yarikoptic-gitmate, line 781: Concatenating the entity values without a separator can produce false matches. Two files with different entities collapse to the same key:

- `fmap/sub-x_acq-ab_rec-c_epi.json` → `"abc"`
- `func/sub-x_acq-abc_bold.json` → `"abc"`

I confirmed these are treated as compatible even though `acq`/`rec` differ. Since BIDS label values are alphanumeric, joining with a character that can't appear in a label (e.g. `"_"`) removes the ambiguity:

```python
key_info = ["_".join(
    BIDSFile.parse(op.basename(json_file))[e] or "" for e in parameter_opts.get("entities", [])
) or None]
```

Minor: `BIDSFile.parse(...)` is called once per entity inside the loop — worth hoisting it out of the loop.

---

heudiconv/tests/test_bids.py
```diff
@@ -197,6 +198,30 @@ def test_get_key_info_for_fmap_assignment(
             json_name, matching_parameter="PlainAcquisitionLabel"
         )
 
+    # 8) matching_parameter = 'MultipleLabels'
+    A_LABEL = gen_rand_label(label_size, label_seed)
+    B_LABEL = gen_rand_label(label_size, label_seed)
```
yarikoptic-gitmate, line 203: `A_LABEL` and `B_LABEL` are generated with the same `label_seed`, so `gen_rand_label` returns the **same** string for both (it re-`seed()`s each call). I verified they're identical (e.g. both `PCI7`).

Because of this, the cases below that are meant to distinguish `acq` vs `rec` (the `BOTH` row, and the `dwi`/`anat` rows) don't actually exercise distinct values, and the concatenation order/collision behavior isn't tested. Using two different seeds so `A_LABEL != B_LABEL` would make the test meaningfully stronger — and, combined with a separator, would catch the collision noted in `bids.py`.

---

docs/heuristics.rst
```diff
@@ -184,9 +184,15 @@ The parameters that can be specified and the allowed options are defined in ``bi
    * ``'PlainAcquisitionLabel'``: similar to ``'CustomAcquisitionLabel'``, but does not change
      behavior for ``func`` modality and always bases decision on the ``_acq-`` label. Helps in
      cases when there are multiple tasks and a shared ``fmap`` for some of them.
+   * ``'MultipleLabels'``: in cases where an ``_acq-`` label is not sufficient to distinguish
+     fmap groups. Useful in scenarios where ``_rec-`` is automatically assigned. Configure the
+     list of entities to consider in the ``'parameter_opts'`` dictionary.
+     ``''``
```
yarikoptic-gitmate, line 190: Stray empty `` `` `` `` here (renders as a literal `''`) — looks like a leftover editing artifact. Should be removed.

---

yarikoptic-gitmate (commented):
Thanks for this — the `MultipleLabels` matcher is a clean, backward-compatible extension. I checked out the branch, ran the fmap/IntendedFor tests (all green), and confirmed the new test passes. The design threads `parameter_opts` through the whole call chain sensibly, and because `convert.py` / `main.py` already invoke `populate_intended_for(session_path, **opts)`, no changes were needed there for the new option to flow through. Coverage on the new lines looks good.

A few things to consider, in rough priority order (details inline):

**1. Concatenation without a separator can cause false matches (correctness, edge case).**
`combined_labels += label or ""` means `acq-ab_rec-c` and `acq-abc` both produce `"abc"` and are treated as compatible. Joining entity values with a separator that can't appear in a BIDS label (e.g. `"_"`) fixes this. See inline comment on `bids.py`.

**2. The two test labels are identical (test strength).**
`A_LABEL` and `B_LABEL` share `label_seed`, so `gen_rand_label` returns the same string for both — the cases meant to distinguish `acq` from `rec` don't actually test distinct values. Using two different seeds (plus the separator above) would let the test catch the collision. See inline comment on `test_bids.py`.

**3. Docs artifact.** A stray `''` line in `heuristics.rst` (leftover editing artifact) should be removed.

**4. `TypedDict` not updated (minor/optional).** `PopulateIntendedForOpts` in `convert.py:58-60` could add `parameter_opts: dict[str, dict]` to stay in sync with the new signature. Not blocking — the `typing` job passes as-is since heuristic dicts aren't type-checked at their source.

**5. Style nit.** New annotations use `None | dict` / bare `dict`, while the surrounding file uses `Optional[...]` and parametrized `dict[str, dict]`. Worth matching the existing convention (e.g. `Optional[dict[str, dict]]`).

On CI: the 3.10–3.13 failures are unrelated to this PR (upstream `requests`/`chardet`, as you noted) — `3.9`, `lint`, `typing`, and `codecov` all pass. Rebasing onto current `master` should let the matrix re-run cleanly.

None of these are hard blockers; (1) and (2) are the ones I'd most encourage addressing before merge.

---

yarikoptic:
Sorry @octomike I missed this PR entirely. I had looked at it and your problem and came up with no better alternative :-/ so let's finalize it?  test should pass now, please update. 

[here](https://claude.ai/code/session_01W97uuHHr5d3HGfqfyKSWse) is claude's review, overall positive too but has a few suggestions worth considering

yarikoptic:
@octomike move from draft when it is ready for the next round

PR #855, opened 2026-04-10
Fix fmap rec dir ordering

octomike:
In some of our converted datasets the bids-validator is complaining about files like this:

```
sub-01/ses-01/fmap/sub-01_ses-01_dir-PA_rec-norm_epi.nii.gz
```

It turns out that this is actually what the reproin heuristic returns right now and it is invalid because of the incorrect position of `rec` ([must be followed](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#case-4-multiple-phase-encoded-directions-pepolar) by `dir`).
It used to be invalid anyway (rec- not part of fmap) and might have slipped that way for PEpolar fmaps.

The PR is slightly refactoring the `rec-moco` workaround, but not touching it's logic. I also added tests (using claude).

### DIFF
--- a/heudiconv/heuristics/reproin.py
+++ b/heudiconv/heuristics/reproin.py
@@ -596,15 +596,6 @@ def infotodict(
             # if there is no _run -- no run label added
             run_label = None
 
-        # yoh: had a wrong assumption
-        # if curr_seqinfo.is_motion_corrected:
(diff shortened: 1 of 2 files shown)


yarikoptic:
Thank you @octomike !

yarikoptic:
FWIW -- look into annotating claude work even more -- check out my commits/PRs ;)  coming to OHBM?

PR #860, opened 2026-06-15
Ensure all IntendedFor matching params must match

### LINKED ISSUE #859: With multiple IntendedFor matching parameters, only the last one matters
When defining `POPULATE_INTENDED_FOR_OPTS`, it is possible to specify `matching_parameters` with multiple values (e.g. `["CustomAcquisitionLabel", "Shims"]`. I assume this is supposed to be treated as a union (AND / ALL), although it could also be (less intuitively) ANY.

However, if I'm not mistaken, in practice, *only the last one* applies.

In `find_compatible_fmaps_for_run`, the inner loop (over `matching_parameters`) ends with a conditional `continue` statement -- but being at the end of the loop, it has no effect (it continues with the next iteration of the innermost loop). The inline comment "don't bother checking more params" suggests it should have been a `break` statement instead.
https://github.com/nipy/heudiconv/blob/7e9e9a6b75b5825f2ad481716e54aa5fe0896326/heudiconv/bids.py#L776-L833

As a consequence, only the `compatible` (boolean) value from the last inner loop iteration makes it to the conditional in the outer loop (over field map groups), and while all parameters are tested, only the last one effectively matters.

With `break` instead of `continue`, the check would stop at the first incompatible parameter, and the operation over `matching_parameters` would be ALL_OF.

Note on multiple values: I used `["CustomAcquisitionLabel", "Shims"]` for a dataset in which acquisition labels should be clear enough, and added shims as a security measure - to say "I'd rather have no matched field map than one with wrong shims". Then I noticed that the pairing was done ignoring the labels.

I have a PR (with a test) ready.

mslw:
The list of `matching_parameters` specified in `POPULATE_INTENDED_FOR_OPTS` can have multiple values. If I'm not mistaken, the code would test all of them, but then act only on the result of the *last test*, which seems unintended.

This assumes that `matching_parameters` are to be interpreted as `all()`, and makes the check behave accordingly.

It seems that the matching functions have so far only been tested with a single parameter. This PR adds a test for `find_compatible_fmaps_for_run_union` being called with two parameters, and a set of files for which there is one pairing which matches both, but more which match just one. The assumption for the test is that the desired operation is `all()`, which seems the more logical alternative to me.

Fixes #859

### DIFF
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -826,7 +826,7 @@ def find_compatible_fmaps_for_run(
                     for x, y in zip(json_info[param], fm_info)
                 )
             if not compatible:
-                continue  # don't bother checking more params
+                break  # don't bother checking more params
         if compatible:
             compatible_fmap_groups[fm_key] = fm_group
 
--- a/heudiconv/tests/test_bids.py
+++ b/heudiconv/tests/test_bids.py
@@ -1112,6 +1112,83 @@ def test_find_compatible_fmaps_for_run(
             assert compatible_fmaps == expected_compatible_fmaps[json_file]
 
 
+def test_find_compatible_fmaps_for_run_union(monkeypatch: pytest.MonkeyPatch) -> None:
+    """
+    Test find_compatible_fmaps_for_run with multiple parameters.
+
+    This tests the behavior of find_compatible_fmaps_for_run when it
+    is given more than one parameter to match on. Expects multiple
+    parameters to be treated as a union (AND).
+    """
+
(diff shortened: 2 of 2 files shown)


yarikoptic:
Sounds/looks logical but not sure how we missed that in the original implementation  or through the use.  @pvelasco please have a look

mslw:
I'm likely not adding much here, but I admit that seeing that the code in question was long unchanged gave me pause, but in the end didn't change my conclusion.

yarikoptic:
It is a fix, so `patch` is appropriate but I want to signal an importance of it with `minor`!

PR #864, opened 2026-07-29
Aggregate events.tsv for invalid entities (and rec)

### LINKED ISSUE #849: Separate events.tsv files are created for both part-mag and part-phase fMRI images, failing bids-validator
We save phase images for fMRI and organize these in BIDS, using the `part-mag` and `part-phase` tags. Heudiconv creates separate `*_events.tsv` files for both magnitude and phase images, which are detected by bids-validator and raise an error. I believe having just a single `events.tsv` with no `part-*` tag is the is the correct behavior. If someone points me in the right direction I can try to fix it and do a quick PR.

octomike:
We currently get two (empty) events.tsv files for our `rec-norm` and `rec-raw` bold pairs. This is redundant and should be aggregated into an events file without a rec entity altogether (the spec allows it, but I would argue that events are independent of reconstruction methods, right?).

I had a look at the code and cleaned it up a little to utilize `BIDSFile.parse()` instead of manual splits around `_echo`. I also realized that there are other entities valid for the bold, but invalid for the corresponding events.tsv and remove them. For that I added a `drop()` and a trivial `__contains__()` method to BIDSFile to easily remove an entity from the filename and have concise tests.

This should also fix #849 

---

I believe the `__contains__`  is an intuitive add-on, but it might lead to confusion with code like this:

```python
bidsfile = BIDSFile.parse(bold)
if 'rec' in bidsfile:
  ...
```

Under the hood it will now check if `rec` is in the `_entities` dictionary, but it might read as: *Is `rec` literally in the string representation of the filename?* I'm not sure if you want to avoid this little trap.

### DIFF
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -259,7 +259,7 @@ def populate_aggregated_jsons(path: str) -> None:
         for invalid_entity in ['chunk', 'echo', 'part']:
             # events.tsv with these entities are not specified
             if invalid_entity in parsed_fpath:
-                parsed_fpath.drop(invalid_entity, silent=True)
+                parsed_fpath.drop(invalid_entity)
         if 'rec' in parsed_fpath:
             # `rec` is technically valid for the events file, but the events
             # are independent of the reconstruction method and so we drop them
@@ -1151,10 +1151,9 @@ class BIDSFile:
     def __contains__(self, entity: object) -> bool:
         return entity in self._entities
 
-    def drop(self, entity: str, silent: bool = False) -> None:
+    def drop(self, entity: str) -> None:
         if entity not in self._entities:
-            if not silent:
-                lgr.warning("File %s does not contain entity '%s'", self.__str__(), entity)
+            lgr.warning("File %s does not contain entity '%s'", self.__str__(), entity)
             return
         self._entities.pop(entity)
 
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -1147,8 +1147,10 @@ class BIDSFile:
         return entity in self._entities
 
     def drop(self, entity: str, missing_ok: bool=False) -> None:
-        if not missing_ok and entity not in self._entities:
-            raise ValueError(f"{self} does not contain entity {entity!r}")
+        if entity not in self._entities:
+            if not missing_ok:
+                raise ValueError(f"{self} does not contain entity {entity!r}")
+            return
         self._entities.pop(entity)
 
     def set(self, entity: str, value: str, overwrite: bool = True) -> None:
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -256,15 +256,9 @@ def populate_aggregated_jsons(path: str) -> None:
 
         # specify the name of the '_events.tsv' file:
         parsed_fpath = BIDSFile.parse(op.basename(fpath))
-        for invalid_entity in ['chunk', 'echo', 'part']:
+        for events_invalid_entity in ['chunk', 'echo', 'part']:
             # events.tsv with these entities are not specified
-            if invalid_entity in parsed_fpath:
-                parsed_fpath.drop(invalid_entity)
-        if 'rec' in parsed_fpath:
-            # `rec` is technically valid for the events file, but the events
-            # are independent of the reconstruction method and so we drop them
-            # as well
-            parsed_fpath.drop('rec')
+            parsed_fpath.drop(events_invalid_entity, missing_ok=True)
         fpath = op.join(op.dirname(fpath), str(parsed_fpath))
         events_file = remove_suffix(fpath, suf) + "_events.tsv"
         # do not touch any existing thing, it may be precious
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -1076,6 +1076,7 @@ class BIDSFile:
         "mt",
         "part",
         "recording",
+        "chunk",
     ]
 
     def __init__(
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -1145,10 +1145,9 @@ class BIDSFile:
     def __contains__(self, entity: object) -> bool:
         return entity in self._entities
 
-    def drop(self, entity: str) -> None:
-        if entity not in self._entities:
-            lgr.warning("File %s does not contain entity '%s'", self.__str__(), entity)
-            return
+    def drop(self, entity: str, missing_ok: bool=False) -> None:
+        if not missing_ok and entity not in self._entities:
+            raise ValueError(f"{self} does not contain entity {entity!r}")
         self._entities.pop(entity)
 
     def set(self, entity: str, value: str, overwrite: bool = True) -> None:
--- a/heudiconv/bids.py
+++ b/heudiconv/bids.py
@@ -253,27 +253,19 @@ def populate_aggregated_jsons(path: str) -> None:
         # create a stub onsets file for each one of those
         suf = "_bold.json"
         assert fpath.endswith(suf)
-        # specify the name of the '_events.tsv' file:
-        if "_echo-" in fpath:
-            # multi-echo sequence: bids (1.1.0) specifies just one '_events.tsv'
-            #   file, common for all echoes.  The name will not include _echo-.
-            # TODO: RF to use re.match for better readability/robustness
-            # So, find out the echo number:
-            fpath_split = fpath.split("_echo-", 1)  # split fpath using '_echo-'
-            fpath_split_2 = fpath_split[1].split(
-                "_", 1
-            )  # split the second part of fpath_split using '_'
-            echoNo = fpath_split_2[0]  # get echo number
-            if echoNo == "1":
-                if len(fpath_split_2) != 2:
-                    raise ValueError("Found no trailer after _echo-")
-                # we modify fpath to exclude '_echo-' + echoNo:
-                fpath = fpath_split[0] + "_" + fpath_split_2[1]
-            else:
-                # for echoNo greater than 1, don't create the events file, so go to
-                #   the next for loop iteration:
-                continue
 
+        # specify the name of the '_events.tsv' file:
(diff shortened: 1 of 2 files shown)


heudiconv/bids.py
```diff
@@ -253,27 +253,19 @@ def populate_aggregated_jsons(path: str) -> None:
...
+        parsed_fpath = BIDSFile.parse(op.basename(fpath))
+        for invalid_entity in ['chunk', 'echo', 'part']:
+            # events.tsv with these entities are not specified
+            if invalid_entity in parsed_fpath:
+                parsed_fpath.drop(invalid_entity)
+        if 'rec' in parsed_fpath:
+            # `rec` is technically valid for the events file, but the events
+            # are independent of the reconstruction method and so we drop them
+            # as well
+            parsed_fpath.drop('rec')
```
yarikoptic: NB see the last suggestion for the ultimate one ;)

thought about it and I see your logic but, I think we better keep it close to BIDS spec

```suggestion
        # specify the name of the '_events.tsv' file:
        parsed_fpath = BIDSFile.parse(op.basename(fpath))
        for invalid_entity in ['chunk', 'echo', 'part']:
            # events.tsv with these entities are not specified
            if invalid_entity in parsed_fpath:
                parsed_fpath.drop(invalid_entity)
```

because
- BIDS [does allow for it](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#task-including-resting-state-imaging-data), e.g. (hypothesizing) could be separate annotations of 'rec' specific aspects (e.g. bad moco volumes etc)
- unclear to me if tools are "ready" to do that shaving off `_rec-` entity to somewhat follow IP to load events (since again -- spec allows for the entity)

but even if we decided to strip it, why it shouldn't just be added to `invalid_entity` listing above?

```suggestion
        # specify the name of the '_events.tsv' file:
        parsed_fpath = BIDSFile.parse(op.basename(fpath))
        for invalid_entity in ['chunk', 'echo', 'part', 'rec']:
            # events.tsv with these entities are not specified
            if invalid_entity in parsed_fpath:
                parsed_fpath.drop(invalid_entity)
```

Also while at it (stripping `rec` here), I would prefer a better suited name for variable

```suggestion
        # specify the name of the '_events.tsv' file:
        parsed_fpath = BIDSFile.parse(op.basename(fpath))
        for events_invalid_entity in ['chunk', 'echo', 'part']:
            # events.tsv with these entities are not specified
            if events_invalid_entity in parsed_fpath:
                parsed_fpath.drop(events_invalid_entity)
```

and with that tune up to drop

```suggestion
        # specify the name of the '_events.tsv' file:
        parsed_fpath = BIDSFile.parse(op.basename(fpath))
        for events_invalid_entity in ['chunk', 'echo', 'part']:
            # events.tsv with these entities are not specified
            parsed_fpath.drop(events_invalid_entity, missing_ok=True)
```

heudiconv/bids.py
```diff
@@ -1156,6 +1148,15 @@ def __setitem__(
     ) -> None:  # would puke with some exception if already known
         return self.set(entity, value, overwrite=False)
 
+    def __contains__(self, entity: object) -> bool:
+        return entity in self._entities
+
+    def drop(self, entity: str) -> None:
+        if entity not in self._entities:
+            lgr.warning("File %s does not contain entity '%s'", self.__str__(), entity)
+            return
```
yarikoptic: Let's make it a bit more versatile but also more explicit -- warnings are largely useless in particular where it is more of a programming shortcoming, hence let's make into  exception

```suggestion
    def drop(self, entity: str, missing_ok: bool=False) -> None:
        if not missing_ok and entity not in self._entities:
            raise ValueError(f"{self} does not contain entity {entity!r}")
```

yarikoptic (changes requested):
THANK YOU for the PR -- we should fix this indeed. BUT I think we should not sanitize `rec`, and also tune code a bit

heudiconv/bids.py
yarikoptic, line 1078: also claude mentioned that we are missing chunk here (to make code below actually work for it ;) ) . Note that likely because it is not part of the diff -- suggestion seems to not work here neatly -- please add manually

```suggestion
        "recording",
        "chunk",
```

heudiconv/bids.py
```diff
@@ -253,27 +253,19 @@ def populate_aggregated_jsons(path: str) -> None:
...
-                # we modify fpath to exclude '_echo-' + echoNo:
-                fpath = fpath_split[0] + "_" + fpath_split_2[1]
-            else:
-                # for echoNo greater than 1, don't create the events file, so go to
-                #   the next for loop iteration:
-                continue
 
+        # specify the name of the '_events.tsv' file:
+        parsed_fpath = BIDSFile.parse(op.basename(fpath))
+        for invalid_entity in ['chunk', 'echo', 'part']:
```
yarikoptic: see my comment about 'chunk' -- we need to add it!

heudiconv/bids.py
yarikoptic: <details>
<summary>here and elsewhere -- claude made a few more (selected) of useful IMHO</summary> 

```shell

Minor / nits
4. Not black-clean. CI's lint job runs only flake8 + codespell, but the project also configures black via pre-commit (.pre-commit-config.yaml). The new code uses single quotes (['chunk', 'echo', 'part'], 'rec', ['dir', 'not_existing']); black 23.3.0 will rewrite these to double quotes. So CI is green but pre-commit run black would produce a diff. Worth running before merge for consistency with the rest of the file.
...

7. drop()'s warning calls self.__str__(), which raises ValueError if sub is absent. Same latent issue already present in set(), so consistent and low-priority — just noting it.

8. Test gap. test_populate_aggregated_jsons_events covers rec and part but not echo/chunk — and echo is exactly the path this PR rewrote. Adding an echo-1/echo-2 case (and a chunk case) would lock in the multi-echo behavior the old code guarded.
 
```
</details>

[link to full](https://claude.ai/code/session_012MXZKkuitzmXLSNQSiWjx8)

yarikoptic-gitmate:
One heads-up when applying the suggestions above: turning `drop()` into an exception is a **two-file** change — the existing `test_BIDSFile` currently asserts a *warning* is logged on a missing-entity drop, so it will start failing once `drop()` raises.

<details>
<summary>What needs updating in <code>test_bids.py</code></summary>

The current block relies on the warning behavior:

```python
caplog.set_level(logging.WARNING)
for entity in ['dir', 'not_existing']:
    caplog.clear()
    my_bids_file.drop(entity)
    assert len(caplog.records) == 1
    assert f"does not contain entity '{entity}'" in caplog.records[0].message
```

With `drop(entity, missing_ok=False)` raising `ValueError`, this should become something like:

```python
for entity in ["dir", "not_existing"]:
    with pytest.raises(ValueError, match=f"does not contain entity {entity!r}"):
        my_bids_file.drop(entity)
    # and the missing_ok escape hatch is a no-op:
    my_bids_file.drop(entity, missing_ok=True)
```

The `import logging` / `caplog` fixture then become unused and can be dropped.
</details>

Also worth adding an `echo` (and now `chunk`) case to `test_populate_aggregated_jsons_events`, since the multi-echo path is exactly the code being rewritten here.

---

octomike:
Hey @yarikoptic - thank you so much for that in-depth review. I feel this was much more work than what I did initially with the PR :sweat_smile: 

Your and Claude's points are very valid and I applied most of them verbatim. Fixed the tests as well.

It's just unfortunate that the reason I dug into this is now off the table with rec- being untouched. But, well, worth it anyway!

yarikoptic:
thanks @octomike !   So, you are still having multiple `_rec-` files for the same sequence and it is not just a matter of removing `_rec-*_events.tsv` ones while leaving also existing ones without `_rec-`?  I wonder if in follow up PR we could add some bids-specific option like `norec-events` which would alter behavior here to the one you desire ... ;-)

github actions is ATM a bust , so I will run tests locally but also wait for a day or so

yarikoptic:
re RTD -- if you update branch on top of master (merge or rebase), should render ok

octomike:
> thanks @octomike ! So, you are still having multiple `_rec-` files for the same sequence and it is not just a matter of removing `_rec-*_events.tsv` ones while leaving also existing ones without `_rec-`? I wonder if in follow up PR we could add some bids-specific option like `norec-events` which would alter behavior here to the one you desire ... ;-)

Yeah, I'll think about it. I'm inclined to just have this artifact fixed after conversion now that the possible neat solution is off the table.  Will check with the "customer" (researcher).

> github actions is ATM a bust , so I will run tests locally but also wait for a day or so

Don't sweat it.

heudiconv/bids.py
```diff
@@ -1156,6 +1143,16 @@ def __setitem__(
     ) -> None:  # would puke with some exception if already known
         return self.set(entity, value, overwrite=False)
 
+    def __contains__(self, entity: object) -> bool:
+        return entity in self._entities
+
+    def drop(self, entity: str, missing_ok: bool=False) -> None:
```
yarikoptic, line 1149:
```suggestion
    def drop(self, entity: str, missing_ok: bool = False) -> None:
```

to please linters

yarikoptic: hm, I need to sleep on it but it says

Image

may be you didn't allow contributors to push to you branch and that's the reason @octomike ?

yarikoptic:
Ran the current head (`121ad10`) locally: tests pass, but the `lint` job will fail once Actions is back — `flake8` flags E252 on the `drop()` signature:

```
heudiconv/bids.py:1149:49: E252 missing whitespace around parameter equals
heudiconv/bids.py:1149:50: E252 missing whitespace around parameter equals
```

One-liner: `def drop(self, entity: str, missing_ok: bool=False)` → `missing_ok: bool = False`. (Ironically that came in verbatim from the suggestion — my bad for typing it un-spaced.)

---

heudiconv/bids.py
yarikoptic: ping on above  -- could you please accept the fix?

PR #866, opened 2026-08-06
DOC: add minimal CLAUDE.md pointing at CONTRIBUTING.rst, and refresh CONTRIBUTING

yarikoptic-gitmate:
Draft for discussion.

## `CLAUDE.md` (new, 19 lines)

Follows the pattern used in [`datalad/datalad`](https://github.com/datalad/datalad/blob/master/CLAUDE.md): rather than duplicating conventions into a second file that will drift, it just tells the assistant to go read the project's contributing guide, and lists what it will find there. Note heudiconv's guide is `CONTRIBUTING.rst`, not `.md`, so the pointer references the `.rst`; no rename was done.

## `CONTRIBUTING.rst`

The pointer is only useful if the guide actually carries the conventions it defers to, so this brings it up to date:

**Corrections**
- "We support Python 3 only (>= 3.7)" → `>= 3.9`, per `heudiconv/info.py`; also notes the CI matrix is 3.9–3.13.
- "Lines are no longer than 120 characters" → line length is whatever `black` produces (88), since `flake8` is configured to ignore `E501`. The old number matched neither the formatter nor the linter.
- Fixed the broken relative link to `heudiconv/info.py` (it was written as link text pointing at a bare path).

**Previously undocumented tooling** — the guide told contributors to run `pytest` and nothing else, despite `.pre-commit-config.yaml`, `mypy.ini`, and `tox.ini` all being in the tree:
- New "Quick reference" block at the top with the handful of commands one actually needs.
- New "Code style" section covering `pre-commit` (black / isort / flake8 / codespell) and the fact that the codebase is fully type-annotated with `mypy` run in CI (`tox -e typing`).
- New "Testing" section: where the tests live, that `--doctest-modules` is on so doctests are executed, that some tests skip without `dcm2niix` / `git-annex` / `datalad` (so a green local run covers less than CI), and that `tox` reproduces the CI matrix.
- Mentioned the external tools HeuDiConv shells out to under "Development environment".

**AI-assisted contributions (new section)**
Requires `@pytest.mark.ai_generated` on any test written by an AI assistant. The marker is already registered in `tox.ini` and already in use at `heudiconv/tests/test_bids.py:1192` — it just was never written down anywhere a contributor would look. The section spells out when to apply it (assistant wrote the bulk of it, even if subsequently edited), how to filter on it (`pytest -m ai_generated` / `-m "not ai_generated"`), and *why* we care: AI-generated tests tend to encode current behavior rather than intended behavior, so they can pass while asserting the wrong thing — being able to find them again later is the point. Also asks contributors to disclose substantial AI use on the non-test parts in the PR description.

## Checks

- `CONTRIBUTING.rst` parses cleanly through docutils (no warnings at level 2+).
- `codespell` clean on both files.
- Docs-only change; no code touched.

## Open questions for reviewers

- Should the `ai_generated` requirement be stronger (e.g. a CI check), or is documenting it enough for now?
- Worth converting `CONTRIBUTING.rst` → `CONTRIBUTING.md` for consistency with `CHANGELOG.md` and the datalad layout? Deliberately left out of this PR to keep the diff reviewable.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JJ7SjBsZeYo7zzCmf69Vzs)_

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -107,10 +107,17 @@ Note that you will likely have these will already be available on your system if
 package manager (e.g. Debian's ``apt-get``, Gentoo's ``emerge``, or simply PIP) to install the
 software.
 
-HeuDiConv also calls out to a number of external tools, most notably ``dcm2niix``, which is not
-installable from PyPI and thus must be provided by your system (e.g. via NeuroDebian, conda, or
-`its own releases <https://github.com/rordenlab/dcm2niix/releases>`_).  Some tests additionally
-require ``git-annex`` and ``datalad``.
+HeuDiConv also calls out to a number of external tools, most notably ``dcm2niix``, and some tests
+additionally need ``git-annex`` and ``datalad``.  All of them are nowadays installable from PyPI,
+so a plain virtualenv is enough to get a complete development environment::
+
+  pip install dcm2niix git-annex
+
+`dcm2niix <https://pypi.org/project/dcm2niix/>`_ and
+`git-annex <https://pypi.org/project/git-annex/>`_ are wheels bundling the corresponding binaries
+for common platforms; ``datalad`` is pulled in by the ``datalad`` (and hence ``all``) extra of
+HeuDiConv itself.  You may of course still prefer to obtain them from your system package manager
+(e.g. NeuroDebian, conda, or Homebrew) if you already have those set up.
 
 Development work might require live access to the copy of HeuDiConv which is being developed.
 If a system-wide release of HeuDiConv is already installed, or likely to be, it is best to keep
@@ -153,6 +160,12 @@ them all against the whole tree at any point with ``pre-commit run -a``, and the
 
 Beyond what the tools enforce:
 
+* **Do not duplicate code.**  Copy-pasted logic is the single most reliable way to introduce bugs
+  into this codebase: the copies inevitably diverge, a fix lands in one of them and not the others,
+  and the discrepancy is then found by users rather than by us.  If you catch yourself
+  copy-pasting, factor the common part out into a helper instead — even for two occurrences, and
+  even when the copies differ in small ways (that is what arguments are for).  This applies with
+  equal force to tests, docs, and heuristics, not just to library code.
 * HeuDiConv is fully type-annotated and ships a ``py.typed`` marker.  New functions must have
   annotated arguments and return values; ``mypy`` is run in CI and can be run locally with
   ``tox -e typing``.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -81,13 +100,18 @@ The following pull request labels are respected:
...
-Dependencies which you will need are `listed in the repository <heudiconv/info.py>`_.
+Dependencies which you will need are listed in `heudiconv/info.py <./heudiconv/info.py>`_.
 Note that you will likely have these will already be available on your system if you used a
 package manager (e.g. Debian's ``apt-get``, Gentoo's ``emerge``, or simply PIP) to install the
 software.
 
+HeuDiConv also calls out to a number of external tools, most notably ``dcm2niix``, which is not
+installable from PyPI and thus must be provided by your system (e.g. via NeuroDebian, conda, or
+`its own releases <https://github.com/rordenlab/dcm2niix/releases>`_).  Some tests additionally
+require ``git-annex`` and ``datalad``.
```
yarikoptic: there is https://pypi.org/project/dcm2niix/ now ! and similarly even git-annex (along with datalad). Adjust accordingly

CONTRIBUTING.rst
```diff
@@ -112,19 +136,81 @@ To build the docs locally:
...
+
+    import pytest
+
+    @pytest.mark.ai_generated
+    def test_something() -> None:
+        ...
+
+Mark the test if the assistant wrote the bulk of it, even if you subsequently edited it; there is
+no need to mark a hand-written test that merely received an AI-suggested tweak.  For a
+parametrized test, place the marker on the test function alongside the ``parametrize`` decorators.
```
yarikoptic, line 233: make it stronger here to prefer parametrized tests where feasible since AI tends to duplicate instead!  State dislike of "duplication" in any form in the code since it breeds bugs/divergences.

yarikoptic (changes requested):

CONTRIBUTING.rst
yarikoptic-gitmate: Fixed in 02f6bb4 — my claim that `dcm2niix` isn't on PyPI was simply wrong. All three are wheels bundling the binaries, so the paragraph now shows `pip install dcm2niix git-annex` as the primary route (noting `datalad` already comes in via the `datalad`/`all` extra), and demotes NeuroDebian/conda/Homebrew to "if you already have those set up".

---

CONTRIBUTING.rst
yarikoptic-gitmate: Done in 02f6bb4, in three places since the concern is broader than the AI section:

1. **Code style** — a lead bullet stating plainly that duplicated code is not acceptable, with the reasoning you gave: copies diverge, a fix lands in one and not the others, and users find the discrepancy instead of us. Explicitly says "even for two occurrences, and even when the copies differ in small ways (that is what arguments are for)", and that it binds tests/docs/heuristics as much as library code.
2. **Testing** — requires a single parametrized test wherever cases differ only in inputs/expected outputs, with a short example, plus fixtures for shared setup.
3. **AI-assisted contributions** — calls out the tendency directly: assistants "emit a pile of copy-pasted test functions where one parametrized test would do", so before submitting you must collapse near-identical functions into one `parametrize`, hoist repeated setup, and drop cases that aren't actually distinct. Framed as cheaper up front than in review, and as a forcing function for the understanding the contributor is expected to have.

---

Issue #537, opened 2021-12-07
Failure when DICOM data does not have attribute 'AcquisitionDate'


tjhendrickson:
### Summary

<details>
<summary>log</summary> 

```
INFO: Doing conversion using dcm2niix
INFO: Converting /output_dir/sub-michaelmyers/ses-112921/anat/sub-michaelmyers_ses-112921_rec-normalized_run-01_T2w (208 DICOMs) -> /output_dir/sub-michaelmyers/ses-112921/anat . Converter: dcm2niix . Output types: ('nii.gz', 'dicom')
211207-14:57:18,631 nipype.workflow INFO:
	 [Node] Setting-up "convert" in "/tmp/dcm2niixqka1kfvu/convert".
INFO: [Node] Setting-up "convert" in "/tmp/dcm2niixqka1kfvu/convert".
211207-14:57:18,704 nipype.workflow INFO:
	 [Node] Running "convert" ("nipype.interfaces.dcm2nii.Dcm2niix"), a CommandLine Interface with command:
dcm2niix -b y -z y -x n -t n -m n -f sub-michaelmyers_ses-112921_rec-normalized_run-01_T2w_heudiconv770 -o /output_dir/sub-michaelmyers/ses-112921/anat -s n -v n /tmp/dcm2niixqka1kfvu/convert
INFO: [Node] Running "convert" ("nipype.interfaces.dcm2nii.Dcm2niix"), a CommandLine Interface with command:
dcm2niix -b y -z y -x n -t n -m n -f sub-michaelmyers_ses-112921_rec-normalized_run-01_T2w_heudiconv770 -o /output_dir/sub-michaelmyers/ses-112921/anat -s n -v n /tmp/dcm2niixqka1kfvu/convert
211207-14:57:20,239 nipype.interface INFO:
	 stdout 2021-12-07T14:57:20.239657:Chris Rorden's dcm2niiX version v1.0.20210317  GCC9.3.0 x86-64 (64-bit Linux)
INFO: stdout 2021-12-07T14:57:20.239657:Chris Rorden's dcm2niiX version v1.0.20210317  GCC9.3.0 x86-64 (64-bit Linux)
211207-14:57:20,240 nipype.interface INFO:
	 stdout 2021-12-07T14:57:20.239657:Found 208 DICOM file(s)
INFO: stdout 2021-12-07T14:57:20.239657:Found 208 DICOM file(s)
211207-14:57:20,240 nipype.interface INFO:
[...]

PR #867, opened 2026-08-06
BF: populate SeqInfo date/time from any available DICOM date/time tags

yarikoptic-gitmate:
Addresses https://github.com/nipy/heudiconv/issues/537 (in particular https://github.com/nipy/heudiconv/issues/537#issuecomment-5198647517).

## Problem

`SeqInfo.date` / `SeqInfo.time` were taken verbatim from `AcquisitionDate` (0008,0022) and `AcquisitionTime` (0008,0032):

```python
date=dcminfo.get("AcquisitionDate"),
time=dcminfo.get("AcquisitionTime"),
```

Those tags are optional, so both fields ended up `None` in `dicominfo.tsv` and in heuristics for

- XA30 enhanced DICOMs, which carry only `AcquisitionDateTime` (0008,002A) — the case reported in the linked comment;
- de-identified data where only `SeriesDate`/`SeriesTime` survived.

`heudiconv.dicoms.get_datetime_from_dcm()` already knows how to pick a datetime out of all three sources, but it was only used for tarball reproducibility (`get_reproducible_int`) and for BIDS `AcquisitionDateTime`, not when filling `SeqInfo`.

## Changes

- New `heudiconv.dicoms.get_datetime_strings_from_dcm()` returning `(date, time)` DICOM-style strings:
  - returns `AcquisitionDate`/`AcquisitionTime` as is whenever both are present (no change in behavior/format for the common case);
  - otherwise formats whatever `get_datetime_from_dcm()` establishes (`AcquisitionDateTime`, then `SeriesDate`+`SeriesTime`) as `%Y%m%d` / `%H%M%S.%f`. Both values come from the same source, so they cannot become mutually inconsistent;
  - returns `None` for whatever is unknown — no exception when the header was stripped, as requested in the issue discussion.
- `create_seqinfo()` uses it for the `date` and `time` fields.
- `get_datetime_from_dcm()` annotation widened from `dcm.FileDataset` to `dcm.Dataset` (it only needs `__contains__`/`__getitem__`; docstring already said so).

A side benefit: reproin's `session=DATE`/`{date}` support now works for scanners which only provide `AcquisitionDateTime`.

## Tests

Added to `heudiconv/tests/test_dicoms.py`, all marked `@pytest.mark.ai_generated`:

- `test_get_datetime_strings_from_dcm_acq_date_time` — plain `AcquisitionDate`/`AcquisitionTime` passed through unchanged;
- `test_get_datetime_strings_from_dcm_acq_datetime` — XA30 enhanced DICOM (no `AcquisitionDate`/`AcquisitionTime`) yields date and time matching `AcquisitionDateTime`;
- `test_get_datetime_strings_from_dcm_series_date_time` — fallback to `SeriesDate`/`SeriesTime`;
- `test_get_datetime_strings_from_dcm_wo_dt` — nothing available → `(None, None)`, no exception;
- `test_get_datetime_strings_from_dcm_partial` — only a date available → `(date, None)`;
- `test_create_seqinfo_date_time` — parametrized over all test DICOMs: `seqinfo.date`/`seqinfo.time` are populated and agree with `get_datetime_from_dcm()`.

Test run: `python -m pytest heudiconv --ignore=heudiconv/tests/test_regression.py` → 205 passed, 14 skipped, 1 failed. The single failure is `test_heuristics.py::test_partial_xa_conversion`, which fails identically on the unmodified base commit in this container (older `dcm2niix` 1.0.20220720). `mypy heudiconv` reports only the one pre-existing error in `heudiconv/heuristics/test_reproin.py`; `black`/`flake8` are clean on the changed hunks.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KA6v8mxbBXjUQwDmpSpg2R)_

### DIFF
--- a/heudiconv/dicoms.py
+++ b/heudiconv/dicoms.py
@@ -115,6 +115,8 @@ def create_seqinfo(
     global total_files
     total_files += len(series_files)
 
+    date, time = get_datetime_strings_from_dcm(dcminfo)
+
     custom_seqinfo_data = (
         custom_seqinfo(wrapper=mw, series_files=series_files)
         if custom_seqinfo
@@ -155,9 +157,9 @@ def create_seqinfo(
         # For demographics to populate BIDS participants.tsv
         patient_age=dcminfo.get("PatientAge"),
         patient_sex=dcminfo.get("PatientSex"),
-        date=dcminfo.get("AcquisitionDate"),
+        date=date,
         series_uid=dcminfo.get("SeriesInstanceUID"),
-        time=dcminfo.get("AcquisitionTime"),
+        time=time,
         custom=custom_seqinfo_data,
     )
 
@@ -517,12 +519,12 @@ def get_reproducible_int(dicom_list: list[str]) -> int:
     )
 
 
-def get_datetime_from_dcm(dcm_data: dcm.FileDataset) -> Optional[datetime.datetime]:
+def get_datetime_from_dcm(dcm_data: dcm.Dataset) -> Optional[datetime.datetime]:
     """Extract datetime from filedataset, or return None is no datetime information found.
 
     Parameters
     ----------
-    dcm_data : dcm.FileDataset
+    dcm_data : dcm.Dataset
         DICOM with header, e.g., as ready by pydicom.dcmread.
         Objects with __getitem__ and have those keys with values properly formatted may also work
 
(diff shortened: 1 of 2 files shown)


yarikoptic:
@tjhendrickson @Moo-Marc  please review this PR / code if would address your issue(s) reliably.

Moo-Marc:
Works for me. Thanks!

PR #869, opened 2026-08-06
feat: rename BIDS files for multi-orientation series (rebase of #788 + fixes)

yarikoptic-gitmate:
Takes @bpinsard's #788 (which adds a `chunk-` entity when dcm2niix splits one series into several images with different orientations, e.g. a 3-plane localizer), rebases it onto current master, and fixes four problems found reviewing it. The first two commits are #788 unchanged; the last four are the fixes. Supersedes #788 — please review there for the entity choice, here for the implementation.

Without this, such a series is silently written out as `sub-X_T1w1`, `sub-X_T1w2`, … — a mangled *suffix*, not valid BIDS.

<details>
<summary>The four fixes</summary>

**`BIDSFile` did not round-trip filenames.** `__str__` rebuilds a name from `_known_entities`, so whatever that list gets wrong comes out wrong, and #788 is the first production path to feed real output basenames through `parse`/`__str__`.
- The entities #788 added (`nuc`/`voi`/`trc`/`stain`/`tracksys`) were appended at the end, but the entity table interleaves them among `task-`, `acq-` and `ce-`. Filled the list in from `rules/entities.yaml`, in the order it mandates, plus heudiconv's own `ch` where `update_uncombined_name` puts it.
- Entities not in the list were dropped: `sub-1_space-T1w_desc-x_bold.nii.gz` → `sub-1_bold.nii.gz`. Now kept, before the suffix, with a warning.
- `parse` looked for `key-value` pairs anywhere and assumed the last one preceded the suffix. reproin marks duplicated series with a trailing `__dup-01`, so `sub-1_task-rest_bold__dup-01` lost `bold` entirely; a `T1w-mod` suffix contains a `w-mod` pair. Now matches the leading run of pairs as a whole.

**The `chunk-` index was arbitrary.** #788 indexed orientations by looking the file's `ImageOrientationPatientDICOM` up in the *lexicographically sorted* set of the series' orientations — a sort of the string rendering of a list of floats. On the localizer fixture that makes Cor `chunk-2` though it was acquired last, and it is not stable across subjects: any difference in obliquity or float formatting reshuffles it, so `chunk-1` could be the sagittal plane for one subject and the coronal one for the next, and the brain/c-spine multi-FoV BOLD case could swap the two FoVs between subjects. Now indexes the distinct orientations in file order, which is deterministic and still groups files that share an orientation (so a series split by echo or mag/phase as well gets one chunk per FoV, not per file).

**Two paths aborted the whole conversion.** A basename with no entities, or without `sub-`, raised out of `save_converted_files`; a sidecar missing `ImageOrientationPatientDICOM` while its siblings have it raised `KeyError`. Both now warn and leave the name alone.

**Colliding names.** dcm2niix splits on more than orientation, so a series can yield two images sharing one. `chunk-` cannot tell those apart, and because the name *was* modified the numeric fallback was skipped — two files then moved onto the same destination, which `safe_movefile` turns into a `RuntimeError` (or, with `--overwrite`, silent data loss). Output names are now computed up front and only used if they came out distinct.

Also: shrank the new test fixture (2.0MB → 508KB; full 512² pixel data buys nothing for a filename test) and emptied its free-text `InstitutionAddress`/`OperatorsName`/`ReferringPhysicianName`.

</details>

<details>
<summary>Notes on where this stands w.r.t. BIDS and dcm2niix</summary>

`chunk` was generalized from microscopy to MRI in BIDS 1.9.0 ([bids-specification#1586](https://github.com/bids-standard/bids-specification/pull/1586)) and its definition in 1.11.1 — "images of the same physical sample with different fields of view acquired in the same imaging experiment … may be used to indicate different anatomical structures or regions of the same structure" — describes this case exactly. It is allowed in `anat`, `func`, `dwi`, `fmap` and `micr`; **not** in `perf`, `pet` or `mrs`, where these names will not validate.

dcm2niix (checked with 1.0.20260724 from PyPI) offers nothing to align with: plain `-b y` disambiguates with a non-BIDS `_i0000<N>` postfix, and its own BIDS naming (`-f %h`/`%H`, both flagged experimental) appends that same postfix *after* the suffix. Its `BidsGuess` for this localizer is `["discard", "_acq-fl2_run-1_localizer"]`, and BIDS still has no `localizer`/`scout` suffix at all — so the localizer use case is out of BIDS regardless, and the entity is really there for the other multi-FoV series.

</details>

Known limitation, unchanged from #788: `is_multiorient` compares orientations as exact strings, so two files of the same plane differing in the last decimal would be read as multi-orient.

An AI assistant (Claude Code) was used substantially for the four fix commits and their tests; the two commits from #788 are @bpinsard's. Full suite green (209 passed, 15 skipped), mypy and lint clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015ZMpRTSaHJ3bnHFtSjehkT

---
_Generated by [Claude Code](https://claude.ai/code/session_015ZMpRTSaHJ3bnHFtSjehkT)_


yarikoptic-gitmate:
Brought this up to date with master (it was 12 commits / ~6 weeks behind). Clean merge, full suite green locally against current master: **225 passed, 15 skipped**. CI on the merge commit `3e50641` is green — `lint`, `typing`, and the whole test matrix.

**Correction.** An earlier revision of this comment claimed `tox -e lint` and `tox -e typing` were red on `master` itself and would go red here once CI re-ran. That was wrong, and CI has since disproved it. What I actually hit was my own environment: I ran `tox` on Python 3.11, while `lint.yml` and `typing.yml` pin the job to Python 3.9. Sorry for the noise — the `tox.ini` patch I proposed is not needed and should not be applied.

The mechanism, from `flake8_builtins.py`:

```python
if hasattr(sys, 'stdlib_module_names'):
    # stdlib_module_names is only available in Python 3.10+
    known_module_names = sys.stdlib_module_names
```

`A005` (module shadowing a stdlib name — `heudiconv/queue.py`) can therefore only fire on Python ≥ 3.10, and CI's 3.9 never sees it. Same shape for the mypy `[unreachable]` at `heudiconv/heuristics/test_reproin.py:138`: reachability there depends on the interpreter mypy targets, which defaults to the one it runs on.

Nothing to do for this PR. The only real (and minor) residue is that a contributor who follows `CONTRIBUTING.rst` and runs `tox` on a current Python — which is most defaults now — gets two failures CI does not, so `tox` is not reproducing CI. Worth a separate issue at most; I'm not touching it here.

**The `lint`/`typing` checks that read "cancelled" on the previous head** were a genuine and unrelated thing: they were the duplicate `push`-event runs of commit `fd839c1`, whose `pull_request` runs both passed ([Linters 536](https://github.com/nipy/heudiconv/actions/runs/31123958428), [Type-check 500](https://github.com/nipy/heudiconv/actions/runs/31123958376)). They queued for 1h40m and were killed without ever starting. #871 fixes that by cutting the duplicate at the trigger.

---

PR #870, opened 2026-08-17
Update dcm2niix bundled in the Docker image to v1.0.20260724

wojzwo:
Suggested label: `dependencies`

## Summary

For non-container use, users can install and update `dcm2niix` independently, but the Docker image was still pinned to `v1.0.20240202`.

This update brings the Docker image to the current dcm2niix release:

- `v1.0.20240202` → `v1.0.20260724`

The dcm2niix version is now defined once in `utils/gen-docker-image.sh`:

```bash
dcm2niix_version="v1.0.20260724"
```

and passed to Neurodocker from there, rather than embedding the version directly in the generator arguments.

`Dockerfile` was regenerated using:

```bash
./utils/gen-docker-image.sh
```

The rest of the Docker image configuration is unchanged.

## Validation

- Built the generated Docker image locally.
- Verified the resulting image reports `dcm2niix v1.0.20260724`.
- Verified `heudiconv`, `dcm2niix`, `pigz`, and `git-annex` are available in the resulting image.
- Ran the repository workflows on my fork:
  - type checking passed,
  - linting passed,
  - release workflow checks passed,
  - Docker image generation and build passed.
- The Docker workflow only failed at the publish step, as the upstream Docker registry credentials are not available in the fork.

## Possible follow-up

The Docker image is currently relatively large and slow to build. There may be
room to improve this separately by making the main image dependencies more
explicit and easier to update together with the rest of the repository.

In a separate experimental image build with an updated dependency stack and a
simplified environment setup, I reduced the local image size from about 2.31 GB
to about 1.06 GB (based on `docker image ls`). On a limited profiling workload,
that image was also around 6% faster, although I would treat that only as an
initial observation rather than a general performance claim.

If this is useful, I would be happy to explore it in a separate PR, for example
by:

- keeping the main Docker image dependency versions in a small manifest/config file,
- updating the currently old pinned components,
- evaluating a faster Conda solver such as `libmamba`,
- or potentially removing Conda from the runtime image if it is not required there.


yarikoptic:
Thank you @wojzwo

PR #872, opened 2026-09-17
Add acquisition duration tracking to BIDS _scans.tsv files

### LINKED ISSUE #875: populate-scans-duration (#872): duration anchored on a different reference point than acq_time -> spurious negative overlap for interleaved multiband series
## Context

While using PR #872's `populate-scans-duration` on the [`dbic/QA`](https://datasets.datalad.org/dbic/QA) phantom dataset (729 `_scans.tsv` files, ~9 years of near-identical automated sessions on the same Siemens Prisma), I found a specific, reproducible pattern: `acq_time + duration` implies a **negative** gap (an impossible overlap) to the next scan in the same session, for one particular sequence-transition pair only.

## The symptom

Across all 388 sessions containing the `dwi_acq-DTI-30-p2-s4` (multiband, `MultibandAccelerationFactor=4`) → `anat-T1w_acq-MPRAGE` transition, `189 (48.7%)` show `acq_time(MPRAGE) < acq_time(DTI) + duration(DTI)`, i.e. an apparent overlap. The overlap is small and tightly bounded (`-0.002s` to `-0.299s`), never anything larger. Example (`sub-qa/ses-20180212`):

```
dwi/..._acq-DTIX30Xp2Xs4_dwi.nii.gz   acq_time=09:19:09.475000  duration=111.582
anat/..._acq-MPRAGE_T1w.nii.gz        acq_time=09:21:00.757500
-> naive end = 09:21:01.057   vs   next acq_time = 09:21:00.7575  =>  gap = -0.299s
```

I checked this is **not** noise or a one-off: across the entire dataset (388 sessions × this pair, plus 87 other distinct sequence-transition pairs / ~3246 transitions total), this is the *only* pair that ever goes negative. Every other transition — even ones with a tiny ~2s baseline gap (e.g. `fa-10 → MPRAGE`, median 2.27s, std 0.41s, n=218) — stays comfortably positive across all instances.

## Root cause (confirmed via raw DICOM inspection, 6 sessions)

I pulled the raw DICOM tarballs (`sourcedata/sub-qa/ses-*/dwi/*_acq-DTIX30Xp2Xs4_dwi.dicom.tgz`, public at `datasets.datalad.org`) and looked at **every** instance's `AcquisitionTime`, not just the one heudiconv/dcm2niix designates as "the" series time. Result: in 5 of 6 sessions checked, `InstanceNumber=2` is acquired **2.585–2.588 seconds before** `InstanceNumber=1` (classic Siemens multiband slice interleaving — the file that ends up as instance #1 is not the chronologically-first-acquired slice). Filed a companion issue upstream for this: **rordenlab/dcm2niix#1039** (still present in the latest release, v1.0.20260724, tested directly).

The actual bug in *this* codebase is that PR #872's `duration` and the existing `acq_time` end up anchored on **two different reference points**:

- `acq_time` — [`get_formatted_scans_key_row()`](https://github.com/nipy/heudiconv/blob/claude/scans-tsv-duration-field-82tvmn/heudiconv/bids.py#L657) reads a single representative file, `dcm_fn_strs[0]` (whatever sorts first — in practice `InstanceNumber==1`), and calls `get_datetime_from_dcm()` on just that file:
  ```python
  dcm_data = dcm.dcmread(dcm_fn_strs[0], stop_before_pixels=True, force=True)
  acq_datetime = dicoms.get_datetime_from_dcm(dcm_data=dcm_data)
  ```
- `duration` — [`estimate_scan_duration_from_times()`](https://github.com/nipy/heudiconv/blob/claude/scans-tsv-duration-field-82tvmn/heudiconv/dicoms.py#L651) correctly scans **every** file's timestamp and computes the true span:
  ```python
  unique_timestamps = sorted(set(timestamps))
  ...
  median_interval = statistics.median(intervals)
  return (unique_timestamps[-1] - unique_timestamps[0]).total_seconds() + median_interval
  ```

`unique_timestamps[0]` here is the **true** earliest instant (Instance #2 in our case), while the `acq_time` column is anchored on **Instance #1**, which — per the dcm2niix interleaving quirk above — is 2.585 s *later*. So:

```
acq_time + duration
  = InstanceTime1 + (true_max − true_min + median_interval)
  = true_max + (InstanceTime1 − true_min) + median_interval
  = true_max + 2.585s + ~0.2s
```

That `≈2.79s` overshoot past the DTI's true end is what eats into the real (and otherwise perfectly ordinary, ~2.5s) gap to the next scan and flips it negative. I confirmed this arithmetic matches observation almost exactly in all 5 affected sessions checked (offset 2.585–2.588s, computed "true gap" using real per-instance min/max instead of the naive formula comes out to a consistent, healthy +2.49s — right in line with other same-family transitions like `fa-10 → MPRAGE`).

**One informative negative control**: `sub-qa/ses-20250324` uses a different (one-instance-per-volume, not one-instance-per-slice) DICOM layout for the same protocol — there, `InstanceNumber==1` *is* the true first instant (offset = 0.000s), confirming the offset really does come from per-slice interleaving, not from the transition pair itself. That session still shows a small residual `-0.257s`, pointing to a second, much smaller (~0.2–0.3s) discrepancy in `duration`'s own magnitude even when correctly anchored — likely just the `median_interval` term not perfectly matching the true final-frame width in this different layout; worth a look but far less impactful than the anchor-mismatch above.

## Suggested fix

Make `duration` anchored on **whatever timestamp ends up as `acq_time`**, not on the estimator's own internally-recomputed minimum — so `acq_time + duration` is guaranteed self-consistent (lands at the true last-acquired instant, regardless of which file dcm2niix/heudiconv happens to treat as "first"):

```python
def estimate_scan_duration_from_times(
    dicom_list: list[str], anchor_dt: Optional[datetime.datetime] = None
) -> Optional[float]:
    ...
    anchor = anchor_dt if anchor_dt is not None else unique_timestamps[0]
    return (unique_timestamps[-1] - anchor).total_seconds() + median_interval
```

and in `get_formatted_scans_key_row()`, pass the already-computed `acq_datetime` through `get_acquisition_duration()` into this function as `anchor_dt`. This is a small, self-contained change scoped entirely to the new PR #872 code — it doesn't touch `acq_time`'s value/semantics (no back-compat risk, no dependency on a dcm2niix fix landing first), and it eliminates the impossible-overlap artifact unconditionally.

**Optional, separate follow-up** (bigger, deliberately not bundled here): `acq_time` itself could be corrected to `min(unique_timestamps)` for multi-instance runs, fixing the underlying inaccuracy rather than just making `duration` consistent with it. That changes historical `acq_time` values for affected sequences by a couple of seconds, so it deserves its own PR/discussion — and ideally waits on / cross-references the dcm2niix fix, since anything else consuming dcm2niix's sidecar `AcquisitionTime` (not just heudiconv) inherits the same bias.

## Reproducer data

Public, no auth needed:
```
https://datasets.datalad.org/dbic/QA/sourcedata/sub-qa/ses-20180212/dwi/sub-qa_ses-20180212_acq-DTIX30Xp2Xs4_dwi.dicom.tgz
https://datasets.datalad.org/dbic/QA/sourcedata/sub-qa/ses-20180212/anat/sub-qa_ses-20180212_acq-MPRAGE_T1w.dicom.tgz
```
5 more confirmed-affected sessions (same protocol, same 2.585–2.588s offset): `ses-20190812`, `ses-20190801`, `ses-20190401`, `ses-20190408` (same tarball path pattern), plus the negative-control `ses-20250324` (different DICOM layout, offset=0).

Full list of all 388 `DTI-p2Xs4 → MPRAGE` transitions in this dataset, with `acq_time`/`duration`/computed gap/sourcedata paths for every session (sorted worst-first), is available on request — happy to attach as a CSV/TSV to this issue if useful for a regression test fixture.

yarikoptic-gitmate:
## Summary

This PR adds support for populating the `duration` column in BIDS `_scans.tsv` files, tracking the wallclock acquisition time of each scan. It includes both automatic population during conversion and a retrospective backfill utility for existing datasets.

Fixes #875.

## Key Changes

- **New `duration` field in scans metadata**: Added `duration` to `SCANS_FILE_FIELDS` with proper BIDS documentation, positioned after `acq_time` in the TSV output.

- **Acquisition duration detection** (`heudiconv/dicoms.py`):
  - `get_dicom_acquisition_duration()`: Extracts duration directly from DICOM tags (standard `AcquisitionDuration` tag or GE private tag)
  - `estimate_scan_duration_from_times()`: Estimates duration from per-file acquisition timestamps when tags are unavailable; accepts an optional `anchor_dt` so it stays self-consistent with `acq_time` (fixes #875)
  - `get_acquisition_duration()`: Unified interface trying tag-based extraction first, then timestamp-based estimation
  - `get_dicom_declared_acquisition_duration()` / `get_dicom_declared_repetitions()`: Siemens CSA/Phoenix-protocol last resort for single-volume sequences, guarded against truncated/aborted multi-volume runs

- **Enhanced scans row generation** (`heudiconv/bids.py`):
  - `get_formatted_scans_key_row()` now accepts a sequence of DICOM files (not just one) to enable timestamp-based duration estimation, and anchors `duration` on the same timestamp used for `acq_time`
  - Optimized to compute the row once per item rather than per output file
  - `_format_duration()`: Formats duration values as 3-decimal floats or 'n/a'

- **Retrospective duration backfill** (`heudiconv/bids.py`):
  - `populate_scans_duration()`: Public API to backfill `duration` in existing datasets
  - `_get_retrospective_duration()`: Best-effort duration estimation from sourcedata tarballs or NIfTI/JSON sidecars, anchored on the scan's existing `acq_time`
  - `_duration_from_dicom_tarball()`: Extracts duration from heudiconv-produced DICOM tarballs
  - `_duration_from_nifti_sidecar()`: Estimates duration from RepetitionTime × number of volumes (multi-volume runs only), or from `VolumeTiming` for sparse designs
  - `_populate_scans_duration_file()`: Processes individual `_scans.tsv` files with proper column handling
  - `_write_scans_tsv_atomically()`: Atomic TSV rewriting for safety with read-only/symlinked files

- **Backward compatibility**: Enhanced `add_rows_to_scans_keys_file()` to handle files with missing/differently-ordered columns by keying values by column name rather than position, preventing silent misalignment when older files lack the `duration` column.

- **CLI integration**: Added `--command populate-scans-duration` with `--overwrite` flag support in `heudiconv/main.py` and `heudiconv/cli/run.py`.

- **Utility functions** (`heudiconv/utils.py`):
  - `tar_extract_filter_kwargs()`: Safe tar extraction with Python 3.12+ filter support
  - `safe_extract_tar()`: Wrapper for secure archive extraction

## Notable Implementation Details

- Duration estimation is robust to multi-echo acquisitions and files sharing timestamps by using median interval between distinct timestamps
- `duration` is anchored on the same timestamp as `acq_time`, not independently recomputed, so `acq_time + duration` cannot land before the next scan's `acq_time` due to interleaved-slice timestamp quirks (#875)
- Handles datasets converted before this feature was added through retrospective backfill
- Gracefully degrades when source DICOMs are unavailable, falling back to NIfTI/JSON sidecar estimation
- Single-volume scans are excluded from TR × nvols estimation as it would vastly underestimate actual acquisition time
- A single-volume-declaring protocol falls back to Siemens' own declared total scan time as a last resort; a truncated/aborted multi-volume acquisition is left as unavailable rather than guessed
- Atomic file rewriting ensures safety even with read-only or git-annex-tracked files
- Comprehensive test coverage including edge cases (missing sourcedata, single volumes, malformed files, interleaved-timestamp anchoring)

https://claude.ai/code/session_01JzwZxQwxZMnoTKq9QkJKob


heudiconv/bids.py
```diff
@@ -532,27 +557,42 @@ def add_rows_to_scans_keys_file(fn: str, newrows: dict[str, list[str]]) -> None:
     fn: str
       filename
     newrows: dict
-      extra rows to add (acquisition time, referring physician, random string)
+      extra rows to add (acquisition time, duration, referring physician,
+      random string), in the order of ``SCANS_FILE_FIELDS`` (excluding
+      "filename")
     """
+    header = list(SCANS_FILE_FIELDS.keys())
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in 8b08a15: `add_rows_to_scans_keys_file` and `populate_scans_duration` now go through a shared `_merge_scans_header()` that preserves any existing column (including custom/dataset-specific ones) and only inserts the missing canonical column(s), at their canonical position.

---

heudiconv/bids.py
```diff
@@ -591,13 +643,323 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+def _format_duration(duration: Optional[float]) -> str:
+    """Format a duration in seconds as a `_scans.tsv` cell (``'n/a'`` if None)."""
+    return "%.3f" % duration if duration is not None else "n/a"
+
+
+def _is_within_directory(path: str, directory: str) -> bool:
+    """Return True if `path` resolves to somewhere at or under `directory`."""
+    path = op.abspath(path)
+    directory = op.abspath(directory)
+    return path == directory or path.startswith(directory + os.sep)
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in 8b08a15: `_is_within_directory` now resolves symlinks via `os.path.realpath` on both sides before comparing.

---

heudiconv/utils.py
```diff
@@ -565,6 +566,45 @@ def file_md5sum(filename: str) -> str:
...
+
+    Parameters
+    ----------
+    tarball : str
+        Path to the tar archive.
+    dest : str
+        Destination directory (created by ``tarfile`` as needed).
+    """
+    with tarfile.open(tarball) as tar:
+        tar.extractall(dest, **tar_extract_filter_kwargs())  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in 8b08a15: added `_safe_tar_members()` in `heudiconv/utils.py`, a manual member-vetting fallback (rejects absolute paths, `..` traversal, device/FIFO members, and symlinks/hardlinks whose target would escape the destination) used whenever `tar_extract_filter_kwargs()` returns `{}` (Python < 3.12). `heudiconv/parser.py`'s own tar extraction now routes through the same hardened `safe_extract_tar()` too, instead of calling `shutil.unpack_archive` with the (on <3.12) empty kwargs directly.

---

heudiconv/bids.py
```diff
@@ -50,6 +53,24 @@
                 ]
             ),
         ),
+        (
+            "duration",
+            OrderedDict(
+                [
+                    ("LongName", "Scan duration"),
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in 8b08a15: `populate_bids_templates()` now calls a new `_sync_scans_json_fields()` helper right after `create_file_if_missing`, adding any missing `SCANS_FILE_FIELDS` key (e.g. `duration`) to an existing `scans.json` without touching anything else (including user-defined keys).

---

heudiconv/bids.py
```diff
@@ -591,13 +643,323 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+    Returns
+    -------
+    Optional[float]
+        Duration in seconds, or None if it could not be determined.
+    """
+    json_fn = _nifti_stem(nifti_fn) + ".json"
+    if not op.exists(json_fn):
+        return None
+    tr = load_json(json_fn).get("RepetitionTime")
+    if not tr:
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in 8b08a15: `_duration_from_nifti_sidecar()` now prefers the sidecar's own `AcquisitionDuration` (when finite and positive) unless `VolumeTiming` is present, in which case it falls through to the `RepetitionTime x nvols` estimate as before (since `AcquisitionDuration` has per-volume semantics there).

---

heudiconv/bids.py
```diff
@@ -591,13 +643,323 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+        return None
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in 8b08a15: now validated via a shared `as_finite_positive_float()` helper (rejects non-numeric, non-positive, `NaN`, and infinite values), used consistently for `RepetitionTime` here and for the DICOM-tag-based duration in `heudiconv/dicoms.py`.

---

heudiconv/bids.py
```diff
@@ -591,13 +643,323 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+    if op.exists(tarball):
+        duration = _duration_from_dicom_tarball(tarball)
+        if duration is not None:
+            return duration
+        lgr.debug(
+            "Could not establish duration from source DICOMs %s for %s",
+            tarball,
+            nifti_fn,
+        )
+    return _duration_from_nifti_sidecar(nifti_fn)
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in 8b08a15: the call to `_duration_from_dicom_tarball()` inside `_get_retrospective_duration()` is now wrapped in `try/except`, logging and falling through to the NIfTI/JSON sidecar estimate on failure instead of propagating and aborting the rest of that `_scans.tsv`'s rows.

---

heudiconv/bids.py
```diff
@@ -591,13 +643,323 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+    directory = op.dirname(scans_tsv) or "."
+    fd, tmp_path = tempfile.mkstemp(
+        dir=directory, prefix=".heudiconv-scans-", suffix=".tsv"
+    )
+    try:
+        with os.fdopen(fd, "w", newline="") as f:
+            writer = csv.DictWriter(f, fieldnames=fieldnames, delimiter="\t")
+            writer.writeheader()
+            writer.writerows(rows)
+        os.replace(tmp_path, scans_tsv)
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Investigated rather than changed. Confirmed by hand in a git-annex repo with `*_scans.tsv annex.largefiles=anything` (matching `heudiconv/external/dlad.py`): before the rewrite the file is a symlink into a read-only annex object; `_write_scans_tsv_atomically`'s `os.replace` does turn it into a plain file at that path -- but that's exactly the same unlink-and-recreate pattern the pre-existing `add_rows_to_scans_keys_file` already used for this exact file, on every heudiconv version before this PR too. Running `git annex add` on it afterward (which `add_to_datalad`/`datalad save` already does after every heudiconv conversion) re-annexes it cleanly into a fresh object with no lingering broken state. So this isn't a regression relative to how heudiconv already manages `_scans.tsv`; rejecting symlink destinations outright would instead make `populate_scans_duration` fail on every DataLad-tracked dataset, which is one of the use cases the retrospective backfill is meant to serve. Happy to revisit if there's a concrete scenario where the existing pattern doesn't round-trip the same way.

---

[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in 8b08a15: `_write_scans_tsv_atomically()` now `os.chmod`s the temp file to match the original's (dereferenced) permission bits before `os.replace`.

---

heudiconv/dicoms.py
```diff
@@ -589,6 +589,164 @@ def get_datetime_strings_from_dcm(
...
+        if dt is not None:
+            timestamps.append(dt)
+    unique_timestamps = sorted(set(timestamps))
+    if len(unique_timestamps) < 2:
+        return None
+    intervals = sorted(
+        (t2 - t1).total_seconds()
+        for t1, t2 in zip(unique_timestamps[:-1], unique_timestamps[1:])
+    )
+    median_interval = intervals[len(intervals) // 2]
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in 8b08a15: now uses `statistics.median()` over the interval list instead of indexing the sorted list's midpoint, so an even number of intervals is averaged correctly (e.g. the 1s/100s example now correctly adds 50.5s).

---

heudiconv/utils.py
```diff
@@ -565,6 +567,120 @@ def file_md5sum(filename: str) -> str:
...
+            lgr.warning("Refusing to extract %r: escapes %s", member.name, dest)
+            continue
+        if member.isdev():
+            lgr.warning("Refusing to extract device/FIFO member %r", member.name)
+            continue
+        if (member.issym() or member.islnk()) and not _resolves_within_dest(
+            member.linkname
+            if op.isabs(member.linkname)
+            else op.join(dest, op.dirname(member.name), member.linkname)
+        ):
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in baad4e3: `_safe_tar_members()` now resolves hard-link (`islnk()`) targets relative to `dest` (the archive root), separately from symlink (`issym()`) targets which remain resolved relative to the member's own directory. Added `test_safe_tar_members_hardlink_target_relative_to_dest` covering both a same-archive-root hard link and an escaping one.

---

heudiconv/utils.py
```diff
@@ -565,6 +567,120 @@ def file_md5sum(filename: str) -> str:
...
+    this is safe to use on every Python version heudiconv supports.
+
+    Parameters
+    ----------
+    tarball : str
+        Path to the tar archive.
+    dest : str
+        Destination directory (created by ``tarfile`` as needed).
+    """
+    tar_filter = os.environ.get("HEUDICONV_TAR_FILTER", "tar")
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in baad4e3: the default `HEUDICONV_TAR_FILTER` is now `"data"` instead of `"tar"`, so on interpreters using the native extraction filter, device/FIFO members and unsafe link targets are rejected the same way our own `_safe_tar_members()` fallback already rejects them. A weaker filter remains available via the env var for anyone who explicitly needs it.

---

heudiconv/bids.py
```diff
@@ -591,13 +721,373 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+
+    Returns
+    -------
+    Optional[float]
+        Duration in seconds, or None if it could not be determined.
+    """
+    json_fn = _nifti_stem(nifti_fn) + ".json"
+    if not op.exists(json_fn):
+        return None
+    meta = load_json(json_fn)
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in baad4e3: `_duration_from_nifti_sidecar()` now wraps the `load_json()` call (and a check that the result is actually a JSON object) in `try/except Exception`, logging a warning and returning `None` on any read/parse/type failure instead of propagating. Added `test_duration_from_nifti_sidecar_tolerates_malformed_json` (invalid JSON, a JSON array, and a JSON string, all parametrized).

---

heudiconv/bids.py
```diff
@@ -591,13 +721,373 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+
+    if "VolumeTiming" not in meta:
+        duration = as_finite_positive_float(meta.get("AcquisitionDuration"))
+        if duration is not None:
+            return duration
+
+    tr = as_finite_positive_float(meta.get("RepetitionTime"))
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in baad4e3: when `VolumeTiming` is present, `_duration_from_nifti_sidecar()` no longer falls through to the (never-available, since `RepetitionTime`/`VolumeTiming` are mutually exclusive) `RepetitionTime x nvols` estimate. It now derives the total from the last volume's onset (`max(VolumeTiming)`) plus one frame's acquisition duration -- preferring `FrameAcquisitionDuration`, falling back to the deprecated per-frame use of `AcquisitionDuration` -- and returns `None` only if neither is available. Added `test_duration_from_nifti_sidecar_derives_from_volume_timing`, `..._prefers_frame_acquisition_duration`, and `..._without_frame_duration`.

---

heudiconv/dicoms.py
```diff
@@ -589,6 +591,153 @@ def get_datetime_strings_from_dcm(
...
+        "SeriesTime",
+    ]
+    timestamps = []
+    for fn in dicom_list:
+        dcm_data = dcm.dcmread(
+            fn, stop_before_pixels=True, force=True, specific_tags=datetime_tags
+        )
+        dt = get_datetime_from_dcm(dcm_data)
+        if dt is not None:
+            timestamps.append(dt)
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in baad4e3: `estimate_scan_duration_from_times()` now catches `ValueError` around `get_datetime_from_dcm()` per file, logs a warning, and skips just that file instead of letting a single malformed timestamp abort the estimate (and, transitively, normal conversion while writing `_scans.tsv`). Added `test_estimate_scan_duration_from_times_skips_malformed_timestamp`.

---

heudiconv/utils.py
```diff
@@ -565,6 +567,120 @@ def file_md5sum(filename: str) -> str:
...
+        Value to parse (e.g. read from a DICOM tag or a JSON sidecar).
+
+    Returns
+    -------
+    Optional[float]
+        The parsed value, or None if it does not qualify.
+    """
+    try:
+        parsed = float(value)
+    except (TypeError, ValueError):
```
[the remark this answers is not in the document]
yarikoptic-gitmate: Fixed in baad4e3: `as_finite_positive_float()` now also catches `OverflowError` alongside `TypeError`/`ValueError`, so an oversized numeric value (e.g. from malformed JSON metadata) is treated as unavailable rather than propagating. Added a parametrized case (`10**400`) to `test_as_finite_positive_float`.

---

heudiconv/bids.py
```diff
@@ -591,13 +727,552 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+    ``RepetitionTime`` is absent entirely (there is no conflicting
+    metadata to arbitrate in that case, just a missing per-frame duration).
+
+    Otherwise, computes ``RepetitionTime`` (from the JSON sidecar) times
+    the number of volumes (from the NIfTI header), which is only
+    meaningful for multi-volume (e.g., functional) runs.  This is a
+    coarse approximation: it cannot account for any preparation/dummy-scan
+    time preceding the first recorded volume, so it is used only as a
+    last resort, when no source DICOMs -- nor a usable
+    ``AcquisitionDuration`` -- are available.
```
yarikoptic: this is too long docstring! gory details of implementation should be THE CODE below. Here -- provide only high level description within terse paragraph!

yarikoptic: go through ALL added functions/descriptions to make them all concise!

heudiconv/bids.py
```diff
@@ -591,13 +727,552 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+    session_dir = op.dirname(scans_tsv)
+    bids_root = _find_bids_dataset_root(session_dir)
+
+    # utf-8-sig transparently strips a leading BOM if present (see
+    # add_rows_to_scans_keys_file for why that matters for the "filename"
+    # header check just below)
+    with open(scans_tsv, newline="", encoding="utf-8-sig") as f:
+        reader = csv.DictReader(f, delimiter="\t")
+        fieldnames = list(reader.fieldnames or [])
+        rows = list(reader)
```
yarikoptic: feels like there should already be or should be some reusable function to read a scans tsv file to be reused, isn't there?

heudiconv/bids.py
```diff
@@ -591,13 +727,552 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+    if not fieldnames:
+        return
+    if "filename" not in fieldnames:
+        lgr.warning("%s has no 'filename' column, skipping", scans_tsv)
+        return
+    if any(None in row for row in rows):
+        # csv.DictReader stashes any fields beyond the header under the
+        # `None` key; writing those back out under our (possibly extended)
+        # `fieldnames` would raise deep inside csv.DictWriter, so bail with
+        # a clear message instead of leaving a half-written file behind
```
yarikoptic: warning text alone should be enough -- no need for extra comment!!!

heudiconv/bids.py
```diff
@@ -931,7 +1606,10 @@ def select_fmap_from_compatible_groups(
         scans_tsv = glob(op.join(sess_folder, "*_scans.tsv"))[0]
     except IndexError:
         raise FileNotFoundError("No '*_scans' file found for session %s" % sess_folder)
-    with open(scans_tsv) as f:
+    # utf-8-sig transparently strips a leading BOM if present, so a BOM'd
+    # "filename" header still matches the literal below instead of making
+    # index() raise
+    with open(scans_tsv, encoding="utf-8-sig") as f:
```
yarikoptic: here is one more use!  generalize into a helper function!

heudiconv/bids.py
```diff
@@ -591,13 +727,552 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+    treated as escaping it, rather than merely comparing lexical
+    (unresolved) absolute paths.
+    """
+    path = op.realpath(path)
+    directory = op.realpath(directory)
+    return path == directory or path.startswith(directory + os.sep)
+
+
+def _parse_bids_entities(filename: str) -> tuple[dict[str, str], str]:
+    """Split a BIDS filename into its entities and suffix.
```
yarikoptic: I feel we already have construct like BIDSFile or alike to be used here and to avoid duplication of functionality!

heudiconv/bids.py
```diff
@@ -591,13 +727,552 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+        dicom_files = sorted(str(p) for p in Path(tmpdir).rglob("*") if p.is_file())
+        if not dicom_files:
+            lgr.warning("No files found within %s", tarball)
+            return None
+        return dicoms.get_acquisition_duration(dicom_files)
+
+
+def _nifti_stem(nifti_fn: str) -> str:
+    """Strip a NIfTI extension (``.nii`` or ``.nii.gz``) off `nifti_fn`."""
+    return remove_suffix(remove_suffix(nifti_fn, ".gz"), ".nii")
```
yarikoptic: this is not specific to nifti, but rather getting out multi-component extension (.tar.gz etc) -- I think we have some helper in datalad for that! reuse that for consistent functionality

heudiconv/bids.py
```diff
@@ -591,13 +727,552 @@ def get_formatted_scans_key_row(dcm_fn: str | Path) -> list[str]:
...
+        Path to a ``*.dicom.tgz`` file, as produced under ``sourcedata/`` by
+        heudiconv's ``-c dicom`` conversion (see
+        :func:`heudiconv.dicoms.compress_dicoms`).
+
+    Returns
+    -------
+    Optional[float]
+        Duration in seconds, or None if it could not be determined.
+    """
+    with tempfile.TemporaryDirectory() as tmpdir:
```
yarikoptic: will this remove that folder at the end automatically or will we keep breeding them in /tmp?

heudiconv/bids.py
```diff
@@ -565,21 +689,33 @@ def add_rows_to_scans_keys_file(fn: str, newrows: dict[str, list[str]]) -> None:
...
 
     """
-    dcm_data = dcm.dcmread(dcm_fn, stop_before_pixels=True, force=True)
+    if isinstance(dcm_fns, (str, Path)):
+        dcm_fns = [dcm_fns]
+    dcm_fn_strs: list[str] = [str(f) for f in dcm_fns]
+    dcm_data = dcm.dcmread(dcm_fn_strs[0], stop_before_pixels=True, force=True)
     # we need to store filenames and acquisition datetimes
     acq_datetime = dicoms.get_datetime_from_dcm(dcm_data=dcm_data)
+    duration = dicoms.get_acquisition_duration(dcm_fn_strs)
```
yarikoptic: may be for consistency, name it `acq_duration`?

yarikoptic (changes requested):
first glance over. Pay attention to not be too verbose and also pay attention to avoid duplication

heudiconv/bids.py
yarikoptic-gitmate: Trimmed in 97ce9c9 -- the docstring is now a short high-level paragraph, with the fallback order and the `HEUDICONV_VOLUME_TIMING_FRAME_DURATION` opt-in left to the code below.

---

yarikoptic-gitmate: Done in 97ce9c9 -- trimmed every new/modified docstring in `bids.py` and `dicoms.py` down to a short, high-level description, leaving implementation details to the code.

---

heudiconv/bids.py
yarikoptic-gitmate: Fixed in 97ce9c9: added a shared `_read_scans_tsv()` helper (also tolerating the UTF-8 BOM some Excel/Windows-authored TSVs carry) and switched `_populate_scans_duration_file`, `add_rows_to_scans_keys_file`, and `select_fmap_from_compatible_groups` to use it instead of each reading the file on its own.

---

heudiconv/bids.py
yarikoptic-gitmate: Fixed in 97ce9c9 -- removed the redundant comment next to the warning.

---

heudiconv/bids.py
yarikoptic-gitmate: Fixed in 97ce9c9 -- `select_fmap_from_compatible_groups` now uses the same `_read_scans_tsv()` helper as the other two call sites, instead of its own manual TSV reading.

---

heudiconv/bids.py
yarikoptic-gitmate: Fixed in 97ce9c9 -- removed `_parse_bids_entities` and switched `_find_json_sidecar` to use `BIDSFile.parse()` instead. That also surfaced (and fixed) a real `IndexError` in `BIDSFile.parse()` for entity-less filenames (e.g. `bold.json`), and added a `.entities` property so `_find_json_sidecar` can compare entities without reaching into `BIDSFile` internals.

---

heudiconv/bids.py
yarikoptic-gitmate: Went with nibabel's own `nibabel.filename_parser.splitext_addext()` instead (it already handles `.nii`/`.nii.gz`/etc.) rather than adding a datalad dependency for this -- fixed in 97ce9c9. Happy to switch to a specific datalad helper instead if you'd rather standardize on that.

---

heudiconv/bids.py
yarikoptic-gitmate: Yes -- it's used as a context manager (`with tempfile.TemporaryDirectory() as tmpdir:`), so it's removed automatically when the `with` block exits, including on exception. Nothing accumulates in `/tmp`.

---

heudiconv/bids.py
yarikoptic-gitmate: Done in 97ce9c9 -- renamed to `acq_duration`.

---

heudiconv/tests/test_bids.py
```diff
@@ -1645,3 +1682,750 @@ def test_sanitize_label() -> None:
...
+    # with overwrite=True, it gets recomputed:
+    populate_scans_duration(str(bids_root), overwrite=True)
+    _, rows = _read_scans_rows(scans_tsv)
+    assert float(rows[0]["duration"]) == pytest.approx(20.0)
+
+
+@pytest.mark.ai_generated
+def test_populate_scans_duration_no_scans_tsv(tmp_path: Path) -> None:
+    # should warn and do nothing, rather than raise, when nothing is found
+    populate_scans_duration(str(tmp_path))
```
yarikoptic: what's the point of this test beyond smoke??? add actual testing of expected behaviors

heudiconv/tests/test_bids.py
```diff
@@ -1645,3 +1682,750 @@ def test_sanitize_label() -> None:
...
+    bids_root = tmp_path / "bids"
+    scans_tsv = _make_bids_dataset_stub(bids_root)
+    _make_multivol_func_scan(bids_root)
+
+    # a single-file "series" cannot yield a duration on its own (neither a
+    # tag nor >= 2 timestamps to estimate from) -- strip the CSA series
+    # header this fixture happens to carry too (Siemens' own declared
+    # duration), since that alone would otherwise let
+    # get_acquisition_duration() succeed and defeat the point of this test
+    import pydicom
```
yarikoptic: imports should be on top of the file!

heudiconv/tests/test_bids.py
```diff
@@ -1645,3 +1682,750 @@ def test_sanitize_label() -> None:
...
+    populate_scans_duration(str(bids_root))
+
+    _, rows = _read_scans_rows(scans_tsv)
+    assert float(rows[0]["duration"]) == pytest.approx(20.0)
+
+
+@pytest.mark.ai_generated
+def test_populate_scans_duration_via_cli(tmp_path: Path) -> None:
+    bids_root = tmp_path / "bids"
+    scans_tsv = _make_bids_dataset_stub(bids_root)
```
yarikoptic: these two lines look too common -- if we are using them through out -- make local `tmp_bids_with_scans` fixture or alike to reuse and shrink?

